PR #4, opened 2026-08-30
Add empty file creation from the folder context menu

mrocha-montes:
## Summary
- Generalizes the existing "New Folder" inline entry into a shared new-entry row that creates either a directory or an empty file
- Adds a "New File" option to the folder background context menu, alongside "New Folder"
- Adds `CreateFileRequest`/`OperationProvider::create_file`, implemented in the local adapter via `gio::File::create_future`
- Checks off "Create an empty file" in `docs/todo.md`

Closes the P0 work-breakdown item for creating an empty file.

## Test plan
- [x] `./scripts/check.sh` (fmt, clippy `-D warnings`, tests, cargo-deny, typos) passes
- [x] `cargo test` — 112 passed
- [x] Manually ran the app and confirmed the folder browser still opens correctly and "New Folder" (Ctrl+Shift+N and context menu) is unaffected

### DIFF
--- a/src/adapters/local_operations.rs
+++ b/src/adapters/local_operations.rs
@@ -10,8 +10,8 @@ use gtk::{gio, glib, prelude::*};
 use crate::{
     model::Location,
     services::{
-        CreateDirectoryRequest, DeleteRequest, LoadHandle, OperationEvent, OperationProvider,
-        PasteRequest, RenameRequest, RestoreRequest, validate_basename,
+        CreateDirectoryRequest, CreateFileRequest, DeleteRequest, LoadHandle, OperationEvent,
+        OperationProvider, PasteRequest, RenameRequest, RestoreRequest, validate_basename,
     },
 };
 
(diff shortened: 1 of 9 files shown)


l0gicgate:
Nice one.

I haven’t yet created templates for PRs (moving a little fast at the moment) but would you be able to include a short video/screenshots of the changes as supporting evidence? That would be great. 

I will pull locally soon and test myself!

src/adapters/local_operations.rs
```diff
@@ -189,6 +189,29 @@ impl OperationProvider for LocalOperationProvider {
         LoadHandle::new(move || task.abort())
     }
 
+    fn create_file(
+        &self,
+        request: CreateFileRequest,
+        emit: Rc<dyn Fn(OperationEvent)>,
+    ) -> LoadHandle {
+        let task = glib::MainContext::default().spawn_local(async move {
+            let file = gio_file(&request.parent).child(&request.name);
```
l0gicgate: `gio_file(parent).child(name)` treats the input as a path, not strictly a `basename`:                                                                                                                                                                                                 
     - `../escaped` creates a file outside the selected folder.                                                                                                                                                                                                                        
     - `nested/name` traverses into a subdirectory.                                                                                                                                                                                                                                    
     - An absolute path causes GLib’s `g_file_get_child` assertion to fail and may panic/crash through the Rust binding.                                                                                                                                                               
                                                                                                                                                                                                                                                                                     
   The existing `validate_rename()` already rejects `/`, `.` and `..`; creation should apply equivalent validation before calling the provider, ideally for both files and folders. `child_for_display_name()` may also be more appropriate for user-entered names.

src/adapters/local_operations.rs
```diff
@@ -189,6 +189,29 @@ impl OperationProvider for LocalOperationProvider {
         LoadHandle::new(move || task.abort())
     }
 
+    fn create_file(
```
l0gicgate, line 348: Can we add tests for this as well please.

mrocha-montes:
Here's a short recording of the new "New File" action: right-click a column, pick "New File", the inline row swaps to the document icon, type a name, Enter creates the empty file.

https://github.com/user-attachments/assets/b474def1-5667-4806-938b-75fb1d55c5b4

l0gicgate:
I will take this PR over and fix all the conflicts. I had to merge some stuff with safe transfer ops that conflict with this. Apologies for that.

Also, always have to make sure a feature works in all 3 modes (List, Grid, Explorer). I am fixing that too.
image

PR #30, opened 2026-08-31
Don't flag an empty name field as an error

mrocha-montes:
Found this while testing something else, but it is not specific to that scenario, it reproduces for a plain local New Folder too.

## The bug

`update_basename_validation` runs on every text change, including the programmatic clears that happen when a New Folder/File prompt opens, cancels, or succeeds after creating. Since `validate_basename("")` treats an empty name as invalid, any of those routine clears can leave the field styled red before the user has typed anything wrong, or anything at all.

Repro:
1. Open New Folder, type an invalid name (for example one with a `/`), press Enter. It gets rejected in place, no dialog, field turns red. So far expected.
2. Press Escape, then open New Folder again.
3. The field is already red and empty, nothing has been typed yet.

Confirmed present on the current `main` (v0.4.0).

## The fix

Treat an empty field as the normal, unstyled starting state rather than an error state. Submission is still blocked (the function still returns `false` for an empty name), only the alarming red border for a field nobody has touched yet is gone.

Added a unit test constructing a real `gtk::Entry` and driving `update_basename_validation` directly, `check.sh` passes (142 tests).

### DIFF
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -4925,17 +4925,6 @@ fn set_cut_path_style(row: &gtk::Box, cut: bool) {
     }
 }
 
-/// Validates a name field live as it changes, including the programmatic
-/// clears that happen when a prompt opens, cancels, or succeeds. An empty
-/// field is left unstyled rather than flagged red: it's the normal starting
-/// state, not a mistake the user made, even though it still can't be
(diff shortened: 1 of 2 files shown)


l0gicgate:
Thank you for the fix @mrocha-montes, validated. I removed Claude's comments and merging now!

Issue #20, opened 2026-08-30
Support SMB, SFTP, FTP, and WebDAV locations and saved connections


l0gicgate:
## Summary

Add first-class SMB, SFTP, FTP/FTPS, and WebDAV locations using GIO/GVfs. A user should be able to paste a supported remote URI into the editable location field and browse it, authenticate when required, and optionally save the destination as a named connection in the sidebar.

This should remain URI-native inside Strata. Do not turn a GVfs network mount into its opaque `/run/user/$UID/gvfs/…` FUSE path; #5 tracks that related sidebar bug.

### Direct navigation

- `Ctrl+L` accepts absolute local paths plus `smb://`, `sftp://`, `ftp://`, `ftps://`, `dav://`, and `davs://` URIs (scheme matching is case-insensitive and normalized).
- Supported examples:
  - `smb://server/share`
  - `smb://user@server/share/folder`
  - `sftp://host/home/user`
  - `sftp://user@host:2222/home/user`
  - `ftp://user@host/public`
  - `ftps://user@host/public`
  - `dav://host/remote.php/dav/files/user`
  - `davs://host/remote.php/dav/files/user`
- Prefer encrypted schemes: SFTP, FTPS, and DAVS. Before connecting over plaintext FTP or DAV, show a one-time warning that credentials and content may be exposed in transit. Do not weaken TLS validation for FTPS/DAVS; certificate errors require an explicit backend trust decision.
- Treat `https://` as a web URL, not implicitly as WebDAV. The Add Connection form may accept an HTTPS endpoint while WebDAV is selected and normalize it to the backend's DAVS representation.
- A URI may point at a saved connection root or a deeper folder. Preserve the clean canonical URI in history, breadcrumbs, tooltips, clipboard/drag data, and errors.
- UNC syntax (`\\server\share`) and SCP shorthand (`user@host:path`) are not part of this task; give a useful validation message rather than treating either as a local relative path.
- Unsupported URI schemes fail explicitly and identify the supported schemes.

### Network discovery

- Add a **Network** location that automatically discovers browsable devices and services on the local network, similar to macOS Finder, so users do not need to know or enter an IP address first.
- Use GIO/GVfs network discovery rather than implementing protocol discovery directly in Strata. Discovered hosts and shares must remain URI-native and flow through the same mount/authentication behavior as manually entered locations.
- Keep direct URI entry available when discovery is unavailable, disabled by local network policy, or does not advertise a device.
[...]

PR #31, opened 2026-08-31
Add SMB network location support (browse, mount, authenticate)

### LINKED ISSUE #5: Mounted-volume sidebar entries resolve GVfs network mounts to their FUSE path, not their URI
## Context

While experimenting locally with SMB/network location support (not yet submitted — still WIP), I noticed that once a GVfs network mount actually succeeds (e.g. via `gio::File::mount_enclosing_volume`/`mount_mountable` on an `smb://` location), it shows up automatically in the sidebar's "DEVICES" section through the existing mounted-volumes code, and it's presented very differently from how the rest of the app treats network locations.

## Where

`src/ui/window.rs`, both in `SidebarState::rebuild` (the `mounts` mapping around line 692) and in `append_volume` (around line 911):

```rust
let root = mount.root();
let location = root
    .path()
    .map(Location::local)
    .unwrap_or_else(|| Location::uri(root.uri()));
```

## Problem

For local/removable-device mounts, `root.path()` is the right thing to prefer. But for a GVfs network backend (smb, sftp, dav, ...), the automounter also exposes the mount through a FUSE bind path (`/run/user/$UID/gvfs/smb-share:server=...,share=...`), so `root.path()` succeeds too — just with that opaque FUSE path instead of `None`. The `unwrap_or_else` never gets a chance to prefer the clean `smb://host/share` URI in that case.

Concretely, once such a mount is registered:

- The sidebar entry's location, breadcrumb, and any further navigation happen through the `/run/user/1000/gvfs/smb-share:server=192.168.1.220,share=music`-style native path rather than `smb://192.168.1.220/music` — inconsistent with how the rest of the app models network locations as explicit URIs (see the "Native paths stay native" boundary in `docs/architecture.md`).
- Descending further from that FUSE-path breadcrumb misbehaved in my testing — one subfolder click surfaced "Unable to open directory: That location does not exist." I haven't dug into why; it may be specific to how that particular path was constructed, but the FUSE-path presentation itself seems like the wrong thing to be navigating through by design, regardless.

## Ask

Decide and document the intended convention for GVfs network mounts appearing in this sidebar list — I'd expect preferring the mount's URI (`root.uri()`) over its FUSE path whenever the mount isn't a plain local/removable device — before network location support (SMB/SFTP) lands, so the "mounted volumes" sidebar feature and network browsing agree on one presentation instead of drifting into two different (and differently-buggy) code paths.

I don't have a concrete fix in mind yet since I haven't decided the right way to distinguish "local device mount" from "GVfs network mount" at this call site — happy to discuss approaches.

mrocha-montes:
## Summary

- "Network" sidebar entry backed by GVfs discovery, same pattern as Trash.
- `Ctrl+L` accepts any well-formed `scheme://` URI, not just `trash:///`, so `smb://`, `sftp://`, `ftp://`, `dav://`, etc. all reach GIO instead of being rejected as "not an absolute path."
- Async auto-mount flow using a native `GtkMountOperation` for credentials. Correctly dispatches to `mount_enclosing_volume` vs `mount_mountable` depending on the GVfs entry type (they are not interchangeable).
- UNC (`\\host\share`) and SCP-style (`user@host:path`) shorthand rejected with a clear message instead of being misread as a local path, per this issue's own product decisions.
- A URI with an embedded password is rejected outright; a bare username is still accepted and prompts normally.
- A missing GVfs backend reports which package to install (e.g. `gvfs-smb`) instead of a raw GIO error.
- Cancelling the credential prompt returns quietly to the prior state, no alarming error.

Closes #5: GVfs mounts now stay URI-native everywhere a `Location` is derived from a `gio::File` (sidebar, volumes, click-to-navigate, and every child entry produced while enumerating a remote directory), instead of leaking the opaque `/run/user/$UID/gvfs/...` FUSE path.

Two more bugs found and fixed while testing against a real SMB share:
- A background directory-batch refresh could steal keyboard focus from an open New Folder/File prompt on slow network shares that stream many batches (local directories load in one fast batch, so this never showed up there).
- Creating an item on a remote location never appeared in the view until navigating away and back, since remote locations have no live directory monitor.

Also improved: Trash on a backend that doesn't support it (SMB has none) now explains that and suggests permanent delete, instead of surfacing a raw GIO error.

## Not included

Left for later slices per this issue's own scope: SFTP/FTP/WebDAV-specific testing (the code is protocol-agnostic through GIO, but only verified against SMB), saved Connections UI/persistence, a Disconnect action, and capability-gated hiding of unsupported operations like Trash.

## Testing

161 automated tests (`./scripts/check.sh` clean), plus manual end-to-end testing against a real SMB share on my LAN. Recording attached in a follow-up comment.

---

One more thing: I'll be away from this for a bit starting tomorrow, work week kicking off. No rush reviewing, take your time. Good luck with everything!


mrocha-montes:
Here's a short recording of the end-to-end flow: typing an smb:// URI, browsing the discovered shares, and the native credential prompt mounting successfully.

https://github.com/user-attachments/assets/39e1c9ee-dd0d-4117-8a02-13977ba3f013

l0gicgate:
Dude that's really cool.

I think the 2 main things I want to address:
1. Making the design inside this dialog cohesive with the rest of the design system. (I can do this)
image

2. The second is adding a right click context menu in the nav on SMB devices, so you can "forget" them or "edit credentials" maybe? Not sure yet.

I can take it from here tho. Solid @mrocha-montes thank you for the foundational work!

spuder:
Love it, just what I am looking for. 

How big of an ask would it be to make it so it could auto discover devices on the network (like OSX Finder), without having to know the ip address? 

Screenshot 2026-08-30 at 22 57 16

l0gicgate:
@spuder, I don't think that's a big ask. I can create a follow up ticket for it right now as this PR is already getting large!

l0gicgate:
Here is the restyled authentication dialog:
image

l0gicgate:
Added right click context menu:
image

l0gicgate:
Addressed the follow-up review findings in `75fe2a7`:

- moved remote location validation to cancellable async GIO queries so unreachable network locations do not block GTK's main thread;
- restricted `Ctrl+L` to the supported remote schemes (`smb`, `sftp`, `ftp`, `ftps`, `dav`, and `davs`, plus Strata's internal Trash/Network locations), normalized scheme casing, and added explicit unsupported-scheme errors;
- routed GTK bookmarks, clipboard sources, and file-list drops through the shared URI-native `location_for_file` helper so GVfs FUSE paths do not leak back into navigation;
- added URI-native parent/breadcrumb handling and refreshes for remote rename plus both sides of remote moves, preventing stale columns where no monitor is available;
- changed authentication retries to restart the backend mount operation and honor the backend's actual `AskPasswordFlags` instead of synthesizing SMB-specific fields/capabilities.

Also expanded coverage for unsupported schemes, async mount rejection events, remote URI parents/breadcrumbs, and remote rename refreshes. `./scripts/check.sh` is clean with 172 tests passing.

The network auto-discovery requirement and acceptance criterion were also added to #20, linked back to the request in this PR.

mrocha-montes:
Found a real bug testing this live against an SMB server on my LAN (192.168.1.220): opening a share that rejects the connection (`Failed to mount Windows share: Permission denied`) made the authentication dialog flicker indefinitely, impossible to type into.

Root cause, confirmed with timing logs: the mount operation resolves (with an error) well under 250ms after `ask-password` fires, with no button clicked and no reply sent from our side. So `mount_enclosing_volume_future`/`mount_mountable_future` is completing on its own before the user ever gets a chance to answer the prompt, and the automatic "retry on auth failure" logic in `mount_then_navigate_attempt`/`mount_then_descend_attempt` had no cap, so it just kept looping.

Pushed `e8ee172` as a quick safety net: caps automatic retries at 3 attempts, after which it falls back to a normal error dialog instead of looping forever. That fixes the symptom, but not the underlying issue: with the custom `gio::MountOperation`-based dialog (replacing the old `gtk::MountOperation`), the mount doesn't seem to actually wait for our reply in every case, at least against this server. Wanted to flag it since you're actively working this file and have more context on why the switch was made. Happy to help dig further if useful.

(Was just curious to peek at how the review was going and couldn't resist poking at it. Back to my actual day job now for real this time!)

l0gicgate:
Followed up in `97bfcbd` with a narrower retry condition.

The retry exists so that, after the user submits incorrect credentials and the GIO mount operation completes, Strata can start a fresh backend mount operation and let the backend request corrected credentials using its real `AskPasswordFlags`. It should not treat every authentication-looking backend error as a reason to retry.

Strata now records whether **Connect** was clicked for each mount operation and retries only when:

1. credentials were actually submitted;
2. the completed operation reports an authentication failure; and
3. the existing retry cap has not been reached.

If the backend fails before the user replies, Strata now reports the failure instead of tearing down and recreating the prompt. `./scripts/check.sh` passes with 179 tests.

@mrocha-montes, could you please retest this against the SMB server that reproduced the sub-250ms failure? In particular, please confirm that the prompt no longer flickers/reopens when the backend fails before any click, and that submitting genuinely incorrect credentials still opens a fresh retry prompt.

l0gicgate:
@mrocha-montes I setup some smb shares to test locally and found more regressions. I'm fixing the PR further. Will report back once done.

l0gicgate:
The local SMB fixture reproduced the remaining UX problem: the backend can return `Permission denied` before the user submits anything, and the previous safety fix then showed a terminal error instead of giving the user a way to authenticate.

Fixed in `fcbc2c7` by making retries explicitly user-driven:

- an authentication failure now leaves/reopens a stable credentials dialog rather than automatically restarting the mount;
- clicking **Connect** starts a fresh mount and supplies those credentials when the backend asks for them;
- rejected submitted credentials return to the dialog with the failure message and previous username/domain;
- when the backend emitted `ask-password` before completing early, its real message, defaults, and `AskPasswordFlags` are preserved for the stable dialog;
- when it never emitted `ask-password`, Strata uses a generic remote-login fallback;
- cancelling the fallback prompt restores the prior navigation state.

There is no timer-driven or failure-driven retry loop now; every subsequent mount attempt requires a user click. `./scripts/check.sh` passes with 179 tests.

@mrocha-montes, could you retest both the original early-completion server and an incorrect-then-correct credential attempt with this revision?

l0gicgate:
Confirmed locally against a loopback Samba fixture that reproduces the fast `Permission denied` path: the backend completed before any credential submission, and Strata now presents a stable login dialog instead of looping or ending at an error.

The final fix is in `fcbc2c7`:

- authentication failure itself no longer triggers a mount retry;
- if the mount completes before the user can answer, Strata preserves the backend's prompt details and presents them in a dialog detached from the completed operation;
- if the backend never emitted `ask-password`, Strata presents a generic remote-login fallback;
- only clicking **Connect** starts a fresh mount attempt, with the entered credentials supplied to the backend's next `ask-password` request;
- rejected credentials return to the dialog with the failure message and prior username/domain;
- cancelling restores the prior navigation state.

This keeps credential correction possible without any automatic retry loop. I verified both the early-failure prompt and a successful login locally; `./scripts/check.sh` passes all 179 tests.

@mrocha-montes, a retest against the original SMB server would still be appreciated to confirm its backend behaves the same way.

mrocha-montes:
Retested against the same SMB server (192.168.1.220) that reproduced the original sub-250ms failure. Both scenarios confirmed working as expected:

1. Backend fails before any click: stable login dialog, no flicker/reopen.
2. Incorrect credentials followed by correct ones: retry prompt reopens with the failure message, then connects normally on the correct submission.

Nice fix, thanks for tracking it down.

SjB:
I have a small changeset to add pin support to smb network filesystem at https://github.com/SjB/strata/tree/smb-pin-support mostly generated by Luna.

l0gicgate:
@SjB let's raise a separate PR for it so I fully understand the context of your changes

PR #37, opened 2026-08-31
fix(preview): allow high-resolution video previews

### LINKED ISSUE #33: MKV Previews Fail
# Description

MKV video previews fail. Since adding sandboxed previews, we introduced a regression I believe.

kallupx:
## Summary

- let full media previews use the existing 12-second wall timeout instead of the process-wide CPU-time limit
- retain the 10-second CPU limit for image, PDF, and thumbnail rendering
- document the operation-specific limits and bounded PNG/WebM outputs
- cover the media exception with a regression test

## Testing

- `cargo fmt --check`
- `cargo test` (129 passed)
- rendered a 1440p MKV preview in about 6 seconds within the sandbox wall limit (Ryzen 9 7900X3D)
- rendered a 4K MKV preview in about 9 seconds within the sandbox wall limit (Ryzen 9 7900X3D)

Closes #33

### DIFF
--- a/docs/preview-sandbox.md
+++ b/docs/preview-sandbox.md
@@ -21,6 +21,6 @@ Strata starts its own executable in a bubblewrap sandbox. The sandbox has:
 - read-only access to `/usr`, required runtime libraries and font/ImageMagick configuration, the Strata executable, and exactly one canonicalized input file;
 - writable access only to private mode-0700 output and temporary directories;
 - an empty environment with a nonexistent home directory;
-- 1.25 GB address-space, 10-second CPU, 32 MB file-size, and 12-second wall-clock limits.
+- 1.25 GB address-space, 32 MB file-size, and 12-second wall-clock limits, plus a 10-second CPU limit for image, PDF, and thumbnail rendering.
 
-The parent accepts only a bounded PNG result. Cancellation or timeout kills bubblewrap, which is the sandbox PID-namespace init process, and therefore tears down all helper descendants. A missing bubblewrap installation, renderer crash, malformed result, timeout, or permission failure is fail-closed and produces the normal fallback icon or **Preview unavailable** message.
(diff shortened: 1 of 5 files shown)


l0gicgate:
I pushed `abdfeff` to address the remaining wall-time failure found during review.

The original change correctly removes `RLIMIT_CPU` for media because FFmpeg's worker-thread CPU time is cumulative, but a shared 12-second wall limit was still too short. On the reported 1080×1920 H.264 MKV at 60 fps, the sandboxed software transcode took about 13–15 seconds on a Ryzen 9 9950X and was therefore still rejected.

This now uses an operation-specific policy:

- image, PDF, and thumbnail rendering retain the 12-second wall / 10-second CPU limits;
- full media previews use a 30-second wall limit without a cumulative CPU limit;
- media output remains bounded by the 1.25 GiB address-space and 32 MiB file-size limits.

A longer timeout alone would only move the hardware-dependent failure threshold, so the software fallback workload is now bounded too:

- first 30 seconds;
- at most 1280 pixels on either axis, preserving aspect ratio;
- at most 30 fps;
- existing two-thread realtime VP8 profile.

For the reproducer, that changes the normalized output from 1080×1920 at 60 fps to 720×1280 at 30 fps and reduced sandbox completion time from about 13 seconds to about 4 seconds while retaining the full 30-second preview segment.

Tests now cover both the operation-specific timeout and the FFmpeg duration/dimension/frame-rate bounds. I also verified the actual bubblewrap path with the reproducer, not only the generated command.

Validation:

- `cargo fmt --check`
- `cargo test` (130 passed on this branch)
- sandboxed reproducer: 4 seconds, 720×1280, 30 fps, 30.008-second WebM

Issue #38 can build hardware acceleration on top of this bounded software baseline. Hardware acceleration should improve the common path, while this fallback remains necessary for systems, codecs, or drivers where VA-API/Vulkan is unavailable.

PR #39, opened 2026-08-31
fix(storage): make managed writes atomic and symlink-safe

### LINKED ISSUE #15: Use atomic, symlink-safe writes for settings, bookmarks, themes, and caches
## Integrity/hardening concern

Several application-managed files are written directly with `fs::write`. Direct writes can leave truncated files after crashes or disk-full errors and follow a pre-existing final-component symlink.

Relevant locations include:
- `src/assets.rs` bundled-font cache writes;
- `src/ui/theme.rs` settings, custom theme, and source-style writes;
- `src/ui/window.rs` GTK bookmark writes.

## Proposed work

- Introduce a shared atomic-write helper:
  - create a unique temporary file in the same directory;
  - use restrictive permissions;
  - write and flush successfully;
  - atomically rename into place;
  - reject unsafe symlink/non-regular-file destinations where appropriate.
- Preserve the previous valid file when serialization or writing fails.
- Consider directory ownership/permissions before writing application-managed cache/config data.

## Acceptance criteria

- Interrupted or failed writes preserve the previous valid file.
- Application-managed writes do not follow a substituted final-component symlink.
- Newly created private configuration/cache files use appropriate permissions.
- Tests cover disk/write failure and symlink destinations.

kallupx:
## Summary

- add a standard-library atomic-write helper using short, unique temporary names in the destination directory
- write, flush, and sync before renaming; use private `0600` permissions and clean up incomplete temporary files
- reject symlink and non-regular final destinations
- migrate settings, custom themes, GTK bookmarks, bundled-font caches, and source-style caches

## Tests

- cover partial-write failure while preserving the previous file
- cover symlink and non-regular destinations
- cover private permissions and long destination filenames

## Verification

- `cargo fmt --all --check`
- `cargo check --all-targets --all-features`
- `cargo test --all-targets --all-features` — 149 passed

Closes #15

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -246,13 +246,12 @@ fn recolor_icon_source(source: &str, color: &str) -> String {
 }
 
 fn write_if_changed(path: &Path, contents: &[u8]) -> io::Result<()> {
-    let is_current = path
-        .metadata()
-        .map(|metadata| metadata.len() == contents.len() as u64)
+    let is_current = fs::symlink_metadata(path)
+        .map(|metadata| metadata.file_type().is_file() && metadata.len() == contents.len() as u64)
         .unwrap_or(false);
 
     if !is_current {
-        fs::write(path, contents)?;
+        crate::storage::atomic_write(path, contents)?;
     }
 
     Ok(())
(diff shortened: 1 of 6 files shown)


l0gicgate (approved):
Awesome stuff @kallupx, thank you for this contribution

PR #53, opened 2026-08-31
feat(browser): add Empty Trash button to header bar

JoeJoeflyn:
## Summary
- The Empty Trash action was only reachable via right-click on the Trash sidebar entry, which is not discoverable when browsing `trash:///`
- Add a danger-styled button to the header bar that appears only when the active location is `trash:///`
- The button triggers the existing `confirm_empty_trash` flow (same dialog as the sidebar context menu)

## Testing
- `cargo test --release` — 155 passed, 0 failed
- New test `empty_trash_button_only_shows_at_trash_root` covers:
  - `trash:///` → visible
  - `trash:///folder` → hidden (not trash root)
  - local path → hidden
  - `None` → hidden
  
  
image


l0gicgate:
@JoeJoeflyn thank you for this contribution, instead, what we should do is add it to the pane's header.
image

Also, we should make sure we add it to the other 2 modes (Grid and Explorer)'s pane headers too:
image
image

JoeJoeflyn:
Thanks for the feedback! I moved the button to the pane header in all three modes
image
image
image

l0gicgate:
I will merge this in later today after I'm done work, I pull all UI changes down from contributors to test manually. This will be in tonight's release! Thank you

l0gicgate:
Applied the final UI fixes:

- Resolved conflicts with `main` and fixed formatting/lint.
- Removed the duplicate Empty Trash action from the window header.
- Kept Explorer’s pane header independent from the resizable table.
- Fixed Explorer cells to truncate consistently when columns are resized.

All 204 tests pass.

PR #55, opened 2026-08-31
fix(preview): size the sandbox file limit for decoded frames

### LINKED ISSUE #54: Thumbnails fail for images above ~11 MP: sandbox --fsize kills the glycin loader
Previews and thumbnails fail with "The sandboxed preview renderer failed" for any image above roughly 11 megapixels, which covers most phone photos. Each attempt also leaves a core dump behind.

## Cause

`src/sandbox.rs:235` passes `--fsize=33554432` (32 MiB) to `prlimit`, mirroring `MAX_OUTPUT_BYTES`. The intent is to cap the output PNG, but `RLIMIT_FSIZE` is process-wide and applies to every file any descendant writes.

Since gdk-pixbuf 2.43.2, `Pixbuf::from_file_at_scale` (`render_pixbuf` in `src/sandbox_helper.rs`) routes decoding through glycin, which spawns `glycin-image-rs` and `ftruncate`s a memfd to `width * height * channels` to hold the full-resolution frame before scaling. For a 3024x4032 JPEG that is 36,578,304 bytes, over the 32 MiB cap, so the kernel kills the loader with SIGXFSZ.

The requested thumbnail size does not help, since the full frame is materialised first. The threshold is `W * H * 3 > 32 MiB`, about 11.2 MP, or about 8.4 MP when an RGBA format is selected.

## Reproduction

```
prlimit --as=1342177280 --cpu=10 --fsize=33554432 -- \
  glycin-thumbnailer -i file:///path/to/12MP.jpg -o out.png -s 256
# Glycin Thumbnailer: Remote error: org.freedesktop.zbus.Error: i/o error
```

Raising only `--fsize` to 512 MiB makes the same command succeed, so `--as` and `--cpu` are not involved. `coredumpctl` shows `glycin-image-rs` terminated with signal 25 (XFSZ), crashing in `SharedMemory::new` -> `nix::unistd::ftruncate`.

## Affected systems

Any distribution whose gdk-pixbuf is built against glycin: Arch (2.44.7), Fedora (2.44.5), Debian trixie/sid, Ubuntu 25.04+. It does not reproduce on gdk-pixbuf 2.42.x, which decodes in-process. That includes `ubuntu-24.04`, which CI runs on, so CI cannot catch this today. `src/sandbox/tests.rs:32` asserts the current value, which locks it in.

## Suggested fix

Decouple the two caps. `--as=1342177280` is already the memory guard, and `MAX_OUTPUT_BYTES` is enforced host-side after the run, so `--fsize` can be raised well above any plausible decoded frame.

spandan11106:
Fixes #54.

`--fsize=33554432` mirrored `MAX_OUTPUT_BYTES` in order to cap the output PNG, but `RLIMIT_FSIZE` is process-wide and also covers the memfd glycin sizes to `width * height * channels` while decoding. Since gdk-pixbuf routes decoding through glycin, any image over roughly 11 MP made the loader exceed the limit and get killed with SIGXFSZ, so previews and thumbnails failed for most phone photos and left a core dump behind.

This decouples the two caps:

- `FILE_SIZE_LIMIT_BYTES` is now 512 MiB, comfortably above a full-resolution frame while staying under `ADDRESS_SPACE_LIMIT_BYTES`, which remains the real memory guard.
- `MAX_OUTPUT_BYTES` is unchanged and still bounds the output host-side after the run, which is where that check belongs.

## Testing

- `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --all-targets --all-features`.
- Verified against the underlying loader directly: with the old limit, `prlimit --as=2147483648 --cpu=10 --fsize=33554432 -- glycin-thumbnailer -i file://<12MP.jpg> -o out.png -s 256` kills the loader, and raising only `--fsize` lets the same command succeed.

Note that CI runs on `ubuntu-24.04`, whose gdk-pixbuf 2.42.10 predates the glycin loader and decodes in-process, so it cannot exercise this path. The added unit test pins the invariant that the limit must clear a full-resolution decoded frame, to stop the value being tightened back into the same failure.


src/sandbox/tests.rs
```diff
@@ -29,7 +67,7 @@ fn sandbox_exposes_only_runtime_input_and_private_output() {
     assert!(joined.contains("--bind /tmp/private-output /output"));
     assert!(joined.contains("--as=2147483648"));
     assert!(joined.contains("--cpu=10"));
-    assert!(joined.contains("--fsize=33554432"));
+    assert!(joined.contains("--fsize=536870912"));
```
l0gicgate, line 70: > a 2 GB address-space limit, allowing modern image loaders to start their isolated worker threads, and a 32 MB file-size limit;

Please also update the preview sandbox docs. We're changing this from 32mb to 512mb which would leave a stale fact in there.

src/sandbox.rs
```diff
@@ -15,6 +15,13 @@ use std::{
 const WALL_TIME_LIMIT: Duration = Duration::from_secs(12);
 const MEDIA_WALL_TIME_LIMIT: Duration = Duration::from_secs(30);
 const ADDRESS_SPACE_LIMIT_BYTES: u64 = 2 * 1024 * 1024 * 1024;
+// RLIMIT_FSIZE applies to every file the sandbox writes, including the memfd
+// glycin sizes to `width * height * channels` to hold a full-resolution frame
+// while decoding. gdk-pixbuf routes decoding through glycin, so a limit tuned
+// to the output PNG kills the loader with SIGXFSZ on any ordinary phone photo.
+// ADDRESS_SPACE_LIMIT_BYTES remains the memory guard, and MAX_OUTPUT_BYTES
+// bounds the output we are willing to read back.
```
l0gicgate: Claude is so verbose lol, I think we can just remove this comment. The variable name is explicit enough.
```suggestion

l0gicgate (changes requested):

PR #56, opened 2026-08-31
fix(startup): fall back when GVFS is unresponsive

### LINKED ISSUE #52: App hangs indefinitely with no UI when gvfsd is unresponsive
## Bug

When `gvfsd` is hung/unresponsive, Strata hangs indefinitely during startup with no window appearing and no error message. The process stays alive but never shows a UI.

## Reproduction

1. Kill or hang `gvfsd`: `killall gvfsd`
2. Launch strata
3. Observe: process starts, `activate` is called, `window.present()` is reached, but the window never appears on screen
4. stderr shows repeated `(strata:PID): GVFS-WARNING **: Error: Timeout was reached` every ~25 seconds
5. The app never recovers — it waits forever on GIO calls

## Expected behavior

Strata should either:

1. **Fall back to direct `std::fs` access for local paths** when GVFS is unresponsive (local files don't need a virtual filesystem daemon), or
2. **Timeout after 2-3 seconds** and show an error state ("Filesystem daemon unresponsive") with a retry option, or
3. **At minimum**, show a loading/error window instead of silently hanging with no UI

## Actual behavior

The app hangs silently. No window, no error, no timeout. A non-technical user would assume the app is broken and force-quit it.

## Environment

- Arch Linux / Omarchy / Hyprland (Wayland)
- GTK 4.22.4
- gvfs 1.60.2-1
- Rust 1.98.0
- Strata 0.4.0 (built from `main`, commit 035d6d6)

## Workaround

Restarting `gvfsd` before launching strata fixes the issue:
```bash
killall gvfsd && /usr/lib/gvfsd &
```

But this is not something a non-technical user would know to do.

## Root cause

Strata uses GIO/GVFS for file operations. When `gvfsd` is unresponsive, every GIO call blocks indefinitely with no timeout. This affects window creation because the initial directory load happens during `BrowserView::new()`, which blocks the GTK main loop.

Local file access via `std::fs` would not have this dependency — GVFS is only needed for virtual locations (trash://, network mounts, etc.), not for browsing local directories.

kallupx:
## Summary

- probe GVFS in an isolated subprocess before GTK initializes
- fall back to GIO's local filesystem backend when the two-second probe times out
- preserve full GVFS and volume support when the daemon is responsive

## Testing

- `./scripts/check.sh`
- suspended private `gvfsd`: window opens in about 2.3 seconds instead of 50.3 seconds
- responsive private `gvfsd`: full backend retained and window presented normally

Closes #52

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,9 +12,13 @@ mod services;
 mod storage;
 mod ui;
 
+use std::{process::Stdio, time::Duration};
+
 use gtk::{gio, prelude::*};
 
 const APPLICATION_ID: &str = "io.github.lgse.Strata";
+const GVFS_PROBE_ARGUMENT: &str = "--gvfs-probe";
+const GVFS_PROBE_TIMEOUT: Duration = Duration::from_secs(2);
 
 fn main() -> gtk::glib::ExitCode {
     let arguments: Vec<_> = std::env::args().collect();
@@ -28,6 +32,16 @@ fn main() -> gtk::glib::ExitCode {
         }
         return gtk::glib::ExitCode::SUCCESS;
     }
+    if arguments
+        .get(1)
+        .is_some_and(|value| value == GVFS_PROBE_ARGUMENT)
+    {
+        let _vfs = gio::Vfs::default();
+        let _volumes = gio::VolumeMonitor::get();
+        return gtk::glib::ExitCode::SUCCESS;
+    }
+
+    fall_back_if_gvfs_is_unresponsive();
 
     metrics::initialize();
     if let Err(error) = tracing_subscriber::fmt::try_init() {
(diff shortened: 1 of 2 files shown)


l0gicgate:
Tested locally on Arch Linux / Omarchy / Hyprland:

- Ran `./scripts/check.sh` successfully (formatting, compilation, Clippy, and all 205 tests passed; `cargo-deny` and `typos` were not installed and were skipped).
- Built the PR branch with `cargo build`.
- Suspended the active `gvfsd` with `SIGSTOP`, with an automatic `SIGCONT` recovery after 8 seconds, then launched `target/debug/strata`.
- Confirmed the fallback was activated from stderr: `GVFS is unresponsive; using local filesystem support for this session.`
- Confirmed Strata presented its window promptly (roughly the 2-second probe timeout plus 211 ms measured from application initialization to presentation), instead of hanging indefinitely.
- Navigated through multiple local directories while running on the local VFS fallback; directory loads completed successfully and the UI remained responsive.
- Confirmed the real `gvfsd` resumed normally after the test.

Result: the startup fallback and local filesystem navigation work as intended when GVFS is unresponsive.

PR #58, opened 2026-08-31
fix(navigation): complete keyboard navigation flow

### LINKED ISSUE #46: Improve keyboard navigation flow for nested folders
Navigating nested folders feels somewhat mouse-dependent.
Keyboard navigation could be smoother when moving through deeper folder hierarchies.
Closing just one level of a nested folder using keyboard shortcuts isn't very intuitive.
In the delete confirmation dialog, switching focus between Cancel and Delete can feel awkward at times.
Better highlight on which level of nested folder is currently under navigation.

kallupx:
## Summary

This extends the existing file-list keyboard controls into a complete focus flow for column headers, nested folders, the sidebar, menus, and delete confirmation. Existing item-level `h/j/k/l` and arrow behavior remains unchanged.

### Nested columns

- `Enter`, `l`, or Right on a folder opens its child column and selects the first item, allowing another press to continue immediately.
- Activating a folder whose child column is already open moves focus into that column instead of doing nothing.
- Backspace closes the focused nested column and its descendants. From the root column, it navigates to the filesystem parent.
- Returning to a parent or already-open child column restores that column's selection.

### Column headers

- `k` or Up from the first visible file moves focus into the column-header controls.
- `h/l` and Left/Right move between the sort-direction, sort-menu, filter, and close controls.
- `j` or Down returns focus to the selected file.
- `h` or Left from the first header control continues outward to the parent column or sidebar.
- Enter and Space activate the focused header control instead of activating the previously selected file.

### Sidebar

- `h` or Left from the first file column focuses the active sidebar row.
- `l` or Right returns from the sidebar to the active file column.
- Sidebar rows now receive native Up/Down and Enter handling instead of those keys affecting the file list.
- `j/k` mirror native Down/Up navigation within the sidebar.
- `Ctrl+Shift+B` moves focus to the sidebar and restores the exact previous control when pressed again.

### Filters, sorting, and deletion

- Backspace closes an empty focused filter while preserving normal text deletion when the filter contains text.
- The Sort By dropdown supports `h/j/k/l`, and Backspace closes it.
- Delete confirmation supports `h`/Left for Cancel and `l`/Right for Confirm.
- The initially selected delete action now receives a reliable visible focus ring.

## Focus presentation

- The active column header receives a complete accent border.
- Keyboard focus rings are shown consistently on header controls.
- Background directory batches, sorting, and filesystem refreshes no longer steal focus from controls.

## Testing

- `./scripts/check.sh` — 207 tests passed
- `git diff --check`
- Manually validated nested-column, header, sidebar, filter, sorting, and delete-dialog keyboard flows
- Manually validated focus visibility and persistence during directory loading and refreshes

Closes #46


kallupx:
While rebasing I noticed that PR #53 added a hidden Empty Trash button as the first header action, which could break k/Up navigation into the header.

Fixed by focusing the first visible header action, restoring file-list focus after trash dialogs, and adding hjkl navigation to the Empty Trash dialog. All tests pass.

PR #60, opened 2026-08-31
perf(startup): defer noncritical work and split preview helper

### LINKED ISSUE #59: perf: reduce startup cost on low-end hardware
## Problem

Strata takes noticeably long to open on lower-end laptops. The main process eagerly loads preview-only dependencies and constructs hidden settings UI on the startup path. Development and release packaging also need to keep the preview renderer available after separating it.

## Scope

- Move untrusted preview rendering into a sibling helper binary so preview-only libraries are not loaded by the main executable.
- Defer hidden settings construction until after the critical startup path while preserving update checks.
- Keep settings pages scrollable in compact windows.
- Build, package, update, and locally install both binaries.
- Add one-step development and local release-install commands.
- Record and compare startup, linker, and memory measurements.

## Acceptance criteria

- Image, PDF, and media previews continue to run through Bubblewrap.
- Release archives and in-place updates contain both executables.
- Startup time/dependency count and initial memory use improve.
- Formatting, Clippy, and all tests pass.

l0gicgate:
## Summary

- split preview rendering into a sibling `strata-preview-helper` executable so the GUI avoids loading Poppler and helper-only code at startup
- construct settings after the critical startup path while preserving automatic update checks, and fix compact-window settings scrolling
- package and atomically update both executables, with clear errors when the helper is missing
- add `make run-dev` and `make install-local` workflows that always build/install both binaries

## Performance

Measurements on the same low-end laptop, comparing the installed build with this release build:

- main-process shared-library dependencies: 144 → 114
- dynamic-loader startup: ~6.6M → ~4.4M cycles
- initial RSS: ~89 MiB → ~78 MiB
- median window presentation: 212 ms → 198 ms
- preview-helper hot launch: ~33 ms → ~12 ms

Directory IO remains variable and dominates portions of time-to-first-directory-render.

## Testing

- `./scripts/check.sh`
- `cargo build --release --bins`
- manual image preview through the Bubblewrap helper
- manual settings scrolling at compact window sizes

Closes #59


l0gicgate:
Closing for now, will revisit later

PR #67, opened 2026-08-31
feat(ui): enable pin action in file properties and add navigate_location

SjB:
Allow SMB filesystem to be pin. If the pin filesystem is not mounted it will auto mount the filesystem.


l0gicgate:
Pushed a follow-up fix in `d91fac3` addressing the review findings:

- Properties now reports the real pinned state instead of always showing `No`.
- The Pin action is disabled for already-pinned and standard sidebar locations, avoiding silent no-op clicks.
- The item context-menu Pin action uses the same availability state.
- Added coverage confirming sidebar navigation validates remote URIs (so unmounted SMB shares enter the mount flow) while native bookmarks still navigate directly.

Validation completed:
- `cargo test --all-targets` (217 passed)
- `git diff --check`

l0gicgate:
Follow-up pushed in `f98c6fe` for the Properties dialog still showing `PINNED: No`.

The status callback previously depended on upgrading a weak `SidebarState`; when that state was unavailable it fell back to `Unavailable`, which rendered `No` and disabled Pin. The pinned-place collection is now shared directly with the status callback, so the dialog reads the same live collection used to build the PINNED sidebar section. I also added a regression test covering pinned SMB locations, available locations, and standard locations.

Validation: 218 tests passed; clippy with `-D warnings` passed; `git diff --check` passed.

l0gicgate:
@SjB thank you for your contribution! It will be in tonight's release.

PR #72, opened 2026-08-31
perf: bound search indexing and trash summarization on very large trees

### LINKED ISSUE #13: Bound search indexing and trash summarization on very large trees
## Availability problem

Search indexing retains an entry for the complete tree, and trash summarization recursively walks all entries. Adversarial or unexpectedly large trees can cause excessive memory usage, I/O, and long-running work.

Relevant code:
- `src/services/search.rs`
- `src/ui/browser.rs:4108-4175`

## Proposed work

- Establish explicit entry, memory, depth, and runtime budgets.
- Keep search result storage bounded and consider a disk-backed/incremental index for large trees.
- Make indexing and trash measurement cancellable end-to-end.
- Detect cycles/backend anomalies and cap recursive depth.
- Surface partial/estimated results rather than blocking indefinitely.

## Acceptance criteria

- Search and trash measurement remain responsive on very large synthetic trees.
- Memory use has a documented bound or graceful degradation behavior.
- Cancellation promptly stops traversal.
- Tests cover deep, broad, inaccessible, and changing trees.

pdwaldrop:
## Description

Search indexing and trash summarization each walked (and, for search, retained) an entire tree with no bound, so an adversarial or unexpectedly large tree could cause excessive memory use, I/O, or long-running work.

- Add entry, depth, and wall-clock time budgets to search indexing (`src/services/search.rs`), and surface a `truncated` flag through to the search UI (a footer hint) when a budget is hit.
- Add the same budget shape to trash summarization (`src/ui/browser.rs`), surfaced via the empty-trash confirmation dialog and the trash-root properties size ("At least N items" / "≥ size").
- Make the confirmation-dialog measurement fully decoupled from deletion: `summarize_trash` never retains an entry list and stays strictly bounded regardless of trash size, while "Empty Trash" does its own fresh, unbounded, non-recursive top-level listing at the moment it's clicked, so it always empties everything and can't go stale.
- Make trash measurement cancellable end-to-end: a `LoadHandle` (aborts a superseded request) plus a "Measuring Trash…" indicator with Cancel/Escape while the walk runs.
- Degrade gracefully instead of failing outright when a trash subdirectory becomes unreadable or disappears mid-walk: treat it as a truncated branch rather than aborting the whole measurement.
- Add `tracing` instrumentation for truncation, completion, and cancellation, matching the existing convention in `adapters/local_files.rs`.

Deliberately left out: a disk-backed/incremental index for search, which the issue marks as "consider" rather than required, and which is a large enough change to warrant its own discussion first.

## Visual evidence

N/A for the primary logic (bounding/cancellation of background work has no visual surface). Two small UI additions exist — a "Measuring Trash…" loading indicator with Cancel, and a search-results footer hint shown only once the entry/time budget is hit on an adversarially large tree — but I don't have a way to drive this native GTK app's mouse input to capture that specific dialog (no click-simulation tooling available in my environment). I did manually verify the surrounding trash browsing flow end-to-end: navigated to `trash:///`, populated it with real files, confirmed they render correctly, then emptied it.

## How to test

1. `./scripts/generate-fixture.sh target/fixtures` to build a large synthetic tree, then open Strata's search on `target/fixtures` and type a query. Search remains responsive; on an even larger tree (or with the budget constants temporarily lowered) a "search limited" hint appears in the footer.
2. Move a large number of files/folders to Trash, then use the sidebar's Trash entry or the "Empty Trash" action. The confirmation dialog appears after a bounded measurement (never longer than ~5s), shows "At least N items" if truncated, and "Empty Trash" always deletes everything regardless of how large the trash is.

**Expected result:** search and trash summarization stay responsive and bounded on very large or adversarial trees, trash deletion is always complete, and none of this blocks the UI or the confirmation dialog from appearing.

## Related issue

Closes #13


src/ui/browser.rs
```diff
@@ -4553,65 +4695,159 @@ async fn summarize_trash(root: &gio::File) -> Result<TrashSummary, glib::Error>
...
             .next_files_future(64, glib::Priority::DEFAULT)
             .await?;
         if children.is_empty() {
             break;
         }
         for info in children {
+            // Checked here, not just when a directory decides whether to recurse: without this,
+            // a trash root holding an enormous number of top-level *files* (no directories to
+            // recurse into) would never trip the budget at all.
+            if visited.get() >= max_entries || Instant::now() >= deadline {
```
kallupx, line 5378: This truncates summary.entries, but “Empty Trash” later deletes only those entries. If the budget expires before every top-level item is visited, the action leaves unmeasured items behind. It would be a good idea to decouple bounded measurement from the deletion worklist

src/ui/browser/tests.rs
```diff
@@ -2,6 +2,11 @@
 
 use super::*;
 
+/// Serializes tests that drive `glib::MainContext::default()` directly, since it is a
+/// process-wide singleton and concurrent access from the test harness's per-test threads panics
+/// with a GLib thread-affinity error. Mirrors `ASYNC_FILE_TEST` in `adapters::local_operations`.
+static ASYNC_FILE_TEST: std::sync::Mutex<()> = std::sync::Mutex::new(());
```
kallupx: This is a separate mutex from the one in local_operations, so the tests can still concurrently use MainContext::default(). On this head, cargo test aborts with a GLib thread-affinity error while the base passes. Please use a private MainContext or one shared lock.

src/services/search.rs
```diff
@@ -13,6 +13,18 @@ use std::{
 const RESULT_LIMIT: usize = 100;
 const PUBLISH_INTERVAL: Duration = Duration::from_millis(50);
 
+/// Caps how many entries a single index will retain, bounding worst-case memory use on
+/// adversarially large or unbounded trees (for example a runaway bind mount or `/proc`).
+/// Each `SearchItem` stores a `PathBuf` plus three `String`s (name, lowercased name, lowercased
+/// relative path), roughly 300-700 bytes per entry including heap data for typical to very long
+/// paths, so this cap bounds worst-case index memory to roughly 60-140 MB.
```
kallupx: I don't think this is a 60–140 MB worst-case bound because paths and lowercase copies are variable-sized. With roughly 4 KiB paths, 200,000 entries can exceed 1.6 GB before overhead. Or maybe I am missing something 😄

src/services/search.rs
```diff
@@ -64,14 +95,15 @@ pub fn index_tree(root: PathBuf) -> (SearchHandle, Receiver<SearchEvent>) {
...
-            let mut matches = Vec::<(i64, SearchItem)>::new();
+            let mut progress = WalkProgress::default();
             let mut last_publish = Instant::now();
+            let walk_start = Instant::now();
             let walker = ignore::WalkBuilder::new(&root)
                 .hidden(true)
                 .follow_links(false)
                 .standard_filters(true)
                 .require_git(false)
+                .max_depth(Some(max_depth))
```
kallupx: truncated is set only for entry and time limits. The depth cap and discarded walker errors also omit parts of the tree, so the UI reports those indexes as complete. Please mark these cases as truncated too.

src/ui/browser.rs
```diff
@@ -4553,65 +4695,159 @@ async fn summarize_trash(root: &gio::File) -> Result<TrashSummary, glib::Error>
...
+                deadline,
+                max_entries,
+                max_depth,
+            )
+            .await?;
+            count = count.saturating_add(child_count);
+            size = size.saturating_add(child_size);
+            truncated |= child_truncated;
+        }
+        if truncated {
```
kallupx: this stops enumerating unrelated siblings when any child is depth or error-truncated. I reproduced 80 sibling directories being counted as only 65. you could separate global-budget exhaustion from branch-local truncation.

kallupx:
Hey, this looks pretty good, I added some comments to get rid of some unwanted behaviour. You could also revisit some of the comments, they are pretty verbose or just restate what the code is doing, which is a bit unnecessary.

pdwaldrop:
@kallupx Thanks, I want to contribute but by no means an expert! I'll work on this feedback.

src/ui/browser.rs
pdwaldrop: Good catch, thank you — you're right that this could silently leave real items in the trash. Fixed in 83414ec: every top-level entry is always listed here now regardless of budget; only the recursive measurement of an entry's own contents is skipped once the budget is spent, and it's still counted as one unmeasured item rather than dropped from the list. Added a test that checks `entries.len()` stays complete even past the budget.

src/ui/browser/tests.rs
pdwaldrop: You're right — two separate mutexes don't actually coordinate with each other, so they don't prevent the cross-thread race at all. Rather than share one lock, I switched every test in this file to a private `glib::MainContext::new()` instead of the shared `default()`, which removes the race at its root instead of coordinating around it. 8 repeated full-suite runs have been clean since (83414ec).

src/services/search.rs
pdwaldrop: You're not missing anything — that was just a bad estimate on my part, I didn't account for path length scaling into the total. Fixed the comment to reflect the real worst case: ~60-140 MB for typical paths, but up to ~1.5-2 GB for paths near PATH_MAX, which matches your math. Thanks for checking the numbers 😄

src/services/search.rs
pdwaldrop: Fixed — the walker now goes one level past max_depth so a depth cutoff can actually be detected instead of silently pruned, and a discarded walker error (e.g. permission-denied) is now treated as truncation too instead of being dropped. Added tests for both cases.

src/ui/browser.rs
pdwaldrop: Reproduced your exact numbers before fixing this — 80 siblings undercounted to 65, matching what you found. Root cause was exactly what you described: the loop was checking the aggregate truncated flag (which a branch-local depth or error truncation also sets), instead of the actual shared budget. Fixed to check the real global-budget condition, and added a regression test built on your 80/65 repro to lock it in.

pdwaldrop:
Done — trimmed the verbose ones and cut anything that just restated the code (5450c46). Also pushed fixes for all five review comments above; thanks for the thorough look, especially the 80/65 repro, that was a real one.

src/ui/browser.rs
kallupx: This fixes incomplete “Empty Trash,” but after the budget expires it still reads and retains every top-level entry, making time and memory unbounded. I think the measurement could remain bounded, then have the deletion job enumerate and delete entries incrementally after confirmation(?)

kallupx:
Great work, @pdwaldrop the previous issues look fixed. There is still a remaining with trash deletion: [relevant line](https://github.com/lgse/strata/pull/72/changes#diff-ee2df4fdf8524e86530332774aa1fe446fa49c18dff4c9372d387c0f0151739dR4705)

src/ui/browser.rs
pdwaldrop: Fixed properly this time — that was a good catch, my fix just moved the problem. Decoupled the two per your suggestion: summarize_trash no longer builds or retains any entries list at all, so measurement stays strictly bounded regardless of trash size. Empty Trash now enumerates the trash root fresh (a new list_trash_top_level_entries, no budget, non-recursive so it's cheap even for many entries) at the moment it's actually clicked. Bonus: this also fixes a staleness bug, since deletion now reflects trash's actual contents at confirm time instead of whatever was there when the dialog opened. (83e52a8)

l0gicgate:

src/ui/browser.rs
kallupx: Thanks, the measurement is now separated from deletion. The original issue is still there though: measurement continues fetching root batches after its budget expires, and deletion still collects every entry into a Vec before starting. 

Stop the root enumeration when the measurement budget is reached. For deletion, fetch one bounded batch, delete and discard those entries, then fetch the next batch until there are no more entries to enumerate. Keep only the current batch and bounded progress/error information and don’t accumulate deleted locations in another collection. This should complete the deletion without ever retaining the full list

pdwaldrop: Fixed properly this time. Two changes:

- `summarize_trash_with_budget` now breaks out of the root-enumeration loop the moment the shared budget (entries or time) is spent, mirroring the stop condition `enumerate_trash_directory` already uses for its own batches — no more requesting further `next_files_future` batches once truncated.
- "Empty Trash" now streams: a new `empty_trash()` enumerates and deletes the trash root one 64-item batch at a time, discarding each batch as it goes and keeping only a running count plus a capped (8-message) error list — never a full per-entry list, regardless of trash size.

(6d26926)

pdwaldrop:
Done — rebased onto the rewritten `main` and force-pushed. Confirmed none of this branch's own commits carried a Claude attribution themselves; the two that did (SMB support, install-updates-from-settings) were only present via the two merge-conflict merge commits pulling in older `main` history, and are now replaced by their rewritten versions from the new `main`.

The rebase surfaced a real structural conflict along the way: `main` had meanwhile standardized dialog construction onto a shared `message_dialog_layout`/`ModalLayout` helper (#91), which this branch's trash dialogs predate. Resolved by porting this branch's bounded/cancellable trash-summary logic onto that shared helper rather than keeping the old ad-hoc widget code, and picked up a `message_dialog_layout` caller — `feat(archive)` (#81) — that needed the same cancel-handler parameterization this branch added.

Build, clippy (`-D warnings`), fmt, and the full test suite (247 tests) all pass on the rebased branch.

src/ui/browser.rs
kallupx: Looks good to me :)

kallupx (approved):

pdwaldrop:
@kallupx What happens now?

l0gicgate:
@pdwaldrop, I'm testing this now. I'll take it over and get it over the line

l0gicgate:
## Manual testing results

Completed an end-to-end manual pass with normal, deep, broad, inaccessible, and changing fixtures.

### Search

- Normal search returned `quarterly-report.txt` without a truncation warning; Enter closed search, navigated to the parent folder, and selected/previewed the result.
- A corrected 200,580-file fixture remained responsive through repeated open/cancel/reopen cycles and continuous typing.
- The entry budget stopped indexing in roughly 3–4 seconds and showed the “Search limited…” hint.
- An 80-level fixture omitted `marker-079` beyond depth 64, showed the limited-results hint, and remained responsive.

### Trash

- Exact measurement reported 4 items / 15 B without “At least”; Cancel preserved the fixture and confirmed deletion emptied Trash.
- An inaccessible branch degraded cleanly to “At least 563 items” / “at least 0 B” without an error.
- The 200,580-file fixture stayed responsive during bounded measurement, stopped at about five seconds, reported “At least 123777 items,” and cancelled promptly without a delayed confirmation.
- A 128-top-level-folder / 12,800-file fixture reported exactly 12,928 items, showed responsive progress, and emptied completely in bounded batches. It completed too quickly to cancel manually; the automated mid-flight cancellation coverage passes.

### UI refinements made during testing

- Added explanatory body copy to the measuring-trash dialog, vertically aligned its header, and yielded between measurement batches so GTK input/rendering stays responsive.
- Added a theme-accent search-indexing spinner, a higher-contrast footer legend, and a proper namespaced Lucide CornerDownLeft icon for the Enter/open hint.
- Let breadcrumbs use the full available header width while keeping them beside the sidebar toggle and preserving fixed header actions.

Testing also exposed a separate issue outside this PR’s bounded search/trash-measurement paths: opening or folder-peeking a directory containing approximately 200,000 entries can freeze ordinary browser rendering and consume substantial memory. I’m filing that separately.

Follow-up issue: #137

l0gicgate:
@pdwaldrop thank you for this contribution. Nice work!

PR #75, opened 2026-08-31
fix(operations): report partial cancellation results

### LINKED ISSUE #11: Report partial results accurately when destructive operations are cancelled
## Data-integrity problem

Cancelling a delete, restore, copy, or move cannot undo items that already completed, and recursive work may stop midway. The current task-abort model can make cancellation appear cleaner or more transactional than it is.

Relevant code:
- `src/app/browser.rs:710-725`
- `src/adapters/local_operations.rs:264-309`

## Proposed work

- Use explicit cancellables and define cancellation semantics for every operation type.
- Track and report completed, failed, and not-attempted items.
- Refresh every affected source and destination after cancellation.
- For recursive copy, remove only known-incomplete staging output; never remove pre-existing data.
- Make the progress UI state that cancellation may leave completed changes in place.
- Return a distinct cancelled/partially-completed event instead of silently dropping callbacks.

## Acceptance criteria

- Cancellation produces a final result describing partial completion.
- The UI does not claim that already completed destructive changes were reverted.
- Source/destination views converge to actual filesystem state.
- Tests cancel between items and during recursive copy/delete/restore.

kallupx:
Closes #11

## Summary

- replace task abortion with explicit GIO cancellation
- report completed, failed, and unattempted items after cancellation
- refresh affected roots and open descendants, including remote locations
- preserve completed move results and cut-clipboard state
- safely stage recursive copies and replacements
- move staging cleanup off the GTK main thread
- provide cancellable progress for transfers, deletion, and restoration

## Safety

- completed operations are not misreported as cancelled
- cancellation cleanup removes only Strata-owned staging data
- existing destinations are never removed during staging cleanup
- affected-location tracking remains linear for large directory trees

## Testing

- cancellation between items and during recursive operations
- cancellation/completion race handling
- partial move failures and clipboard preservation
- replacement source and target refresh
- remote descendant refresh
- `./scripts/check.sh` — 214 tests passed

### DIFF
--- a/src/adapters/local_files.rs
+++ b/src/adapters/local_files.rs
@@ -4,7 +4,7 @@ use std::{
     cell::{Cell, RefCell},
     collections::HashMap,
     io::ErrorKind,
-    path::PathBuf,
+    path::{Path, PathBuf},
     rc::Rc,
     time::{Duration, Instant},
 };
@@ -20,6 +20,7 @@ use crate::{
 };
 
 const ATTRIBUTES: &str = "standard::display-name,standard::name,standard::type,standard::is-hidden,standard::is-symlink,standard::size,time::modified";
+const MAX_PENDING_MONITOR_CHANGES: usize = 256;
 
 #[derive(Default)]
 pub struct LocalFileSource;
@@ -308,6 +309,9 @@ impl FileSource for LocalFileSource {
         let timeout_for_change = timeout.clone();
         let cancelled_for_change = cancelled.clone();
         monitor.connect_changed(move |_, file, other_file, event| {
+            if pending_for_change.borrow().contains_key(Path::new("")) {
+                return;
+            }
             let path = file.path();
             let other_path = other_file.and_then(gio::File::path);
             let change = match event {
@@ -341,13 +345,9 @@ impl FileSource for LocalFileSource {
                 PendingMonitorChange::Move { to, .. } => to.clone(),
                 PendingMonitorChange::Rescan => PathBuf::new(),
             };
-            pending_for_change
-                .borrow_mut()
-                .entry(key)
-                .and_modify(|pending| {
-                    *pending = merge_pending_change(pending.clone(), change.clone());
-                })
-                .or_insert(change);
+            if !queue_monitor_change(&mut pending_for_change.borrow_mut(), key, change) {
+                return;
+            }
 
             if let Some(source) = timeout_for_change.take() {
                 source.remove();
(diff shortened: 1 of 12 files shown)


l0gicgate:

l0gicgate:
## Implementation and manual verification update

PR #75 is up to date with `main` and now includes the cancellation-accounting fix plus the performance and reconciliation fixes found during staged manual testing.

### Testing methodology

We tested one operation at a time using large, uniquely named fixture sets so the on-disk state could be compared directly with Strata's reported totals. For each scenario we:

1. created or staged a known set of files/directories;
2. started the operation from Strata and cancelled it after measurable progress;
3. compared **completed**, **failed**, and **not attempted** counts in the result UI;
4. verified the source and destination contents on disk;
5. checked that affected panes refreshed to the same state;
6. checked responsiveness while progress, cancellation, context menus, and reconciliation were active;
7. checked for leaked Strata-owned staging artifacts (`.strata-*`).

The restore case used 3,000 uniquely named Trash entries and a dedicated restore destination. We also restarted the debug build between iterations and repeated scenarios after each performance/UI change rather than relying only on automated tests.

### Scenarios exercised

- recursive copy cancellation;
- same-filesystem bulk move cancellation;
- permanent recursive-delete cancellation;
- large Trash restore cancellation;
- large selection/context-menu behavior during these workflows;
- post-cancellation model and filesystem reconciliation.

### Confirmed behavior and fixes

- Cancellation now reports partial results accurately: completed work remains completed, failures remain failures, and untouched items are reported as not attempted.
- Completion wins a cancellation race when GIO has already completed the item.
- Recursive copy/replacement uses Strata-owned staging and cancellation removes only the incomplete staging output, preserving existing destinations.
- Completed move sources are removed from the cut clipboard while uncompleted sources remain available.
- Affected local/remote roots and open descendants are refreshed after partial operations.
- Copy cancellation left the expected partial destination state and no `.strata-*` artifacts.
- Same-filesystem move cancellation remained responsive and reconciled to the reported result.
- Permanent recursive-delete cancellation preserved the unfinished root and matched the reported partial state.
- Large restore cancellation matched the reported completed/not-attempted split at both the Trash and restore destination.
- Large operation progress no longer performs expensive per-item GTK model reconciliation.
- Bulk transfer checks, selection synchronization, cut styling, and context-menu summaries were changed from quadratic behavior to linear/set-based lookups.
- Large filesystem-monitor bursts now collapse into one rescan instead of starving the UI with thousands of notifications.
- Context menus remain responsive for large selections, and summaries are width-bounded with a 60-character truncation limit.
- Cancellation statistics are shown before the slower Trash reconciliation begins; refresh starts after the result dialog closes.
- Column, grid, and explorer reloads now retain a loading skeleton until `LoadFinished`, avoiding the previous several-second blank pane.
- Filtered/selection views are detached while backing models reload and reconnected once at completion, avoiding repeated filtered-view recalculation.
- Skeleton padding was aligned more closely with loaded rows to reduce the visible layout shift.

### Additional defects discovered

These are separate from cancellation accounting and were filed independently:

- #149 — expand `~/...` paths entered in the address bar.
- #150 — support moving directories across filesystems (currently fails with `Can’t recursively copy directory`).

### Remaining visual limitation

In the largest filtered Trash reconciliation case, the loading skeleton is visible but the small header spinner can still appear stationary. The results and final filesystem/model state remain correct, and the previously blank loading period is now represented by the skeleton. Given that this is an edge-case visual limitation rather than a correctness failure, we are not blocking the cancellation fix on it.

### Automated verification

Latest branch verification:

- `cargo fmt`
- `cargo test` — **282 passed**
- `git diff --check`

Branch head: `a84b006` (`perf(ui): reconnect filtered views after reload`).

PR #97, opened 2026-09-01
feat(updates): add RC release channel with stable rollback

### LINKED ISSUE #61: feat(updates): add RC and nightly release channels with stable rollback
## Problem

The updater currently follows the stable GitHub release only. We need a safe opt-in path for users who want preview builds, while ensuring stable users never see release candidates or nightly update prompts. Preview users must also have an obvious way back to the latest stable release.

### 1. Publish RC builds from GitHub Actions

- Extend the release workflow with an RC release mode.
- Produce semver prerelease tags such as `v0.5.0-rc.1`.
- Mark RC GitHub releases as **prerelease**, never as `latest`.
- Build and publish both supported architectures and both Strata executables.
- Publish checksums, provenance attestations, source commit metadata, and release notes using the same security controls as stable releases.
- Keep stable and prerelease version/tag generation deterministic and prevent accidental replacement of a stable release.

### 2. Add an opt-in preview/nightly update channel

In **Settings → Updates**, add a release-channel preference with clear language:

- **Stable** (default): final releases only.
- **Nightly / preview** (explicit opt-in): nightly builds and eligible RC prereleases.

The selected channel must persist. Opting in should immediately check the preview feed and clearly identify the offered build, channel, version/tag, source commit, and publication date before installation. Existing checksum and provenance expectations must continue to apply.

### 3. Support rollback to stable

While running a nightly or RC build, the Updates page must offer **Return to stable** even when the stable semantic version compares lower than the installed prerelease/nightly version.

Rollback must:

- fetch the latest final stable release explicitly;
- show the target version and confirmation before replacing binaries;
- install both `strata` and `strata-preview-helper`;
- switch the persisted channel back to Stable after a successful rollback;
- preserve user preferences and data;
- report download, verification, installation, and restart failures clearly.

## Channel isolation requirements

- Users on Stable must not receive, see, or be notified about RC/nightly releases.
- Stable checks must use a final-release-only source such as GitHub’s `/releases/latest` endpoint and defensively reject prerelease/draft metadata.
- Preview checks may enumerate prereleases, but must ignore drafts, unsupported artifacts, malformed tags, and releases for other channels.
- An RC/nightly must never update a stable user merely because its version sorts newer.
- Automatic checks and the sidebar update notice must respect the selected channel.
- Release notes and update dialogs must visibly label prerelease software.

## Acceptance criteria

- [ ] A maintainer can publish an RC through a protected GitHub Actions workflow without changing the stable `latest` release.
- [ ] Stable remains the default for new and existing installations.
- [ ] Stable users cannot see RC/nightly builds anywhere in the update UI.
- [ ] A user can explicitly opt into the preview/nightly channel and install an eligible build.
- [ ] A preview user can update between eligible preview builds.
- [ ] A preview/RC installation can downgrade to the latest stable build.
- [ ] Channel choice survives restart and automatic checks obey it.
- [ ] Both executables, checksums, and provenance are handled for upgrade and rollback.
- [ ] Unit tests cover channel filtering, prerelease ordering, stable isolation, persisted preferences, downgrade selection, malformed metadata, and unsupported assets.
- [ ] Workflow tests/documentation cover stable, RC, and nightly publication behavior.

## Design note

Define the relationship between “nightly” and “RC” explicitly during implementation. The proposed UI uses one opt-in preview channel that may consume both, but metadata should retain their distinct labels so a user always knows what they are installing.

spandan11106:
Closes #61.

Adds an opt-in preview release channel to the updater, an RC publication mode to the release workflow, and a "Return to stable" path back off a prerelease.

## Root fix

`parse_version` read three `u64` segments and zero-filled anything unparseable, so `0.5.0-rc.1` parsed as `(0, 5, 0)` — indistinguishable from final `0.5.0`. It is replaced by a `Version` type implementing real semver §11 precedence, where `Version::parse` returns `Option` and is the single place any tag is interpreted anywhere in the codebase.

## Architecture

A new pure module `src/services/release_channel.rs` owns every channel, version, eligibility, and ordering decision — no I/O, no serde, no GitHub types. `update_check.rs` is now GitHub transport only and delegates all judgement to it. That boundary is what makes the "a Stable user never sees a prerelease" guarantee provable in unit tests with no network access.

## Channel isolation

Stable and preview use separate fetch paths, so a Stable user's process never enumerates prereleases at all. Stable results are *additionally* filtered through `is_eligible`, which requires both that the parsed tag carries no prerelease and that GitHub's own `prerelease` flag is false — the defensive redundancy the issue asks for. Everything fails closed to Stable: unknown config values, a missing config key, unparseable build metadata.

## Prerelease identity

Travels via `STRATA_RELEASE_TAG` / `STRATA_BUILD_KIND`, injected by `build.rs` exactly as `STRATA_BUILD_COMMIT` already is. An RC run tags the source commit and never pushes a version bump to the default branch, so a prerelease cannot poison the stable version line. `--prerelease` on the GitHub release keeps `/releases/latest` pointing at the last final release, which is the endpoint Stable users poll.

## Design decisions made where the issue left them open

- **RC outranks nightly** at equal core version, matching semver's alphanumeric ordering of prerelease identifiers (`"nightly"` sorts before `"rc"`). Once an RC is cut, that line has stabilized; pulling a preview user onto a same-core nightly would be a stability regression.
- **Preview = stable ∪ prereleases**, newest by semver precedence, so a user on `0.5.0-rc.2` is offered final `0.5.0` as an ordinary update instead of being stranded on a stale RC. "Return to stable" remains the separate explicit channel switch, and is gated on the *running build* being a prerelease rather than on the selected channel — otherwise a user who already switched their preference back to Stable would be stranded on an RC with no exit.

## Scoped out

The issue's "both executables" wording came from #59, which is deferred — this repo ships one binary (`src/main.rs` dispatches `--preview-helper` as `argv[1]`). The installer takes a *list* of binaries to place, so #59 becomes a small change rather than a rewrite if it lands.

## Partially met

The **"workflow tests"** acceptance criterion. There is no workflow test harness in this repo. The version-calculation logic is extracted to `scripts/release_version.py` with 18 tests wired into CI as a `release-scripts` job, but the workflow wiring itself is only verifiable by a manual `rc` dry run. Flagging rather than papering over.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` (235 passed, up from a 176 baseline), and `python3 scripts/test_release_version.py` (18 passed) are all clean.

The GTK wiring for the channel selector and rollback row is not unit-testable and has not been visually confirmed — worth an eyeball before merge.

## Known follow-ups

Non-blocking, found in review and left deliberately:

- A hard-killed install now leaks a uniquely-named staging directory with no sweep (previously the pid-named directory was reused).
- The install-in-progress guard is per-window rather than per-process; with unique staging paths the worst case is two complete verified binaries installed in sequence, not corruption.
- A channel toggle during an install causes cosmetic status churn in the update row; the guard holds and no second install starts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


l0gicgate:
@spandan11106 can I ask you that you remove the claude attributions to the commits. I do not want Claude to become the biggest contributor on the repo at some point 😅 

Also please add some visual evidence (screenshots, videos)

Thank you 🙏🏼

spandan11106:
Yea sure

spandan11106:
Visual evidence, as requested. Both captures are from real builds on Wayland — no mockups.

## 1. Stable channel (default build)

Updates page on the stable channel

The channel selector renders with **Stable — "Final releases only."** as the default. The check ran against the live GitHub API and resolved `v0.5.0`, pulling real release notes through the existing markdown renderer. Note there is **no "Return to stable" row** here, which is correct: it is gated on the running build being a prerelease, not on the selected channel.

The sidebar notice is visible bottom-left in the full-window shot:

Full window

## 2. Preview channel on an actual RC build

Built with `STRATA_RELEASE_TAG=v0.5.1-rc.1 STRATA_BUILD_KIND=rc` — the same env-var injection `release.yml` uses when publishing an RC.

Updates page on the preview channel with rollback

Four things this confirms end to end:

- **Nightly / preview** is selected, loaded from the persisted `release_channel` key in `settings.toml` (it round-tripped across a restart).
- **"Up to date — version 0.5.1-rc.1"** — the build correctly identifies itself as an RC via `build.rs` injection, rather than reporting the bare `0.4.0` package version.
- **"Return to stable"** appears, because `build_kind() != Stable`.
- It reports up to date even though stable `v0.5.0` exists on GitHub, because `0.5.1-rc.1 > 0.5.0` by semver precedence — the preview channel correctly declines to offer a downgrade, while "Return to stable" remains available as the explicit way back.

The release-notes card degrades gracefully for a tag that does not exist upstream.

## How these were captured

The Updates pane was opened via a temporary local patch that set the settings stack's initial child, since I had no pointer automation available. The patch was reverted before pushing — the branch is clean and `cargo test` is at 235 passing. The widgets shown are the real ones built by the committed code.

Still not covered by these stills: live theme switching across the three new CSS classes, and a full download-verify-install rollback cycle against a real published RC. The latter needs an actual RC on the releases page, so it is genuinely blocked until the first `rc` dispatch.

spandan11106:
Fixed a UI defect spotted while running the branch: **the channel row rewrote its own label as the switch moved.**

With the switch off it read "Stable — Final releases only"; with it on, "Nightly / preview". The behaviour underneath was right, but it inverts how every other row on the page works. Elsewhere the label names a fixed thing and the switch enables it, so a row reading **"Stable"** beside an **off** switch looks like stable releases are being disabled — the opposite of the truth.

Now the label names what the switch actually enables and stays put:

**Off — stable only:**

Channel row, switch off

**On — opted into preview builds:**

Channel row, switch on

The row is now built from the existing `settings_option` helper rather than hand-rolling a duplicate of it, so it is structurally identical to its neighbours — net −27 lines. The test that pinned the old mutating copy is replaced by one asserting the label names the thing the switch turns on.

These captures are also from a different theme than the earlier ones, which incidentally confirms the new row's semantic `@theme_*` colours follow live theme changes.

`cargo fmt --check`, `cargo clippy --all-targets`, and `cargo test` (235 passing) all clean.

spandan11106:
## Live end-to-end test of channel isolation

Tested against real GitHub releases rather than fixtures. Two releases were published to a fork — a final `v0.5.0` and a prerelease `v0.5.1-nightly.20260901` — each with a properly named archive and `.sha256`, built through the same `STRATA_RELEASE_TAG` / `STRATA_BUILD_KIND` injection `release.yml` uses. The client was pointed at that fork with a temporary local edit (reverted; not in this branch). The running build reports `v0.4.0`, so both releases sort newer than it.

**Server side, before the client is involved:** `/releases/latest` on that fork returns `v0.5.0`, not the nightly — confirming `--prerelease` keeps prereleases out of the endpoint stable users poll.

### Switch off — stable only

Stable channel

Offers `v0.5.0` with its real release notes. The nightly appears nowhere — not in the check, the notes card, or the sidebar — **even though `0.5.1-nightly.20260901` sorts strictly newer than both the installed `0.4.0` and the offered `0.5.0`.** That is the guarantee under its actual adversarial condition: a prerelease that would win on version precedence alone is still not offered, because it never enters the stable code path.

### Switch on — preview

Preview channel

Flipping the switch re-checks immediately rather than waiting for a restart, and the offer changes to `v0.5.1-nightly.20260901`, carrying a **`Nightly` badge** so the build is labelled as prerelease software before anything is installed.

Toggling back returns to offering `v0.5.0`, and the persisted `release_channel` key survives a restart in both states.

### What this covers

- Stable isolation against a live feed, in the case where version ordering alone would leak the prerelease.
- Preview enumeration, prerelease labelling, and re-check on opt-in.
- Preference persistence across restarts.

Still uncovered: the download → checksum → install → restart cycle itself, and rollback from an installed prerelease. Those need a real published RC to exercise honestly.

l0gicgate:

spandan11106:
### Rollback tested; one follow-up noted

Installing the nightly worked end to end — download, checksum verification against the published `.sha256`, binary replacement, and on restart the build correctly identified itself as `0.5.1-nightly.20260901` and grew the **Return to stable** row.

Clicking that row returned *"GitHub API rate limit reached"*. That turned out to be accurate rather than a defect — unauthenticated GitHub API access is capped at 60 requests/hour per IP, and the testing session had exhausted it (`/rate_limit` reported `remaining: 0`, and `/releases/latest` returned a genuine 403). It confirms two things working as intended: rollback queries `/releases/latest` rather than reusing the preview feed, and a failed rollback surfaces a readable message while leaving the user on the preview channel and able to retry.

Worth recording as a follow-up though, since the message says what happened but not what to do about it: GitHub returns an `x-ratelimit-reset` header, so the error could name the time the quota refills instead of leaving the user to guess whether the updater is broken. Preview users are marginally more exposed, since enumerating the preview feed costs an API call that the single `/releases/latest` stable check does not.

Deliberately left out of this branch to keep it scoped to what #61 asked for.

kallupx:
Hey, I’d suggest making release-channel changes process-wide. Switching one window back to Stable only refreshes that window, so another open window can retain and install a cached RC offer ([channel toggle](https://github.com/lgse/strata/blob/9e1d2b25de3830711e9a3ff64ad4c58d8044ff11/src/ui/settings.rs#L456-L464)) Or at minimum notify every update view or re-check the current channel before installation.

Also, installation serialization should happen per process. Each window currently creates its own guard ([guard creation](https://github.com/lgse/strata/blob/9e1d2b25de3830711e9a3ff64ad4c58d8044ff11/src/ui/window.rs#L264-L270)), so updates and rollbacks from different windows can race to replace the same executable. The last replacement wins, potentially leaving an RC binary installed while the rollback reports success and saves the Stable preference.

spandan11106:
@kallupx  you are right that the guard was per-window, present() built its own Rc<Cell<bool>>, so it only serialized the update row, rollback row, and dialog within one window. There's now a single thread_local guard (settings::install_guard()) that every window shares; all windows are built on the one GTK main thread from connect_activate, so no locking is needed, and it mirrors the existing ThemeManager::shared() pattern. That also removes the specific rollback failure you described: the rollback only persists Channel::Stable on install success, and a concurrent update in another window can no longer start.

Worth noting the ceiling: this bounds races to one process. Two separate strata processes can still install over the same executable, but the installer stages into a unique path and lands it with one atomic rename, so that case is last-writer-wins rather than a corrupted binary.

spandan11106:
Also some more stuff.

I took your "re-check the current channel before installation" option, since it holds regardless of how many views cached an offer. The offered BuildKind is now kept beside the pending request, and the channel is re-tested at the click: the update row discards the offer and re-checks instead of installing, and the dialog withdraws the offer and turns its action button into Close. InstallRequest stays URL-only; the UI layer already has the ReleaseMetadata. 

I didn't do the broadcast half (notifying every open view on a toggle so a stale window stops advertising an RC). That needs an observer mechanism on ThemeManager, which doesn't exist yet, and it's a UX improvement on top of the install-time gate rather than a substitute for it. 

I think we could do this in another issue rather then this PR since it is not in its scope.

spandan11106:
@l0gicgate I think this is merge ready now.

kallupx:
Yup the shared install guard and install-time channel check fix the multi-window issues, thanks!

Some notes still:
Rollback uses /releases?per_page=30 instead of /releases/latest, so it can miss the stable release after enough (30) prereleases.
The displayed commit comes from target_commitish, which GitHub can return as main rather than the exact SHA.

Also, remove the committed scripts/__pycache__/*.pyc files.

l0gicgate:
@spandan11106 incredible stuff. I will spend some time reviewing this one after work tonight. It's a big one. Thank you for this contribution 🙏🏼

l0gicgate:
Okay, I'm pulling this down now. Apologies, I've been a little busy 😅

l0gicgate:
Pushed a maintainer review/update pass to this branch.

### What changed

- Merged the current `main` branch and resolved the Settings/window test conflicts.
- Split update selection into three persisted channels:
  - **Stable** — final releases only.
  - **Preview** — alpha, beta, RC, and stable releases; excludes nightlies.
  - **Nightly** — receives every recognized release, including nightly development builds.
- Added alpha and beta tag parsing, labels, ordering, and manual publication modes.
- Defined equal-core precedence as `nightly < alpha < beta < rc < stable`, so a nightly user is offered an RC when one becomes available and a same-core nightly cannot pull them back from an RC.
- Kept RC publication manual through the protected **Release** workflow. Alpha and beta publication use the same manual path and numeric ordinals.
- Replaced the old combined preview/nightly switch with a Stable / Preview / Nightly selector.
- Moved the release-channel selector below **Automatically check for updates** and disabled it whenever automatic checks are off.
- Tightened staged prerelease parsing so alpha, beta, and RC ordinals must be positive.
- Updated the release runbook and README for the new channel behavior.

Automatic nightly **publication** is intentionally not wired yet. Consumption is supported, but scheduling nightlies requires choosing the next target core version (for example `0.8.0`) and a retention policy; it should not guess patch versus minor from `Cargo.toml`.

Validation is clean: Rust formatting, Clippy with warnings denied, all 355 Rust tests, and all 20 release-script tests.

### Updated Settings UI

Stable, Preview, and Nightly release-channel selector

spandan11106:
@l0gicgate thanks for the pass — the three-way split reads well and the alpha/beta ordinals are clean. I went through the four commits and found a few things worth resolving before merge. Line links are against `6876b5c`.

### 1. Preview can silently report "up to date" once nightlies start publishing

[`fetch_update`](https://github.com/lgse/strata/blob/6876b5c6e5412e2da2f9eb8b111c3b6fabac6b22/src/services/update_check.rs#L293-L297) routes Preview and Nightly through the same `fetch_preview()`, which is one bounded page ([`PREVIEW_PAGE_SIZE = 30`](https://github.com/lgse/strata/blob/6876b5c6e5412e2da2f9eb8b111c3b6fabac6b22/src/services/update_check.rs#L23)). That bound was harmless while Preview accepted every entry on the page, but [`is_eligible`](https://github.com/lgse/strata/blob/6876b5c6e5412e2da2f9eb8b111c3b6fabac6b22/src/services/release_channel.rs#L334) now filters nightlies out of it. Once nightly publication is wired, ~30 consecutive nightlies push every alpha/beta/rc/stable release off the page, nothing eligible remains, and a Preview user is told `UpToDate` while a newer RC exists.

Stable avoids this by construction — it polls `/releases/latest`. Preview has no equivalent. Fix is either to page until a non-nightly is found, or to raise the page size enough that the nightly cadence can't bury a release stage.

### 2. A Nightly-channel user who lands on an RC stops receiving nightlies

[`BuildKind`](https://github.com/lgse/strata/blob/6876b5c6e5412e2da2f9eb8b111c3b6fabac6b22/src/services/release_channel.rs#L57-L63) orders `Nightly < Alpha < Beta < Rc`, and `Version::cmp` compares kind before ordinal. So with `0.5.0-rc.1` installed, a later `0.5.0-nightly.20260910` is never offered — the user is pinned until the core version bumps.

The original justification for that ordering was that pulling a *preview* user onto a same-core nightly is a stability regression. That still holds for Preview, but Nightly is now an explicit separate opt-in, and someone who selects it is asking for dailies. Worth deciding deliberately rather than inheriting it from the old two-channel model.

### 3. The channel selector is disabled while manual checks still honour it

[`channel_row.set_sensitive(auto_check_enabled)`](https://github.com/lgse/strata/blob/6876b5c6e5412e2da2f9eb8b111c3b6fabac6b22/src/ui/settings.rs#L472) greys out the whole selector when automatic checks are off, but "Check now" stays sensitive and reads `manager.release_channel()` at click time. A user who turns off automatic checks while on Nightly is locked onto Nightly with no way out, and manual checks keep serving them nightly builds. Either disable "Check now" alongside it, or leave the selector sensitive.

### 4. The segmented control desyncs from the persisted channel after "Return to stable"

[`selected`](https://github.com/lgse/strata/blob/6876b5c6e5412e2da2f9eb8b111c3b6fabac6b22/src/ui/settings.rs#L532-L536) is computed once at construction, and the rollback success path calls [`set_release_channel(Channel::Stable)`](https://github.com/lgse/strata/blob/6876b5c6e5412e2da2f9eb8b111c3b6fabac6b22/src/ui/settings.rs#L1326) without touching the toggle buttons. Within the same settings session the control still shows Preview/Nightly active while the stored preference is `stable`. This predates your change with the old switch, but since the widget was rewritten here it's the natural place to wire it up.

### 5. The release workflow's supersede guard only matches RC tags

[`git tag -l "v${VERSION}-rc.*"`](https://github.com/lgse/strata/blob/6876b5c6e5412e2da2f9eb8b111c3b6fabac6b22/.github/workflows/release.yml#L94) was written when RC was the only prerelease stage. Now that alpha and beta are publishable, a stable release can supersede a `v0.5.1-beta.3` whose commit isn't reachable from the release source, with no error — precisely the case the guard exists to block. Widening the glob to `v${VERSION}-{alpha,beta,rc}.*` covers it.

### Minor

- [`is_eligible`](https://github.com/lgse/strata/blob/6876b5c6e5412e2da2f9eb8b111c3b6fabac6b22/src/services/release_channel.rs#L334)'s Preview arm consults only the parsed tag, so the two-signal redundancy its own doc comment asserts ("a release mislabelled on either signal alone must still be caught by the other") now holds for Stable only. A nightly published under a mislabelled tag but correctly flagged `prerelease: true` passes for Preview.
- [`build_info.rs`](https://github.com/lgse/strata/blob/6876b5c6e5412e2da2f9eb8b111c3b6fabac6b22/src/build_info.rs#L26) still says "not one of those **three** values" — it's five now.

### Checked and clean

The `n == 0` rejection for staged ordinals, `parse_prerelease`'s kind mapping and the `Display` round-trip, `next_prerelease_ordinal`'s per-stage independence and numeric comparison, the `Channel::parse`/`as_str` persistence round-trip, `build.rs`'s widened allowlist, and every `BuildKind` consumer — all of them test `== Stable` / `!= Stable`, so alpha and beta are treated as prereleases everywhere they need to be.

Happy to take 1, 3, 4, and 5 on this branch; 2 is your call on intent.

l0gicgate:
### Channel-transition UX update

The RC test exposed a confusing state: after selecting **Stable** while running an RC, the ordinary update card said the RC was “up to date” while a separate card offered **Return to stable**. Both statements followed the old implementation, but together they made the selected channel look ineffective.

The UX now treats the selected channel as the user’s desired destination:

- There is one **Updates** status/action card instead of separate update and rollback cards.
- Changing Stable / Preview / Nightly immediately refreshes that one card.
- When a prerelease installation selects **Stable**, the stable feed is allowed to return the newest final release even when its semantic version is lower than the installed RC/nightly.
- The card names the exact destination, for example **Stable channel target: v0.7.0**, and changes its action to **Return to stable**.
- Preview and Nightly use the same card for ordinary forward updates.
- Release notes for the offered destination remain inline below the action, so the version and changes can be reviewed before installation.
- The channel selector remains below **Automatically check for updates** and is disabled when automatic checks are off.

This removes the contradictory “up to date” plus “return to stable” presentation and makes switching channels a single, explicit flow.

Stable channel selected on an RC, with the unified Updates card offering v0.7.0

PR #107, opened 2026-09-01
fix(ui): dismiss modal dialogs on backdrop click

### LINKED ISSUE #105: Modal dialogs do not dismiss when clicking the backdrop
### Strata version

0.6.0 (source build at commit `a6c872105a33a8857776e9fe44ca3cf55636f255`).

### Installation method

Built from source.

Arch Linux, Hyprland (Omarchy), Wayland.

### What happened?

Confirmation and prompt dialogs (delete, replace, empty trash, rename, connect-to-server, etc.) are rendered as custom modal layers with a blurred backdrop. Clicking on the backdrop outside the dialog does nothing — the dialog stays open. The only ways to dismiss are the Cancel/Close buttons or pressing Escape. This is inconsistent with the popover/menu behavior users expect, where an outside click dismisses the surface.

### Steps to reproduce

1. Open Strata and select a file.
2. Press Delete to open the delete-confirmation dialog.
3. Click anywhere on the blurred backdrop outside the dialog.
4. Observe that the dialog remains open.

### Expected behavior

Clicking the backdrop (the area outside the dialog content) should dismiss the modal the same way the Cancel button does, without triggering the confirm action.

### Screenshots or video

N/A — interaction issue, not visible in a static screenshot.

### Relevant logs

N/A — no errors observed.

### Additional context

The modals are built by `modal_layer` in `src/ui/browser.rs`, which composes a fullscreen `gtk::Box` with `top`/`bottom` spacer boxes around the content. The spacers act as the backdrop but have no click gesture attached. There are ~20 call sites that wire Cancel/Close/Escape to `dismiss_modal_layer`; the backdrop click should reuse the same dismiss path.

JoeJoeflyn:
## Summary
- Wire a `GestureClick` on the modal layer that uses `gtk::Widget::pick` to distinguish backdrop clicks from dialog-content clicks, dismissing the modal only when the click lands on the backdrop
- The "creating destination" in-progress state passes a `block_dismiss` guard (`Rc<Cell<bool>>`) so the backdrop cannot dismiss the dialog mid-operation, matching the existing Escape guard
- Updated all 10 `modal_layer` call sites (9 in `browser.rs`, 1 in `settings.rs`) to pass the overlay, root, and optional guard

## Manual steps
1. Open Strata, select a file, press Delete
2. Click on the blurred backdrop outside the delete-confirmation dialog
3. The dialog dismisses without deleting
4. Repeat with the "Move/Copy to" dialog: click backdrop before confirming — dismisses; click backdrop while "Creating folder…" is shown — does not dismiss

## Expected result
Clicking the backdrop dismisses any modal dialog the same way Cancel/Escape does, except during non-cancellable in-progress states.

Closes #105

#### Test plan
- [x] `cargo build --release` passes
- [x] `cargo test` — 234 passed, 0 failed
- [x] Manual: backdrop click dismisses delete confirmation
- [x] Manual: backdrop click does not dismiss during "Creating folder…" state
screenrecording-2026-09-01_15-34-09

### DIFF
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -1379,11 +1379,16 @@ impl ViewState {
             );
         });
 
+        let initial_text = folder_input_path(&base);
+        let dirty_field = field.clone();
+        let dirty_creating = creating_destination.clone();
         let layer = modal_layer(
             &content,
             &window_overlay,
             blurred_root.clone(),
-            Some(creating_destination.clone()),
+            Some(Rc::new(move || {
+                dirty_creating.get() || dirty_field.text().to_string() != initial_text
+            })),
         );
         window_overlay.add_overlay(&layer);
         let cancel_layer = layer.clone();
@@ -1964,6 +1969,7 @@ impl ViewState {
         title: &str,
         subtitle: &str,
         confirm_label: &str,
+        block_dismiss: Option<Rc<dyn Fn() -> bool>>,
     ) -> (gtk::Box, gtk::Button, Rc<dyn Fn()>) {
         let Some(window_overlay) = self
             .overlay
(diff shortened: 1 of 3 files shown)


l0gicgate:
We can make them dismissable if it’s a confirmation dialog. But if it’s a form, and it gets dirty we shouldn’t. 

Can you adjust this PR accordingly.  

Thank you btw!

JoeJoeflyn:
sure

JoeJoeflyn:
updated, thanks for the feedback

PR #115, opened 2026-09-01
fix(ui): cancel rename on click outside

JoeJoeflyn:
### What
- Clicking outside the rename entry cancels the rename and restores the original name
- Works in all three modes: columns, grid, explorer

### Why
- Previously clicking outside during rename left the field open indefinitely — only Escape or Enter could end it
- The fix reuses the existing hidden inline-rename field in grid/explorer mode instead of creating a new Entry each time, which also fixes a grid card resize/reflow bug (7 items jumping to 6)

#### Manual verification
- [x] In columns mode, start rename (F2), click another row — rename cancels
- [x] In columns mode, start rename, click the sidebar — rename cancels
- [x] In grid mode, start rename, click another card — rename cancels, no reflow
- [x] In explorer mode, start rename, click another row — rename cancels
- [x] Start rename, press Enter — rename submits
- [x] Start rename, press Escape — rename cancels
- [x] Start rename, click inside the entry — rename continues
screenrecording-2026-09-01_20-26-26

### DIFF
--- a/src/ui/browser_modes.rs
+++ b/src/ui/browser_modes.rs
@@ -283,7 +283,8 @@ impl ModeViews {
             return false;
         };
         rename.label.set_visible(true);
-        rename.field.unparent();
+        rename.field.set_visible(false);
+        rename.field.set_sensitive(true);
         true
     }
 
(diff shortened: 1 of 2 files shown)


l0gicgate:
Annnd another one let's goooo 🚀

JoeJoeflyn:
thank youuu <3

PR #116, opened 2026-09-01
fix(archive): confine extraction to destination

### LINKED ISSUE #109: Archive extraction can write outside the selected destination
### Strata version

0.6.1 (source build at commit `e00117a79f1ab3d9ee2150a53b87a94300397d8c`).

### Installation method

Built from source.

Omarchy 4.0.2 (Arch Linux), Hyprland 0.56.2, Wayland.

### What happened?

Archive extraction can create files outside the destination selected by the user. The shared path check joins an archive-controlled name to the destination and performs a lexical `starts_with` comparison. A name such as `folder/../../escaped.txt` still begins with the destination lexically, but filesystem resolution writes it beside that destination.

Isolated tests using Strata's production extraction functions confirmed the escape for ZIP, TAR, and 7z. TAR.GZ uses the same extraction function as TAR after wrapping the input in a gzip decoder. A separate test confirmed that an existing dangling symlink at the destination is followed by `File::create`, allowing a nominally in-destination filename to redirect the write outside it.

Opening an untrusted archive can therefore create attacker-chosen files outside the selected destination at locations writable by the user.

### Steps to reproduce

1. Create an isolated root with an empty `destination/` directory.
2. Create a ZIP, TAR, or 7z archive containing a regular-file entry named `folder/../../escaped.txt`.
3. Extract the archive into `destination/` using Strata.
4. Observe that `escaped.txt` is created in the isolated root, outside `destination/`.
5. Separately, create `destination/victim.txt` as a dangling symlink to `../escaped-via-link.txt`, then extract an archive entry named `victim.txt`.
6. Observe that the symlink target outside the destination is created.

### Expected behavior

Every extracted object must remain beneath the selected destination even when entry names contain traversal components or destination path components are symlinks. Unsafe entries should fail the extraction without creating out-of-tree files.

### Screenshots or video

N/A — no screenshot was captured; filesystem assertions provide deterministic evidence.

### Relevant logs

N/A — extraction completes without reporting an error.

### Additional context

- `safe_extract_path` performs the insufficient lexical check at `src/adapters/local_operations.rs:831-836`.
- `ExtractNameResolver::resolve` preserves inner `..` components at `src/adapters/local_operations.rs:948-966`.
- Raw entry names reach the shared resolver and filesystem writers for ZIP (`984-1001`), TAR (`1023-1043`), and 7z (`1115-1141`).
- Temporary adjacent unit tests called the production ZIP, TAR, and 7z extraction paths and asserted that each created its marker outside the destination; all three passed. A fourth test confirmed the dangling-symlink redirect. The diagnostic tests were removed after the audit.
- Archive support was introduced by `d3e5847`; PR #81 stated that extraction included path-traversal protection.
- Open issue #8 tracks descriptor-relative hardening for pre-existing recursive operations. This archive-specific deterministic traversal was introduced later and needs explicit format coverage.

#### Acceptance criteria

- [ ] Reject absolute paths, prefixes, root components, and every parent-directory component before conflict-name rewriting.
- [ ] Use archive-native enclosed-path helpers where available, but keep one shared invariant for all formats.
- [ ] Create directories/files descriptor-relatively with no-follow/beneath guarantees so pre-existing or swapped symlinks cannot redirect output.
- [ ] Tests cover ZIP, TAR, TAR.GZ, and 7z traversal names plus dangling and intermediate destination symlinks.
- [ ] A failure leaves no files outside the destination.

kallupx:
## Description

Validate every archive entry as a non-empty relative path before conflict resolution, with additional format-specific checks for ZIP and 7z archives. Extraction now creates directories and files relative to a single open destination descriptor, preventing path traversal and symlink redirection outside the selected folder.

Preserves nested extraction, password handling, progress reporting, and conflict suffixing for ZIP, TAR, TAR.GZ, and 7z archives. Adds regression coverage for traversal attempts, dangling and intermediate symlinks, normal nested extraction, and existing-name conflicts.

## Visual evidence

N/A - this is an archive extraction security fix without interface changes.

## How to test

1. Extract ZIP, TAR, TAR.GZ, and 7z archives containing nested files.
2. Extract an archive when a regular file or directory with the same name already exists.
3. Attempt to extract archives containing `../` or absolute entry paths.
4. Attempt extraction over a dangling symlink or through a symlinked directory.
5. Run `cargo test adapters::local_operations::tests` and `./scripts/check.sh`.

**Expected result:** Normal archives extract successfully, existing regular names receive numbered suffixes, and traversal or symlink-based extraction fails without writing outside the selected destination.

## Related issue

Closes #109


l0gicgate:
Solid stuff @kallupx, tested, everything works as expected

PR #117, opened 2026-09-01
fix(ui): make Enter activate confirm in modal dialogs

JoeJoeflyn:
### What
- Pressing Enter in the paste conflict and delete confirmation dialogs now activates the confirm button instead of cancel
- Confirm button receives initial keyboard focus

### Why
- Cancel had keyboard focus by default, so Enter activated cancel — confirming a destructive or intentional action via keyboard was impossible without tabbing first

#### Manual verification
- [x] Trigger a paste conflict, press Enter — Replace activates
- [x] Trigger a delete (trash), press Enter — Delete activates
- [x] Trigger a permanent delete, press Enter — Delete activates
- [x] Press Escape in any dialog — dialog dismisses
- [x] Tab between buttons — focus moves normally
- [x] Arrow keys still move focus in delete dialog

### DIFF
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -1174,19 +1174,24 @@ impl ViewState {
         });
 
         let escape = gtk::EventControllerKey::new();
+        escape.set_propagation_phase(gtk::PropagationPhase::Capture);
         let escaped_layer = layer.clone();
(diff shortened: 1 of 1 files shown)


l0gicgate:
Just tested, flawless!

PR #119, opened 2026-09-01
fix(preview): play GIFs with bounded media streaming

### LINKED ISSUE #128: fix(preview): avoid hardware media crashes from sandbox file limits
### Strata version

Development build at `c209c63` (package version 0.6.1).

### Installation method

Built from source.

Omarchy 4.0.0 (Arch-based), Hyprland 0.56.2 on Wayland; Intel UHD Graphics 620 using `i915`; Mesa/Vulkan Intel 26.1.7; FFmpeg 9.0.1.

### What happened?

Opening an animated GIF in quick preview produces no playable preview and causes one or more “Process crashed: ffmpeg” notifications. Strata itself remains running; the crashing process is its sandboxed FFmpeg hardware-preview subprocess.

This is not specific to one malformed GIF. It reproduces with independently generated and downloaded animated GIFs.

The immediate failure is caused by applying a process-wide 512 MiB `RLIMIT_FSIZE` to hardware-accelerated FFmpeg. During Vulkan initialization, Intel's Vulkan driver calls `ftruncate` for a 1 GiB driver-managed file or memory object. The kernel sends FFmpeg `SIGXFSZ` before it decodes the input. Other drivers may use different file sizes or reserve large virtual-address ranges, so increasing a fixed limit is not a general solution.

Software conversion succeeds for both reproducing GIFs. Hardware initialization failure should be an ordinary backend failure followed by software fallback, without generating a core dump or preventing the preview.

### Steps to reproduce

1. Run Strata from the current development branch on the environment above.
2. Select an animated GIF to open quick preview.
3. Observe that the preview does not play.
4. Observe a desktop notification that FFmpeg crashed.
5. Run `coredumpctl info ffmpeg` and observe `SIGXFSZ` during `vkCreateDevice`.

### Expected behavior

The animated GIF should preview. Unsupported or incompatible hardware acceleration should fail quietly and fall back to software decoding. Preview resource controls should work across drivers without relying on guessed limits for opaque driver internals.

### Screenshots or video

N/A — the visible result is an empty media preview plus the FFmpeg crash notification.

### Relevant logs

```shell
Signal: 25 (XFSZ)
Command Line: ffmpeg ... -init_hw_device vulkan=vk:0 ... -i /input ...

#0  ftruncate64 () from /usr/lib/libc.so.6
#1  ... from /usr/lib/libvulkan_intel.so
#8  vkCreateDevice () from /usr/lib/libvulkan.so.1
#11 av_hwdevice_ctx_create () from /usr/lib/libavutil.so.61

# Register values at ftruncate:
rsi = 0x40000000 (1073741824 bytes / 1 GiB)

# Strata sandbox limit:
prlimit --fsize=536870912 (512 MiB)
```

The machine had approximately 11 GiB available RAM and 30 GiB unused swap. There was no OOM kill. No core dump is attached because it may contain private data.

### Additional context and proposed long-term solution

Do not try to enumerate every GPU/driver/media combination or maintain per-driver limits. Instead:

1. Do not apply `RLIMIT_FSIZE`, and likely not `RLIMIT_AS`, to opaque hardware-driver internals. Drivers can legitimately create large sparse objects or reserve large virtual-address ranges without consuming equivalent resources.
2. Stream FFmpeg's fragmented MP4/WebM output through stdout and enforce Strata's existing `MAX_OUTPUT_BYTES` while reading. Terminate FFmpeg if the retained artifact exceeds that bound.
3. Preserve the existing containment: read-only input, private mount namespace, no network, minimal device exposure, wall timeout, duration/dimension/frame-rate/bitrate limits, and output validation.
4. Treat hardware acceleration as best-effort. Try discovered backends with short timeouts and always retain a software fallback.
5. As a follow-up defense in depth, use cgroup v2 accounting for actual resident memory, CPU, process count, and tmpfs usage rather than limits on virtual address space.

The immediate implementation should be moderate in scope: adjust media limits in `src/sandbox.rs`, adapt `src/sandbox_helper.rs` to bounded streaming while preserving timeout/fallback behavior, and add adjacent tests for oversized output and backend fallback. Cgroup integration can remain a separate follow-up and should not block the preview fix.

JoeJoeflyn:
## Description

Play animated GIFs through Strata's sandboxed media-preview pipeline instead of rasterizing only their first frame. GIF previews autoplay and loop, while regular video behavior remains unchanged.

Testing exposed a process-wide sandbox limit that killed FFmpeg during Intel Vulkan initialization: the driver creates a 1 GiB object, exceeding the inherited 512 MiB `RLIMIT_FSIZE`. Media previews now stream fragmented MP4 or WebM through bounded pipes instead of writing directly into a host-backed output directory. Both the helper and parent enforce `MAX_OUTPUT_BYTES`; hardware attempts retain short timeouts and software fallback. Driver-wide file-size and virtual-address limits are no longer applied to media processes, while non-media renderer limits remain unchanged.

This keeps untrusted GIF decoding outside Strata's main process and directly bounds the only media artifact returned to the application. The sandbox also caps temporary storage as a whole, while FFmpeg rejects excessive individual allocations and decoded frame dimensions without constraining sparse driver-managed files. GTK playback errors are now shown in the drawer; missing GStreamer plugins include actionable Arch/Omarchy package guidance, and the source-development instructions list those runtime dependencies.

## Visual evidence

Animated GIF preview after the change:

https://github.com/user-attachments/assets/613738ea-74b2-43d0-aa94-28168cac7118

## How to test

1. Preview an animated GIF, including one with transparency.
2. Confirm it animates, autoplays, and loops without a 30-second video notice.
3. Preview a regular video and confirm playback still works with the existing notice.
4. On a system with hardware acceleration, confirm an unsupported or failed backend falls back to software without preventing preview.
5. On Intel Vulkan, confirm previewing media does not create an FFmpeg `SIGXFSZ` core dump.
6. With GStreamer playback plugins unavailable, confirm the drawer recommends `gst-plugins-good` and `gst-libav` instead of showing only a broken player.

**Expected result:** GIF and video previews remain sandboxed and playable; retained output is bounded, hardware acceleration is best-effort, and driver initialization is not killed by guessed file or address-space limits.

## Related issue

Closes #128


kallupx:
We could probably classify GIFs as Media and reuse the sandboxed PreviewMedia transcoding here https://github.com/lgse/strata/blob/bd0e98493d431bfb741ab89be31bd7914ed3a43e/src/adapters/local_preview.rs#L69-L97 Otherwise GIF previews bypass the sandbox: fs::read loads the entire untrusted file without a size limit, and GTK/GStreamer decodes it inside Stratas process https://github.com/lgse/strata/blob/bd0e98493d431bfb741ab89be31bd7914ed3a43e/src/ui/preview.rs#L436-L444 Malicious file could consume excessive memory or CPU and hang or terminate the application, or any possible decoder vulnerability would also be exposed directly to the main process instead of contained in the preview helper.

JoeJoeflyn:
thanks for the feedback

l0gicgate:
While testing this PR on an Intel UHD Graphics 620 machine, animated GIF previews produced no playable preview and triggered `Process crashed: ffmpeg` notifications. Investigation showed that the GIF itself was not the cause: Strata applies a 512 MiB `RLIMIT_FSIZE` to the sandboxed media process, while Intel's Vulkan driver calls `ftruncate` for a 1 GiB object during `vkCreateDevice`. The kernel therefore terminates the hardware FFmpeg attempt with `SIGXFSZ` before decoding begins. There was no OOM event, and both test GIFs converted successfully through the software FFmpeg command.

This is a pre-existing hardware-preview sandbox issue exposed while testing this PR, rather than evidence of a malformed GIF. I filed #128 with the sanitized crash evidence and proposed long-term approach: bound FFmpeg's retained output stream directly, treat hardware acceleration as best-effort with reliable software fallback, and avoid applying guessed file/address-space limits to opaque driver internals.

Issue: https://github.com/lgse/strata/issues/128

l0gicgate:
@JoeJoeflyn I will take this over and apply part of #128

JoeJoeflyn:
yep, feel free to do it

l0gicgate:
Added actionable missing-media-plugin feedback to the preview drawer. Instead of leaving users with a broken player, Strata now shows a centered, theme-aware Lucide warning icon, explains that GStreamer playback support is missing, and presents the exact Arch/Omarchy install command in a read-only field. The embedded copy action copies the command and temporarily changes to a primary-colored checkmark for confirmation.

Missing media plugin guidance in the preview drawer

PR #124, opened 2026-09-01
build: publish split debug symbols for release binaries

### LINKED ISSUE #100: build: strip release binaries before packaging
### Problem or use case

Release archives currently contain unstripped ELF binaries. The x86_64 release binary is 12,455,400 bytes locally, while stripping unnecessary symbols reduces it to 9,621,720 bytes (about 23% smaller) without changing runtime behavior.

### Proposed solution

Strip the copied release binary in the packaging step before creating each release archive. Apply this consistently to both x86_64 and aarch64 artifacts.

### Alternatives considered

Configure Cargo release profile stripping in `Cargo.toml`. Keeping the optimization in the release packaging workflow avoids changing locally built release binaries and makes the artifact-specific intent explicit.

### Visual examples

N/A — this is a non-visual release packaging change.

### Strata version

0.5.0

### Additional context

GNU `strip --strip-unneeded` reduced the current local x86_64 binary from 11.88 MiB to 9.18 MiB.

kallupx:
## Description

Build each Linux release with line-table debug information, extract it into a zlib-compressed `.debug` release asset, strip the binary included in the release archive, and embed a GNU debug link.

Each architecture publishes a matching debug file and checksum. Placing that file beside the installed binary lets GDB resolve Rust function names and source lines without increasing the shipped binary size.

The README and release runbook document how to use the debug artifact with `coredumpctl debug strata`.

## Visual evidence

N/A. Release packaging only.

## How to test

1. Build with `CARGO_PROFILE_RELEASE_DEBUG=line-tables-only`.
2. Extract symbols using `objcopy --only-keep-debug`.
3. Strip the packaged binary and add its GNU debug link.
4. Confirm the matching `.debug` file lets GDB resolve `strata::main` to `src/main.rs`.

Local x86_64 results:

- Current v0.9.0 release binary: 13,601,952 bytes
- Proposed shipped binary: 10,540,944 bytes
- Compressed debug artifact: 17,607,936 bytes
- GDB resolved `strata::main` to `src/main.rs:26`
- Formatting, compilation, and Clippy passed
- Full tests: 441 passed; one unrelated logging test flaked under the full suite and passed in isolation

## Related issue

Closes #100

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -225,6 +225,8 @@ jobs:
           key: release-${{ matrix.target }}
 
       - name: Build release binary
+        env:
+          CARGO_PROFILE_RELEASE_DEBUG: line-tables-only
         run: cargo build --release --locked --target "${{ matrix.target }}"
 
       - name: Package binary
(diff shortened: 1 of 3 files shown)


l0gicgate:
So I want to hold off on merging this because once we do, this will strip debugging symbols from the binaries and people won't be able to report stack traces that are as relevant to us fixing issues. Thanks for raising the PR, let's hold for now.

PR #126, opened 2026-09-01
feat(updates): refresh every update view when the release channel changes

### LINKED ISSUE #123: feat(updates): notify open update views when the release channel changes
### Problem or use case

The release-channel preference is process-wide (it lives in the `ThemeManager` singleton), and since #97 an install re-tests the offered build kind against that preference at the moment of the click, so a preview build can no longer be installed after the user has switched back to Stable.

What is still missing is telling the open views. An offer produced by a check that already completed stays cached in whichever view produced it — the Settings update row's pending request, an open update dialog, and the sidebar update notice. Switching the channel in one window leaves every other window still *advertising* that build: the sidebar keeps its "Install Strata vX-rc.N" notice, and the update row keeps its "Install update" button. The user only discovers the offer is gone by clicking it, at which point the row silently re-checks and the dialog withdraws the offer.

The per-row generation counter (`is_stale_check`) does not help here: it only discards the result of a check still in flight, and only within the row that started it.

### Proposed solution

Give `ThemeManager` a way to notify interested views when the release channel changes, and have `set_release_channel` fire it. Each update view registers a listener that clears its cached offer and re-runs its check:

- the Settings update row: drop `pending_download` and call `run_check`
- the sidebar notice: clear `available_update` and hide the update area
- an open update dialog: withdraw the offer immediately rather than waiting for the click

Listener storage needs `Weak` handles (or GObject signals) so closures capturing widgets do not keep a closed window's UI alive.

### Alternatives considered

Re-checking the channel at install time — shipped in #97. It is the correctness gate and holds no matter how many views cached an offer, but it is a refusal at the click rather than the views staying current, so it does not cover this.

Having each view poll the preference on a timer: cheaper to write, but it burns work on every window forever to fix a rare event.

### Visual examples

N/A — this is about when existing UI updates, not new UI. The affected surfaces are the sidebar update notice and the Settings → Updates row.

### Strata version

0.6.1 (`main`). Reported against `feat/61-release-channels` at commit 7a0aefe, which is 3 commits behind `main` and still carries `0.5.0` in `Cargo.toml`.

### Additional context

Follow-up to review feedback on #97, which asked for release-channel changes to be made process-wide "or at minimum notify every update view or re-check the current channel before installation". #97 implemented the re-check and the process-wide install guard; this issue tracks the notify half.

spandan11106:
## Description

Adds the channel-change broadcast tracked in #123.

The release-channel preference is process-wide, but an offer produced by a check that already finished is not — it stays cached in the view that produced it. Switching the channel in one window therefore left every other window still advertising a build it would no longer offer: the sidebar kept its update notice, the Settings update row kept its "Install update" button, and an open update dialog kept its offer. Since #97's install-time re-check those were refusals waiting to happen rather than wrong installs, but the user only discovered that by clicking.

`ThemeManager` now broadcasts on a real channel change:

- Listeners register with an **anchor widget** and are dropped once it dies, so a closed window stops being refreshed instead of running update checks forever. The anchor is deliberately separate from the widgets a listener captures — update views hold their own widgets strongly (a button's click handler holding a clone of that button is unavoidable), so anchoring on one of those would never be pruned.
- Writing an unchanged channel is now a no-op, which is what lets a listener drive a selection without looping back into the handler that wrote it.
- Each settings page subscribes: it resyncs its release-channel selector and re-runs its check, which is also what clears that window's sidebar notice.
- A check would reset a row that is mid-install or waiting on a restart back to "Check now", re-offering the build it just installed. The update row now reports whether an install is under way and the listener leaves it alone when it is — the same guard the update dialog applies before withdrawing its offer.
- An open update dialog withdraws its offer, unless an install has already started or finished.
- The selector no longer re-checks directly — the broadcast covers every window, not just its own.

Rebased onto `main` now that #97 has merged, and adapted to the three-way `Stable / Preview / Nightly` segmented control it shipped with.

## Visual evidence

N/A — no new interface elements. The change is *when* existing surfaces update: the sidebar update notice and the Settings → Updates row now clear on a channel change instead of on the next click.

## How to test

1. Build a preview binary so the update surfaces are reachable, and open two Strata windows.
2. In window A, with Settings → Updates → "Automatically check for updates" on, set the release channel to **Preview** and let the check finish, so both windows show a prerelease offer in the sidebar and an "Install update" button in Settings.
3. In window B, open Settings → Updates and set the channel back to **Stable**.
4. Look at window A without clicking anything.
5. Repeat with an update dialog left open in window A (click the sidebar notice first), changing the channel in window B.
6. Start an install in window A, then change the channel in window B while it downloads and again once it shows "Restart now".

**Expected result:** window A's selector follows window B's, its sidebar notice and "Install update" button clear on their own, and it re-checks on Stable. An open dialog in window A withdraws its offer and its action button becomes Close. An install in progress, and a finished install waiting on a restart, are both left alone.

## Related issue

Closes #123


spandan11106:
Marked as draft until #97 merges.

The commit list here is misleading right now: \`main\` has no release-channel feature yet (no \`release_channel.rs\`, no \`Channel\`, no \`offer_still_eligible\`), so this branch is stacked on \`feat/61-release-channels\` and GitHub compares it against \`main\` — hence 20 of #97's commits showing alongside the one commit that belongs to this PR (`3266bcf`). GitHub can't base a PR on a branch that lives in a fork, so there is no way to narrow the diff before #97 lands.

Once #97 merges I'll rebase and force-push, which drops those 20 commits automatically, and mark this ready for review. Until then, `3266bcf` is the only commit to read.

spandan11106:
Two windows of the **same process** (a second `strata` launch adds a window to the running instance), both on Settings → Updates. Toggling "Nightly / preview builds" in the left window moves the right window's switch in the same frame, in both directions, and the right window re-runs its own check - no click in that window.


https://github.com/user-attachments/assets/fd07de5a-51e4-4c7d-a771-2daccedbb908

Not covered: the update dialog withdrawing an offer. That needs a prerelease to be offered and then switched away from, and `lgse/strata` currently publishes none (`prerelease: true` count is 0), so there is nothing for the preview channel to serve — both channels resolve to the same v0.6.1 stable release, which is why the offer text is unchanged after the toggle. The dialog path is covered by the eligibility unit tests instead.

spandan11106:
Now that #97 is merged, this pr can be reviewed and merged.

PR #132, opened 2026-09-01
fix(ui): polish peek popover depth, alignment, and slide animation

JoeJoeflyn:
## Summary
- Header border-radius was `6px 6px 0 0` while the container was `9px` — visible gap between header background and container border. Now matches at `9px 9px 0 0`.
- The popover had only an accent glow with no depth shadow, so it didn't read as a floating surface. Layered `0 6px 20px alpha(@theme_bg, 0.45)` under the existing accent glow.
- Header label used `alpha(@theme_accent, 0.58)` which read as a broken/placeholder color. Switched to `@theme_dim_text` — the established pattern for secondary labels in this app.
- List items touched the top/bottom edges with no breathing room. Added `padding: 6px 0` to the scroll/list area.
- The popover used a `Crossfade` transition, ignoring its spatial origin. Now slides directionally (`SlideRight` when positioned right of the anchor, `SlideLeft` when on the left) so enter and exit follow the same path.
- Fade duration was 100ms — below the 125–200ms range for small popovers. Bumped to 150ms for a more deliberate feel.
- The 4px gap between anchor and popover was too tight, making them visually merge. Widened to 8px.

#### Test plan
- [x] Hover a folder — popover slides in from the anchor direction with the accent glow + depth shadow
- [x] Check header top corners match the container's rounded corners
- [x] Check header label reads as a proper dim label, not a faded accent
- [x] Check list items have vertical breathing room at top and bottom
- [x] Hover a folder near the right edge — popover falls back to the left and slides in from the left
- [x] Move away — popover slides out along the same path
image
screenrecording-2026-09-02_01-21-09

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1479,9 +1479,9 @@ paned > separator {
 
 .peek-popover {
   background: @theme_surface;
-  border: 1px solid alpha(@theme_border, 0.6);
+  border: 1px solid alpha(@theme_border, 0.8);
   border-radius: 9px;
-  box-shadow: 0 0 14px alpha(@theme_accent, 0.24);
+  box-shadow: 0 0 14px alpha(@theme_accent, 0.24), 0 6px 20px alpha(@theme_bg, 0.45);
   color: @theme_text;
   padding: 0;
 }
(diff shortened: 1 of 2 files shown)


l0gicgate:
Wow, man just tested, that is SO clean @JoeJoeflyn!!

PR #139, opened 2026-09-01
fix(preview): default Polaris video previews to software

### LINKED ISSUE #127: GPU hang + full session crash generating video preview (VA-API hardware encode on AMD Polaris/VCE)
### Summary

Generating a video preview for a particular mp4 hung the GPU's hardware video-encode engine (AMD VCE), the kernel's GPU reset failed to recover cleanly ("VRAM is lost due to GPU reset!"), and the entire Wayland session crashed (Hyprland aborted on an OpenGL assertion once its GPU context was lost). A file-manager preview probably shouldn't be able to take down the whole desktop, so I'd like to suggest a software fallback or an option to disable hardware-accelerated preview generation.

### What happened

1. Clicked a video in Strata to preview it.
2. Screen showed corrupted (pink) output, then the session died.

Kernel log at the moment of the crash:

```
amdgpu 0000:01:00.0: ring vce0 timeout, signaled seq=745, emitted seq=746
amdgpu 0000:01:00.0:  Process ffmpeg pid 1031715 thread ffmpeg:cs0 pid 1031716
amdgpu 0000:01:00.0: GPU reset begin!. Source:  1
amdgpu 0000:01:00.0: suspend of IP block <vce_v3_0> failed -22
amdgpu 0000:01:00.0: GPU pre asic reset failed with err, -22 for drm dev, 0000:01:00.0
amdgpu 0000:01:00.0: PCI CONFIG reset
amdgpu 0000:01:00.0: VRAM is lost due to GPU reset!
```

The ffmpeg process the kernel names is Strata's preview job (recovered from its coredump):

```
ffmpeg -nostdin -v error -threads 1 -filter_threads 1 -hwaccel vaapi \
  -hwaccel_device /dev/dri/renderD128 -hwaccel_output_format vaapi -i /input \
  -map 0:v:0 -map 0:a:0? -sn -dn -t 30 \
  -vf scale_vaapi=w=1280:h=1280:force_original_aspect_ratio=decrease:force_divisible_by=2:format=nv12 \
  -c:v h264_vaapi -fpsmax 30 -b:v 2M -maxrate 3M -bufsize 4M \
  -c:a aac -b:a 96k -f mp4 -y /output/result.media
```

So the preview pipeline uses VA-API for both decode and encode (`-c:v h264_vaapi`). The hang happened on the `vce0` ring — the hardware *encoder* — while transcoding the 30s preview clip.

### Why I'm reporting it here rather than (only) to the kernel

The underlying robustness bug is arguably in amdgpu/VCE on this older GPU, and a failed reset is the kernel's fault. But older Polaris-era hardware encoders are known to be temperamental, and Strata currently has no way to avoid them:

- a config option (or automatic fallback) to generate previews with software encode (`libx264` or similar) would confine a bad file to a failed preview instead of a dead session;
- alternatively, detecting a previous preview crash for a file and skipping/falling back on retry would help.

The same file reproduces the hang (I'm avoiding previewing it again since each occurrence kills the session). Happy to test a build with a software fallback.

### System

- Strata 0.4.0 (release tarball, x86_64)
- Arch Linux, kernel 7.1.9-arch1-2
- GPU: AMD Ellesmere (Polaris, RX 500 series), amdgpu driver
- mesa 26.2.1, libva 2.24.1, ffmpeg n9.0.1 (system)
- Hyprland (Omarchy), Wayland

kallupx:
## Description

Makes video-preview acceleration selectable while using a safe default on AMD Polaris GPUs.

Independent reports show Polaris H.264 encoding producing VCE timeouts, GPU resets, session crashes, and data loss across applications including [Vesktop](https://gitlab.freedesktop.org/mesa/mesa/-/work_items/16220), [Discord](https://gitlab.freedesktop.org/mesa/mesa/-/work_items/14270), [RustDesk](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1146343), and [gpu-screen-recorder](https://github.com/omacom/omarchy/issues/9555).

This change:

- Adds persistent hardware-acceleration and Automatic / VA-API / Vulkan settings.
- Defaults an unset acceleration preference to software when a Polaris 10, 11, or 12 render node is detected.
- Keeps the switch available so users can explicitly opt in; their choice then persists.
- Retains enabled Automatic mode as the default on non-Polaris systems.
- Applies preference changes to the next preview without interrupting an active preview.
- Uses the existing explorer-style menu and themed visual states.
- Limits sandbox device exposure according to the selected policy; software previews receive no GPU devices or `/sys`.
- Aligns VA-API and Vulkan output dimensions to 16 pixels while preserving aspect ratio.
- Falls back to software when the selected hardware backend fails.

## Visual evidence

screenshot-2026-09-01_22-04-56

## How to test

1. Open **Settings > General > Video Previews**.
2. On a non-Polaris system with no stored preference, confirm acceleration defaults to On with Automatic selected.
3. On a Polaris system with no stored preference, confirm acceleration defaults to Off while the switch remains available.
4. Enable acceleration and select Automatic, VA-API, and Vulkan in turn.
5. Start a new video preview after each selection and confirm the chosen backend is attempted.
6. Disable acceleration and confirm a new preview uses software.
7. Restart Strata and confirm an explicit On or Off choice persists.

**Expected result:**

Polaris systems start safely in software mode but permit explicit hardware opt-in. Other systems retain accelerated previews by default. Each new preview follows the selected policy and falls back to software when necessary.

## Validation

- `cargo fmt --all --check`
- `cargo check --all-targets --all-features`
- `cargo test --all-targets --all-features` - 263 tests passed

## Related issue

Closes #127


l0gicgate:
Tested locally on an NVIDIA GeForce RTX 5080 plus AMD Granite Ridge Radeon Graphics (non-Polaris), using a fresh isolated XDG config/cache.

Automated checks:
- `cargo fmt --all --check`
- `cargo check --all-targets --all-features`
- `cargo test --all-targets --all-features` (365 passed)
- `cargo deny check`

Manual coverage:
- Confirmed a fresh non-Polaris configuration defaults to acceleration enabled with Automatic selected.
- Tested synthetic landscape H.264/AAC MP4 (854×480, 30 fps) and portrait VP9/Opus WebM (480×854, 24 fps).
- Automatic played both fixtures smoothly with correct aspect ratio and audio.
- Disabling acceleration disabled the backend selector; software preview played successfully.
- Selected VA-API and Vulkan independently; both landscape and portrait previews loaded and played correctly with audio and no visible corruption.
- Confirmed acceleration/backend settings persisted in `settings.toml` and across application restarts.
- Checked the selector's enabled/disabled, selection, focus/hover, and layout states.
- No application warnings/errors or kernel GPU faults appeared during testing.

Limitation: this machine has no Polaris GPU, so the real-hardware Polaris default-to-software path still needs confirmation on the affected RX 500-series system. The Polaris detection logic is covered by the automated tests.

l0gicgate (approved):

PR #144, opened 2026-09-01
fix(settings): persist missing preferences across restarts

### LINKED ISSUE #129: Persist all General settings across restarts
## Summary

Ensure every toggle under **Settings → General** is saved and restored across Strata restarts.

## Current verification

A source audit shows that three of the four current General settings are already persisted through `Preferences` in `~/.config/strata/settings.toml`:

- [x] **Folder peeking** — loaded via `ThemeManager::folder_peeking()` and saved by `set_folder_peeking()`
- [x] **Single-click file previews** — loaded via `single_click_previews()` and saved by `set_single_click_previews()`
- [x] **Open search results directly** — loaded via `search_open_files_directly()` and saved by `set_search_open_files_directly()`
- [ ] **Reduce motion** — currently initialized to `false` in `settings.rs` and only updates the process-local `REDUCE_MOTION` cell, so it resets on restart

Folder peeking persistence was previously fixed in #98 / #99. This issue covers the remaining gap and adds coverage for the General menu as a whole so future settings do not regress.

## Expected behavior

Changing any General setting should take effect immediately, be written to the user's preferences, and be restored on the next launch. The toggle shown after restart must match the effective behavior.

## Acceptance criteria

- Add a backward-compatible persisted `reduce_motion` preference (defaulting to `false` for existing settings files).
- Restore the Reduce motion value early enough at startup that application animations honor it before the Settings dialog is opened.
- Initialize the Reduce motion toggle from the persisted value rather than hard-coding `false`.
- Verify all four General settings by changing each value, restarting Strata, and confirming both toggle state and behavior are restored.
- Add automated serialization/default/round-trip coverage for every General preference to prevent regressions.

### LINKED ISSUE #146: fix(settings): persist hidden files and folders-first preferences
### Strata version

0.6.1 / commit 6991374

### Installation method

Built from source

Omarchy 4.0.0, Hyprland

### What happened?

The Hidden files and Folders first appearance preferences take effect for the current session but reset when Strata restarts. The adjacent view mode, density, sort key, and sort direction preferences are persisted.

### Steps to reproduce

1. Open the Appearance menu and enable Hidden files.
2. Open a column sort menu and disable Folders first.
3. Quit Strata completely and restart it.
4. Reopen the corresponding menus.

### Expected behavior

Both preferences should be restored after restart and reflected in the menus and browser behavior.

### Screenshots or video

N/A — persistence-only behavior.

### Relevant logs

No relevant logs; the application does not report an error.

### Additional context

Discovered while reviewing #144. `ViewPreferences` already carries both values, but `ThemeManager::Preferences` only serializes sort key and direction. Hidden-file changes also do not currently notify preference observers.

kallupx:
## Description

Persist Reduce motion alongside the other General preferences and restore it before the interface is constructed. Also persist the Hidden files and Folders first view preferences discovered during review, and keep the Hidden files menu state synchronized with keyboard changes.

Add backward-compatible defaults and round-trip coverage for the persisted preferences.

## Visual evidence

https://github.com/user-attachments/assets/accb54ab-20d0-4796-b84c-fd40e1230344

## How to test

1. Open Settings → General and enable Reduce motion.
2. Open the Appearance menu and enable Hidden files.
3. Open a column sort menu and disable Folders first.
4. Quit Strata completely and restart it.
5. Reopen Settings, the Appearance menu, and a column sort menu.

**Expected result:**

All three preferences remain selected and effective after restart. Sidebar animations remain disabled, hidden files remain visible, and files retain the selected folder ordering.

## Related issue

Closes #129
Closes #146

### DIFF
--- a/src/app/browser.rs
+++ b/src/app/browser.rs
@@ -206,6 +206,13 @@ impl Browser {
             .push(Rc::new(observer));
     }
 
+    fn notify_preferences_observers(&self) {
+        let preferences = self.preferences.get();
+        for observer in self.preferences_observers.borrow().iter() {
+            observer(preferences);
+        }
+    }
+
     pub fn set_operation_provider(&self, provider: Rc<dyn OperationProvider>) {
         self.operation_provider.replace(Some(provider));
     }
@@ -525,6 +532,7 @@ impl Browser {
         let mut preferences = self.preferences.get();
         preferences.show_hidden = !preferences.show_hidden;
         self.preferences.set(preferences);
+        self.notify_preferences_observers();
 
         let locations = {
             let mut state = self.state.borrow_mut();
@@ -583,10 +591,7 @@ impl Browser {
                 browser.preferences.set(preferences);
                 result
             };
-            let preferences = browser.preferences.get();
-            for observer in browser.preferences_observers.borrow().iter() {
-                observer(preferences);
-            }
+            browser.notify_preferences_observers();
             if let Some((entries, focused, positions)) = result {
                 browser.emit(BrowserEvent::EntriesReplaced { depth, entries });
                 if let Some(focused) = focused {
--- a/src/app/browser/tests.rs
+++ b/src/app/browser/tests.rs
@@ -623,11 +623,21 @@ fn hidden_file_preference_is_applied_to_reloaded_requests() {
     let browser = Browser::new(Rc::new(RecordingFileSource {
         include_hidden: include_hidden.clone(),
     }));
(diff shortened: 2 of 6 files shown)


l0gicgate:
Expanded this PR to cover the two additional persistence gaps found during review:

- **Hidden files** is now serialized, restored, and kept synchronized between the Appearance menu and `Ctrl+H`.
- **Folders first** is now serialized and restored with the existing sort preferences.

Added backward-compatible defaults and preference-observer/round-trip coverage. Tracked in #146.

PR #145, opened 2026-09-01
fix(uri): sanitize embedded remote credentials

### LINKED ISSUE #111: Remote URI authentication parameters can leak into bookmarks and UI
### Strata version

0.6.1 (source build at commit `e00117a79f1ab3d9ee2150a53b87a94300397d8c`).

### Installation method

Built from source.

Omarchy 4.0.2 (Arch Linux), Hyprland 0.56.2, Wayland.

### What happened?

Strata rejects the common `scheme://user:password@host/` form, but accepts GLib's separate URI authentication-parameter form, for example `smb://alice;password=secret@host/share`. GLib parses `password=secret` as `auth_params`, and `GFile::get_uri()` preserves it.

That URI can enter through the address bar or an existing GTK bookmark. Once committed, Strata uses the raw URI for display/tooltips and writes it back to `~/.config/gtk-3.0/bookmarks`, exposing the credential in plaintext. Diagnostic logging already removes `AUTH_PARAMS`, so the current ingress/display/persistence behavior is inconsistent with the established secret boundary.

### Steps to reproduce

Use only a disposable server and credential.

1. Enter a URI such as `smb://alice;password=not-a-real-secret@host/share` for a reachable test server, or add it to an isolated GTK bookmarks file.
2. Navigate to or load the bookmark in Strata.
3. Pin/save the location as a GTK bookmark.
4. Inspect the displayed location/tooltip and the isolated `gtk-3.0/bookmarks` file.
5. Observe that the `password=...` authentication parameter is preserved in plaintext.

### Expected behavior

Every URI ingress should reject embedded passwords and authentication parameters before navigation. Committed locations, display text, history, bookmarks, clipboard data, errors, and logs must contain only a sanitized URI. Authentication should proceed through the GIO/GVFS prompt and desktop keyring.

### Screenshots or video

N/A, no screenshot was captured; the reproduction uses a disposable credential.

### Relevant logs

```shell
GUri parse of smb://alice;password=not-a-real-secret@host/share:
userinfo:    alice;password=not-a-real-secret
auth_params: password=not-a-real-secret
GFile URI:  preserves the auth parameter
```

### Additional context

- Validation environment: GLib 2.88.3 with an isolated GTK bookmarks file.
- `uri_has_embedded_password` checks only for `:` before `@` at `src/services/file_source.rs:89-104`.
- Address-bar navigation relies on that check at `src/app/browser.rs:212-240`.
- `Location::display_path` returns the raw URI at `src/model/mod.rs:121-126`.
- GTK bookmarks are loaded and saved as raw URIs at `src/ui/window.rs:1616-1668`.
- `diagnostic_path` already hides GLib `PASSWORD` and `AUTH_PARAMS` at `src/model/mod.rs:101-117`.
- Closed issue #20 explicitly required that no secret enter GTK bookmarks, history, UI, or logs. Open #64 covers future FTP/WebDAV credential-retention policy, and open #65 covers a future saved-connection model; neither covers this current auth-parameter ingress and GTK-bookmark regression.

#### Acceptance criteria

- [ ] Parse supported URIs once with GLib's password/auth-parameter flags and reject both forms at every ingress, including GTK bookmarks.
- [ ] Never echo the rejected secret in UI or errors.
- [ ] Sanitize or omit legacy bookmarked URIs containing password/auth parameters rather than displaying or rewriting them raw.
- [ ] Tests cover colon passwords, `;password=...` auth params, percent-encoded variants, and safe username-only URIs.

kallupx:
## Description

Parse remote URIs with GLib's password and authentication-parameter flags, including percent-encoded delimiters. Address-bar credentials are removed from the URI immediately, supplied only to the pending mount attempt with password saving disabled, and discarded after that attempt. Failed authentication falls back to the normal sign-in flow.

Route other URI ingress through the same sanitization boundary so clipboard items, dropped files, directory entries, mount roots, and GTK bookmarks cannot introduce credential-bearing locations. Unsafe bookmarks retain their username and location while their secrets are removed on load and save; display sanitization remains a defense for legacy values.

## Visual evidence

N/A: the only visible change is that the address returns to its sanitized form instead of showing a credential-rejection error.

## How to test

1. Enter `smb://alice:secret@host/share`, `smb://alice;password=secret@host/share`, and their percent-encoded delimiter variants in the location bar.
2. Confirm the address becomes `smb://alice@host/share` and the supplied credentials are used for the connection attempt without an initial duplicate prompt.
3. Confirm failed credentials open the normal sign-in dialog and are not remembered unless a storage option is explicitly selected there.
4. Add safe and credential-bearing remote URIs to the GTK bookmarks file, then reopen Strata and save the bookmarks.
5. Confirm the bookmarks remain available with their username but no password or authentication parameters.

**Expected result:** Embedded credentials are consumed only for one connection attempt, are never shown or persisted, and sanitized remote locations remain usable.

## Related issue

Closes #111

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,6 +7,11 @@
 - Write commits and pull request titles in Conventional Commits format: `<type>(optional-scope): <imperative description>`.
 - Keep commits focused. Use `!` and a `BREAKING CHANGE:` footer for breaking changes, and reference the issue in the pull request body.
 
+## Pre-push checks
+
+- Do not push until the full local CI suite passes: `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --all-targets --all-features`.
+- Fix failures before pushing rather than relying on CI for feedback. Keep tests portable across supported environments and avoid assertions that depend on platform-specific URI normalization or other incidental system behavior.
+
 ## Issues and pull requests
 
 - Automated agents must follow the same issue-first workflow and pull request template as human contributors; do not remove or bypass template sections.
--- a/src/adapters/local_files.rs
+++ b/src/adapters/local_files.rs
@@ -15,7 +15,7 @@ use crate::{
     model::{EntryKind, FileEntry, Location, MetadataValue},
     services::{
         DirectoryChange, DirectoryEvent, DirectoryRequest, FileSource, LoadHandle,
-        LocationValidationError, RequestId, backend_unavailable_message,
+        LocationValidationError, RequestId, backend_unavailable_message, validate_uri_credentials,
     },
 };
 
@@ -45,13 +45,16 @@ fn map_validation_error(error: std::io::Error) -> LocationValidationError {
 /// SFTP, ...) can still return a `.path()` via its FUSE mirror even though the
 /// file isn't native; using that path would leak the mirror's opaque
 /// `/run/user/$UID/gvfs/...` location instead of the clean URI (lgse/strata#5).
-pub(crate) fn location_for_file(file: &gio::File) -> Location {
+/// Returns `None` when GIO preserves embedded credentials in that URI.
+pub(crate) fn location_for_file(file: &gio::File) -> Option<Location> {
     if file.is_native()
         && let Some(path) = file.path()
     {
-        return Location::local(path);
+        return Some(Location::local(path));
     }
-    Location::uri(file.uri())
+    let uri = file.uri();
+    validate_uri_credentials(&uri).ok()?;
+    Some(Location::uri(uri))
 }
 
 fn uri_validation_result(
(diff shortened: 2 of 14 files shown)


l0gicgate:
Updated the approach to avoid making secure behavior unnecessarily hostile.

Instead of rejecting an address such as `smb://alice:secret@host/share`, Strata now immediately separates it into:

- the persistent/displayed location: `smb://alice@host/share`
- short-lived credentials used only by the pending GIO mount attempt

The one-shot credentials use `PasswordSave::Never`, never enter `Location`, bookmarks, history, diagnostics, or errors, and are dropped after success or failure. If they fail, the existing authentication dialog is shown so the user can retry and explicitly choose whether to remember the password.

This is safer than merely hiding the original URI in the UI, while being more user-friendly than rejecting credentials the user has already supplied. Non-interactive URI ingress (bookmarks, GIO file lists, and mount roots) is sanitized without retaining the secret.

PR #151, opened 2026-09-01
feat(columns): make miller columns adaptively share available width

### LINKED ISSUE #122: feat(columns): explore adaptive Miller column widths
### Problem or use case

Miller columns currently open at a fixed minimum width, leaving unused space when only one or a few columns are visible. This makes the initial view feel constrained even when the window has ample room and does not progressively communicate the transition from a single-pane view to a multi-column hierarchy.

### Proposed solution

Explore an adaptive width model for Miller columns:

1. With one column open, it fills the available browser width.
2. As child columns open, all visible columns progressively reduce in width and share the available space.
3. Columns continue shrinking until they reach a defined minimum usable width.
4. Once another column cannot fit without violating that minimum, the column strip overflows horizontally and uses horizontal scrolling.
5. Opening or focusing a deeper column keeps the active column visible without unexpected jumps.

The exploration should determine:

- an appropriate minimum width across window sizes and display scaling;
- whether columns divide space equally or use a weighted strategy that favors the active column;
- how adaptive sizing interacts with the existing manual column-resize handles;
- whether a manually resized column remains fixed, resets on navigation, or opts out of automatic sizing;
- how resizing the application window, opening/closing previews, and truncating the navigation path reflows columns;
- whether width transitions should animate while respecting reduced-motion preferences.

The result should preserve readable rows and controls, predictable horizontal scrolling, keyboard navigation, and access to the active column.

### Alternatives considered

- Keep every column at the current fixed default width. This is predictable but underuses space when few columns are open.
- Make only the active column consume leftover space. This preserves context-column widths but may create a more abrupt layout transition.
- Remove manual resizing in favor of fully automatic widths. This simplifies layout behavior but takes control away from users with long filenames or metadata needs.

### Visual examples

N/A — a prototype or short before/after recording should be produced during exploration to compare one-, two-, multi-column, and overflow states.

### Strata version

Current development build (`9a9381f`).

### Additional context

The current implementation initializes each Miller column at a 300 px minimum, allows manual expansion, and places the column strip in a horizontal scroller. This issue is specifically about making widths responsive before the strip reaches its overflow state.

berrimi:
## Description

Adaptive Miller column widths (explored per #122): visible columns now share the available browser width instead of opening at a fixed 300px minimum.

- One column fills the available width; as child columns open, all columns shrink equally until each reaches the 300px usable minimum, then the strip overflows and scrolls horizontally.
- Reflow is driven from the scroller's horizontal adjustment, so columns re-share space on window resize, navigation, and path truncation.
- Manual column resizing and double-click autofit pin a column's width; pinned columns keep their size while the remaining columns share the rest of the space.
- `preview_occupied_width` now tracks rendered column widths so the preview pane is positioned correctly when columns stretch beyond the minimum.
- The active column stays visible in the horizontal scroller when opening or closing columns.

## Visual evidence

N/A — no screenshot captured from this environment; the width behaviors are covered by the unit tests in `src/ui/browser/tests.rs` and the manual steps below produce each visual state.

## How to test

1. `cargo run`, navigate to a folder in the active path.
2. Resize the window and confirm the single column fills the browser width (previously it stayed at 300px).
3. Open child folders and confirm all visible columns share the width and stop shrinking at 300px.
4. Open enough columns that the 300px minimum can't fit, and confirm the strip scrolls horizontally with the active column kept visible.
5. Drag a column's handle to resize it, then open more columns and confirm the pinned column keeps its width while the others adapt.

**Expected result:** visible columns always fill the available width without leaving unused space at the right edge, down to the 300px per-column floor, with manually resized columns opting out of automatic sharing.

Closes #122

### DIFF
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -55,6 +55,7 @@ struct BoundRow {
 #[derive(Clone)]
 struct ColumnView {
     shell: gtk::Box,
+    user_width: Rc<Cell<Option<i32>>>,
     animation_generation: Rc<Cell<u64>>,
     presentation: LoadPresentation,
     model: gtk::StringList,
@@ -490,6 +491,14 @@ impl BrowserView {
         });
         breadcrumb_scroller.add_controller(edit_location);
 
+        let weak_state = Rc::downgrade(&state);
+        let adjustment = state.scroller.hadjustment();
+        adjustment.connect_changed(move |_| {
+            if let Some(state) = weak_state.upgrade() {
+                state.apply_column_widths();
+            }
+        });
+
         Self { state }
     }
 
@@ -546,8 +555,9 @@ impl BrowserView {
             .columns
             .borrow()
             .iter()
-            .map(|column| column.shell.width().max(COLUMN_WIDTH))
+            .map(|column| column.shell.width())
             .fold(0, i32::saturating_add)
+            .min(self.state.scroller.width())
     }
 
     pub fn view_mode(&self) -> BrowserMode {
@@ -4601,12 +4611,16 @@ impl ViewState {
         resize_handle.set_cursor_from_name(Some("col-resize"));
         let resize = gtk::GestureDrag::new();
         resize.set_button(1);
+        let user_width = Rc::new(Cell::new(None));
         let resize_start = Rc::new(Cell::new(COLUMN_WIDTH));
         let pointer_start = Rc::new(Cell::new(None));
         let last_press = Rc::new(Cell::new(0u64));
         let shell_for_resize_start = shell.clone();
-        let shell_for_autofit = shell.clone();
         let column_for_autofit = column.clone();
+        let user_width_for_begin = user_width.clone();
+        let user_width_for_drag = user_width.clone();
+        let weak_for_autofit = Rc::downgrade(self);
+        let weak_for_resize = Rc::downgrade(self);
         let resize_start_for_begin = resize_start.clone();
         let pointer_start_for_begin = pointer_start.clone();
         let last_press_for_begin = last_press.clone();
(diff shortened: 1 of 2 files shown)


src/ui/browser.rs
```diff
@@ -490,6 +491,14 @@ impl BrowserView {
         });
         breadcrumb_scroller.add_controller(edit_location);
 
+        let weak_state = Rc::downgrade(&state);
+        let adjustment = state.scroller.hadjustment();
+        adjustment.connect_changed(move |_| {
+            if let Some(state) = weak_state.upgrade() {
+                state.apply_column_widths();
```
kallupx: Could we also reveal the focused/active column when the viewport shrinks? Otherwise crossing the 300px overflow threshold can leave the active column off screen.

src/ui/browser.rs
```diff
@@ -546,8 +555,9 @@ impl BrowserView {
             .columns
             .borrow()
             .iter()
-            .map(|column| column.shell.width().max(COLUMN_WIDTH))
+            .map(|column| column.shell.width())
```
kallupx: This makes the preview way too small. Automatic columns report all stretched space as occupied, so the preview opens at its 280px minimum. Could we reserve only user_width.unwrap_or(COLUMN_WIDTH) so the preview can reclaim automatic excess width?

kallupx:
Yeah this works well! Thanks @berrimi 

I added two comments, especially the preview functionality needs some attention.

berrimi:
Thanks @kallupx ! Great work on this btw, the aesthetics look really good.

berrimi:
@kallupx I have another idea for previews: why not make it like Yazi? When you hover over a folder or a file, it could just reuse the exact same split/column logic we just implemented to render the preview. That way, we keep the code easy to maintain by reusing the same layout system, but we still get all the benefits of a GUI like the horizontal scrolling and smooth navigation we just built!

kallupx:
Now that you say it, that actually might be the way. @l0gicgate thoughts?

l0gicgate:
@berrimi I like that idea, can you experiment with that concept and show videos of the UI/UX of that?

berrimi:
https://github.com/user-attachments/assets/7ce00b75-971b-478e-97e2-9c3b79113ec8

Here’s a quick demo video of the proof of concept! @l0gicgate  @kallupx  let me know what you think.

l0gicgate:
@berrimi that looks pretty cool, the folder peeking is indistinguishable tho I'm not sure I like it.

Also, you have to think about the Grid mode which also has folder peeking. 

image

berrimi:
Ok, I think I'm going to step back from this feature. The PR is a solid proof of concept, so feel free to use it as a reference or iterate on it later. Thanks for the feedback and great work btw

l0gicgate:
Sounds good @berrimi, I can probably take it over the line. Thank you for putting this together.

l0gicgate:
Closing for now. May revisit in the future.

PR #154, opened 2026-09-01
fix(browser): bound directory loads to stop freezes on huge directories

### LINKED ISSUE #137: Browser freezes while loading or peeking directories with 200,000 entries
## Strata version

Source build at commit `37fe01911be7afcae761936cef6ef3a174c034f8` (PR #72 test branch, based on Strata 0.6.1).

## Installation method

Built from source.

Omarchy, Hyprland 0.56.2, Linux 7.1.8-arch1-3.

## What happened?

Opening a directory containing approximately 200,000 entries makes the ordinary browser UI unresponsive while it loads and renders the directory. Hover-based folder peeking can trigger the same work before the directory is explicitly opened.

In the observed run, buttons stopped responding to hover/input, the loading skeleton stopped animating, and Strata reached approximately 649 MB RSS and 15% CPU after about 11 minutes without completing the view. This also caused unrelated search and Trash dialogs to appear frozen when folder peeking started the large directory load concurrently.

One retry caused the application to disappear, but there was no core dump, OOM event, panic output, or relevant journal entry, so that disappearance cannot be classified as a confirmed crash.

## Steps to reproduce

1. Create a synthetic directory with 200,000 empty files:

   ```shell
   mkdir -p /tmp/strata-large-directory/tree
   python3 - <<'PY'
   import os
   from pathlib import Path

   root = Path('/tmp/strata-large-directory/tree')
   for index in range(200_000):
       descriptor = os.open(root / f'entry-{index:06}.txt', os.O_CREAT | os.O_WRONLY, 0o644)
       os.close(descriptor)
   PY
   ```

2. Open `/tmp/strata-large-directory` in Strata.
3. With folder peeking enabled, hover `tree` long enough to trigger a peek; alternatively, explicitly open `tree`.
4. Attempt to hover or activate header buttons while entries load.

## Expected behavior

Directory browsing and folder peeking should remain responsive on very broad directories. Entry storage/rendering should be bounded or virtualized, and the user should be able to cancel or navigate away promptly rather than waiting for the complete directory to render.

## Screenshots or video

N/A in this report. The freeze is visible as a permanently animating/loading column with all controls unresponsive.

## Relevant logs

```shell
INFO directory load started request_id=3 backend=trash
INFO first directory batch ready request_id=3 entries=128 elapsed_ms=53
# No corresponding completion before the process was stopped.

# Observed after approximately 11 minutes:
# RSS: ~649 MB
# CPU: ~15%

# coredumpctl: No coredumps found.
# Kernel journal: no OOM event.
```

## Additional context

Discovered while manually testing PR #72 with a 200,580-file fixture. Moving the large tree one level below the active search root and disabling folder peeking showed that PR #72's search indexing itself remained responsive; the freeze belongs to the ordinary directory browsing/peek rendering path.

No user data was lost. The fixture was disposable, and Trash was verified empty after testing.

pdwaldrop:
## Description

Directory loading and hover peek both enumerated a location to completion with no cap, so an adversarially large or unbounded directory (200k+ entries, per the issue) blocked the UI for the entire scan. `DirectoryRequest` now carries a `max_entries`/`time_budget` pair. A normal load caps at 100,000 entries / 10s, matching this project's own published baseline (`docs/performance-baseline.md`: 100k entries = 3,755ms / 286 MiB — past that, per-batch merge cost stops feeling responsive). Peek uses a much tighter 64 entries / 3s budget, since a hover preview only ever displays a handful of items and has no reason to enumerate an entire large directory. A load that hits either bound reports itself as truncated, surfaced as a "Partial" indicator in Columns, Grid, and Explorer mode.

Also consolidates the async GLib-main-context test mutex, which had drifted into a separate copy per test module (a latent race flagged in a prior review), into one shared lock in `test_support.rs`. This was needed to write correct tests for the new enumeration budget logic, since `LocalFileSource::enumerate` drives `glib::MainContext::default()` directly and can only be exercised through that same shared context.

Out of scope: `merge_entries()`'s O(n²) per-batch merge cost in `app/navigation.rs` is a separate, already-acknowledged optimization target per the performance baseline doc; fixing it touches shared selection/sort/monitoring code well beyond this issue.

## Visual evidence

Strata pointed at a genuine 200,000-entry directory: it loads responsively (~3s) instead of freezing, and the new "Partial" badge appears in the header. Screenshots to follow in a comment below.

## How to test

1. Create a directory with more than 100,000 entries (e.g. `for i in $(seq 1 200000); do : > "entry-$i.txt"; done`).
2. Open it in Strata (`strata <path>`) in any view mode, and also hover it from its parent to trigger peek.

**Expected result:** the load completes quickly instead of freezing the UI, and a "Partial" badge appears in the header indicating the listing was capped.

## Related issue

Closes #137

### DIFF
--- a/src/adapters/local_files.rs
+++ b/src/adapters/local_files.rs
@@ -181,20 +181,41 @@ impl FileSource for LocalFileSource {
         log_directory_load_started(request_id, &location);
 
         let task = glib::MainContext::default().spawn_local(async move {
+            let deadline = started + request.time_budget;
+            let finish_truncated = |entries: usize, reason: &'static str| {
+                tracing::warn!(
+                    request_id = request_id.0,
+                    entries,
+                    elapsed_ms = started.elapsed().as_millis() as u64,
+                    reason,
+                    "directory load truncated"
+                );
+                emit(DirectoryEvent::Finished {
+                    request_id,
+                    truncated: true,
+                });
+            };
             let directory = location
                 .native_path()
                 .map(gio::File::for_path)
                 .unwrap_or_else(|| gio::File::for_uri(location.uri_value().unwrap_or_default()));
-            let enumerator = match directory
-                .enumerate_children_future(
+            let remaining = deadline.saturating_duration_since(Instant::now());
+            if remaining.is_zero() {
+                finish_truncated(0, "time budget");
+                return;
+            }
+            let enumerator = match glib::future_with_timeout(
+                remaining,
+                directory.enumerate_children_future(
                     ATTRIBUTES,
                     gio::FileQueryInfoFlags::NONE,
                     glib::Priority::DEFAULT,
-                )
-                .await
+                ),
+            )
+            .await
             {
-                Ok(enumerator) => enumerator,
-                Err(error) => {
+                Ok(Ok(enumerator)) => enumerator,
+                Ok(Err(error)) => {
                     tracing::warn!(
                         request_id = request_id.0,
                         error_domain = ?error.domain(),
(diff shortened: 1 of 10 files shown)


pdwaldrop:
05-strata-only
06-badge-closeup-clean

l0gicgate:
@pdwaldrop, thank you for this contribution.

Instead of "Partial", let's put a triangle alert icon, same color as the folder name, right beside the folder name which can be hovered and explains why we're showing the partial directory only.

pdwaldrop:
image
@l0gicgate Is this hover message acceptable or should it change in any way?

pdwaldrop:
@l0gicgate Done in 67c28e5 — swapped the "Partial" text badge for a triangle-alert icon placed directly beside the folder name, colored to match it (both driven by the same accent token via `assets::primary_icon`), in both Columns and Grid/Explorer mode headers. Hover still shows the same explanatory tooltip.

image

image

PR #164, opened 2026-09-02
feat(settings): add show permanently delete toggle and menu entry

### LINKED ISSUE #163: Add Permanently Delete option to context menu and General settings
### Problem or use case
Strata currently only offers 'Move to Trash' (or 'Permanently delete' inside Trash) in the right-click menu. Shift+Delete works for permanent delete, but there is no visible permanent option and no setting like Nautilus's 'Permanently Delete' toggle. Users who want files removed instantly via right-click have no discoverable way.

### Proposed solution
- Add a toggle in **General settings** (BROWSING section) — e.g. 'Show permanently delete' / 'Permanently delete' — default off, persisted in settings.toml.
- When enabled, show an additional 'Permanently delete' (Shift+Del) entry alongside 'Move to Trash' in the file context menu for single and multiple selections (outside Trash). Inside Trash keep single 'Permanently delete'.
- Keep Del -> Trash and Shift+Del -> permanent via existing confirmation dialog; the toggle only controls menu visibility (matching Nautilus).

### Alternatives considered
- Replace 'Move to Trash' entirely when toggle on — more destructive, not matching Nautilus.
- Always show both — clutters menu for users who never need it.

### Additional context
Follows terminal feature #160/#161. Branch will be `feat/permanently-delete-option`.

Deoxizn:
## Description

Add safe Nautilus-style permanently delete. Adds `Show permanently delete` toggle to **General settings** (BROWSING, default off, persisted in `settings.toml` via `ThemeManager::show_permanent_delete`) and shows an additional `Permanently delete (Shift+Del)` entry alongside `Move to Trash (Del)` in the file context menu for single and multiple selections outside Trash. Inside `trash:///` keeps single permanent entry. `Del` still moves to trash and `Shift+Del` permanently deletes via existing `confirm_delete` + `show_delete_confirmation` dialog; the toggle only controls menu visibility and is evaluated live at popup.

## Visual evidence

- General page: `Show permanently delete` switch under Browsing (between `Open search results directly` and `Motion`).
- Right-click single file (outside Trash, toggle off): only `Move to Trash` at bottom.
- Right-click single file (toggle on): `Move to Trash (Del)` + `Permanently delete (Shift+Del)` (danger) at bottom.
- Same for multi-select; Trash view unchanged.

## How to test

1. `cargo run` then open Settings -> General -> enable `Show permanently delete`.
2. Right-click a file (not in Trash) -> menu shows both `Move to Trash` and `Permanently delete`; click `Permanently delete` -> confirmation `Permanently delete 1 item?` -> deletes instantly.
3. Select multiple files -> right-click -> same two entries appear; `Permanently delete` removes all.
4. Disable toggle -> right-click shows only `Move to Trash`; `Shift+Del` still triggers permanent confirm.
5. Navigate to `trash:///` -> right-click -> only `Permanently delete` (no duplicate) regardless of toggle.
6. Check `~/.config/strata/settings.toml` persists `show_permanent_delete = true/false` across restart.

**Expected result:** Toggle controls extra menu entry; trash is still default, permanent requires explicit Shift+Del or extra menu click with danger confirmation.

## Related issue

Closes #163

### DIFF
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -5352,9 +5352,13 @@ pub(super) fn install_item_context_menu(
     } else {
         "Move to Trash"
     };
-    let move_to_trash =
-        item_context_danger_option(crate::assets::icons::TRASH, delete_label, "Del");
-    move_to_trash.add_css_class("danger");
+    let move_to_trash = if in_trash {
+        let option = item_context_danger_option(crate::assets::icons::TRASH, delete_label, "Del");
+        option.add_css_class("danger");
+        option
+    } else {
+        item_context_option(crate::assets::icons::TRASH, delete_label, "Del")
+    };
     let permanent_delete = item_context_danger_option(
         crate::assets::icons::TRASH,
         "Permanently delete",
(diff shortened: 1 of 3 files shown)


l0gicgate:
@Deoxizn cool! I think we don’t need it to be a setting. We should just have it be there by default. There’s a confirmation dialog anyway. We’re all adults, use at your own risk.

Deoxizn:
I added the setting as a safe "fallback" as it was highly recommended but I
agree

l0gicgate:
Updated this based on the latest feedback:

- **Move to Trash** now uses the regular theme text and icon color.
- **Permanently delete** remains styled as destructive.
- Removed the **Show permanently delete** setting and its persisted preference.
- **Permanently delete** is now always available outside Trash; Trash still shows only one permanent-delete action to avoid duplication.
- The existing confirmation dialog remains in place.

Updated context menu showing neutral Move to Trash and destructive Permanently delete

PR #166, opened 2026-09-02
fix(preview): keep sandbox input extensions so RAW files decode

### LINKED ISSUE #165: fix: Sony ARW previews fail in the sandbox
**Strata version**  
`0.7.0` `e840c7d474e4ed989d51bc6ef2971783985057dc`

**Installation method**  
Built from source

Arch Linux, bubblewrap 0.11.2, ImageMagick 7.1.2-30 (libraw), libraw 0.22.2-1

---

**What happened?**

Selecting a Sony `.ARW` file shows:

```
Preview unavailable
The sandboxed preview renderer failed
```

Thumbnails for the same files also fail (fallback icon). The files are valid: ImageMagick identifies and converts them outside Strata, and ImageMagick also succeeds inside a Strata-like bubblewrap **if the filename keeps a `.ARW` suffix**.

This is not a missing package or a bubblewrap block on `/run/media`.

---

**Steps to reproduce**

1. Open a folder of Sony `.ARW` files (example: ILCE-7RM2, ~8000×5320, ~86 MiB).
2. Select one file so the preview pane loads.
3. Observe “Preview unavailable” / “The sandboxed preview renderer failed”.
4. Confirm the grid thumbnail stays on the generic file icon.

---

**Expected behavior**

The preview pane and folder thumbnails show a rendered image for camera RAW, using the existing Pixbuf → ImageMagick → embedded-thumbnail fallbacks.

---

**Screenshots or video**  
N/A — the preview pane shows the standard unavailable-preview message.

---

**Relevant logs**

Helper (when invoked as `thumbnail-raw`) ends with:

```
Preview helper failed: No embedded RAW thumbnail could be decoded
```

That message is the **last** backend, not an early abort. ImageMagick already ran and failed.

Reproduced with Strata’s sandbox shape (`--clearenv`, `/usr` ro-bind, 512 MiB tmpfs, `prlimit --as=2G --cpu=10 --fsize=512M`):

```
magick /input -auto-orient -thumbnail 256x256 png:-
# fails:
# magick: TIFF directory is missing required "ImageLength" field.

magick /input.ARW -auto-orient -thumbnail 256x256 png:/output/test.png
# succeeds: PNG 256×171

dcraw_emu -e -c file.ARW
# fails: Unknown option "-e"; Non-numeric argument to "-c"
# Arch has no classic `dcraw`. LibRaw’s equivalent is `simple_dcraw -e`.
```

The file’s MIME is `image/tiff` (ARW is a TIFF container). ImageMagick then picks the TIFF coder unless the path ends in `.ARW` (or is forced with `dng:`).

---

**Additional context**

Root cause: [`sandbox_command`](src/sandbox.rs) bind-mounts every file as extensionless `/input`. ImageMagick sniffs TIFF magic and uses the TIFF coder instead of libraw.

Second issue: the preview pane uses `preview-image` → GDK Pixbuf only. It never reaches the ImageMagick / dcraw fallbacks that `thumbnail-raw` already has.

Third issue: `render_dcraw` calls `dcraw_emu -e -c`, which is incompatible with LibRaw 0.22’s `dcraw_emu`. `simple_dcraw -e` extracts the embedded JPEG (1616×1080 on this body) but writes `file.thumb.jpg` beside the input, so it needs a writable `/tmp` path, not the read-only `/input` bind.

ogarza:
## Description

The sandbox bind-mounted every file as extensionless `/input`, so ImageMagick treated Sony `.ARW` (TIFF-container RAW) as TIFF and failed. The preview pane also skipped the ImageMagick/dcraw fallbacks that thumbnails already had, and the last fallback called LibRaw's `dcraw_emu` with classic `dcraw -e -c` flags.

Keep a sanitized original extension (`/input.ARW`), run `preview-image` through `render_raw`, and extract embedded thumbnails with classic `dcraw` or LibRaw `simple_dcraw`.

## Visual evidence

image
image

Closes #165

### DIFF
--- a/src/sandbox_helper.rs
+++ b/src/sandbox_helper.rs
@@ -128,10 +128,13 @@ fn render_simple_dcraw(path: &Path, size: i32) -> Result<Vec<u8>, String> {
         return Err("simple_dcraw failed".to_owned());
     }
     for thumb in ["/tmp/raw-thumb.thumb.jpg", "/tmp/raw-thumb.thumb.ppm"] {
-        let Ok(data) = fs::read(thumb) else {
+        let Ok(file) = fs::File::open(thumb) else {
             continue;
         };
-        if data.is_empty() || data.len() as u64 > MAX_OUTPUT_BYTES {
+        let Ok(data) = read_limited(file, MAX_OUTPUT_BYTES) else {
+            continue;
+        };
+        if data.is_empty() {
             continue;
         }
         if let Ok(png) = scale_embedded_thumbnail(&data, size) {
@@ -538,31 +541,38 @@ fn render_media(path: &Path, size: i32) -> Result<Vec<u8>, String> {
     }
 }
 
+fn read_limited(reader: impl Read, max_bytes: u64) -> io::Result<Vec<u8>> {
+    let mut data = Vec::new();
+    reader
+        .take(max_bytes.saturating_add(1))
+        .read_to_end(&mut data)?;
+    if data.len() as u64 > max_bytes {
+        return Err(io::Error::other(
+            "Preview provider output exceeded its limit",
+        ));
+    }
+    Ok(data)
+}
+
 fn bounded_output(command: &mut Command, max_bytes: u64) -> io::Result<Output> {
     let mut child = command
         .stdin(Stdio::null())
         .stdout(Stdio::piped())
         .stderr(Stdio::null())
         .spawn()?;
-    let mut stdout = Vec::new();
     let read = child
         .stdout
         .take()
-        .ok_or_else(|| io::Error::other("Unable to capture provider output"))?
-        .take(max_bytes.saturating_add(1))
-        .read_to_end(&mut stdout);
-    if let Err(error) = read {
-        let _killed = child.kill();
-        let _waited = child.wait();
-        return Err(error);
-    }
-    if stdout.len() as u64 > max_bytes {
-        let _killed = child.kill();
-        let _waited = child.wait();
-        return Err(io::Error::other(
-            "Preview provider output exceeded its limit",
-        ));
-    }
+        .ok_or_else(|| io::Error::other("Unable to capture provider output"))
+        .and_then(|stdout| read_limited(stdout, max_bytes));
+    let stdout = match read {
+        Ok(stdout) => stdout,
+        Err(error) => {
+            let _killed = child.kill();
+            let _waited = child.wait();
+            return Err(error);
+        }
+    };
     let status = child.wait()?;
     Ok(Output {
         status,
--- a/src/sandbox_helper.rs
+++ b/src/sandbox_helper.rs
@@ -44,7 +44,7 @@ pub(crate) fn run(arguments: &[String]) -> Result<(), String> {
         "thumbnail-raw" => (render_raw(input, value.clamp(16, 256))?, None),
         "thumbnail-pdf" => (render_pdf_thumbnail(input, value.clamp(16, 256))?, None),
         "thumbnail-video" => (render_media(input, value.clamp(16, 256))?, None),
-        "preview-image" => (render_pixbuf(input, 1400)?, None),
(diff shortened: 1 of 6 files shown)


src/sandbox_helper.rs
```diff
@@ -93,42 +93,77 @@ fn render_imagemagick(path: &Path, size: i32) -> Result<Vec<u8>, String> {
...
+        .stdin(Stdio::null())
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .status()
+        .map_err(|error| error.to_string())?;
+    if !status.success() {
+        return Err("simple_dcraw failed".to_owned());
+    }
+    for thumb in ["/tmp/raw-thumb.thumb.jpg", "/tmp/raw-thumb.thumb.ppm"] {
+        let Ok(data) = fs::read(thumb) else {
```
kallupx: fs::read allocates the entire untrusted extractor output before enforcing MAX_OUTPUT_BYTES. You should use a bounded read so the 32 MiB limit applies during allocation.

src/sandbox_helper.rs
```diff
@@ -93,42 +93,77 @@ fn render_imagemagick(path: &Path, size: i32) -> Result<Vec<u8>, String> {
...
+}
+
+fn render_simple_dcraw(path: &Path, size: i32) -> Result<Vec<u8>, String> {
+    use std::os::unix::fs::symlink;
+
+    // Writes `<file>.thumb.jpg` next to the input, which is a read-only bind.
+    let staging = Path::new("/tmp/raw-thumb");
+    let _ = fs::remove_file(staging);
+    symlink(path, staging).map_err(|error| error.to_string())?;
+    let status = Command::new("simple_dcraw")
```
kallupx, line 119: This replaces dcraw_emu with simple_dcraw, but README.md and docs/preview-sandbox.md still advertise dcraw_emu. Please update those references.

src/sandbox_helper.rs
```diff
@@ -44,7 +44,7 @@ pub(crate) fn run(arguments: &[String]) -> Result<(), String> {
         "thumbnail-raw" => (render_raw(input, value.clamp(16, 256))?, None),
         "thumbnail-pdf" => (render_pdf_thumbnail(input, value.clamp(16, 256))?, None),
         "thumbnail-video" => (render_media(input, value.clamp(16, 256))?, None),
-        "preview-image" => (render_pixbuf(input, 1400)?, None),
+        "preview-image" => (render_raw(input, 1400)?, None),
```
kallupx, line 47: Could we add a regression test for this routing change? Looks like the current tests cover suffix preservation and thumbnail scaling, but not that preview-image uses the RAW fallback.

kallupx:
Hey, thanks a lot for the PR @ogarza 

I left a few comments, could you take a look?

kallupx (changes requested):

ogarza:
Thanks, addressed the three notes:

- `simple_dcraw` thumbnails now go through `read_limited` so the 32 MiB cap applies while reading, same as command stdout.
- Docs no longer mention `dcraw_emu`. README says ImageMagick, classic `dcraw`, and LibRaw `simple_dcraw`.
- Added `preview_image_uses_raw_fallbacks` so `preview-image` cannot silently go back to pixbuf-only.

src/sandbox_helper.rs
ogarza: Done in 3ec8659

src/sandbox_helper.rs
ogarza: Done in f0fa823 and 31d9622

src/sandbox_helper.rs
ogarza: Done in 4289611

kallupx (approved):
All three earlier issues are fixed at 31d9622

Tested that ARW preview and thumbnail work correctly, looks good to me.

PR #167, opened 2026-09-02
feat(preview): custom themed video controls with keyboard shortcuts

JoeJoeflyn:
## Summary
- Replace GTK's default media controls with a custom Strata-themed control bar (play/pause, seek slider, time display, volume toggle + slider)
- Add click-to-pause on the video and a center play/pause button overlay
- Add keyboard shortcuts: Space (play/pause), Left/Right (seek ±5s), Up/Down (volume ±10%), M (mute/unmute)
- Persist mute state and volume preference across restarts
- Apply to both GIF and non-GIF video previews (GIFs get minimal bar — play/pause only)
- Add Lucide play, pause, volume-2, volume-x icons (strata- namespaced, ISC attribution preserved)

#### Test plan
- [x] Select a non-GIF video — auto-starts, custom bar visible, no GTK default controls
- [x] Click video to pause/play — center play button appears when paused
- [x] Drag seek bar — seeks on release, time label updates
- [x] Volume slider and mute toggle stay in sync (muting drops slider to 0, unmuting restores)
- [x] Keyboard: Space, Left/Right, Up/Down, M all work without closing preview or moving selection
- [x] Select a GIF — minimal bar with play/pause only, loops, click to pause
- [x] Restart app — mute and volume preferences persist
screenrecording-2026-09-02_15-36-46


src/ui/preview.rs
```diff
@@ -754,12 +877,202 @@ impl PreviewState {
...
+
+        bar.append(&play_button);
+
+        let media_for_notify = media.clone();
+        let play_button_for_notify = play_button.clone();
+        let play_icon_for_notify = play_icon.clone();
+        let pause_icon_for_notify = pause_icon.clone();
+        let center_for_notify = center_play.clone();
+        let media_for_playing = media.clone();
+        media_for_notify.connect_notify_local(Some("playing"), move |_, _| {
```
kallupx: This keeps old media streams alive because the signal is owned by media, while the closure captures a strong clone of it. The timestamp/duration handlers below do the same. I tried this on my own env and reproduced this: RSS grew from ~596 MiB to ~1.85 GiB after opening and closing several videos. Could we use weak captures or disconnect these handlers when clearing the preview?

src/ui/window.rs
```diff
@@ -478,6 +478,24 @@ fn install_keyboard_navigation(
         let alt = modifiers.contains(gtk::gdk::ModifierType::ALT_MASK);
         let control = modifiers.contains(gtk::gdk::ModifierType::CONTROL_MASK);
         let shift = modifiers.contains(gtk::gdk::ModifierType::SHIFT_MASK);
+        if preview.has_video()
```
kallupx: Could we skip video shortcuts while a text field is active? This block runs before the rename, new-entry, location, and filter guards, so m/space/arrows control the video instead of editing the filename or path.

src/ui/preview.rs
```diff
@@ -754,12 +877,202 @@ impl PreviewState {
...
+            let last_volume = last_volume.clone();
+            let updating = updating_slider.clone();
+            move || {
+                let muted = !media.is_muted();
+                updating.set(true);
+                if muted {
+                    if slider.value() > 0.0 {
+                        last_volume.set(slider.value());
+                    }
+                    set_preview_volume(&media, &preferences, &Some(slider.clone()), &icon, 0.0);
```
kallupx: Muting here also saves the preferred volume as zero, so the previous level is lost after a restart.

Repro: set 70%, mute, restart then unmute returns at 10%. Maybe mute should update only preview_muted and leave the saved volume unchanged.

src/ui/preview.rs
```diff
@@ -754,12 +877,202 @@ impl PreviewState {
...
+            toggle_volume_for_click();
+        });
+
+        let media_for_volume = media.clone();
+        let icon_for_volume = volume_icon.clone();
+        let preferences_for_volume = preferences.clone();
+        let last_volume_for_volume = last_volume.clone();
+        let updating_for_volume = updating_slider.clone();
+        let slider_for_volume = volume_slider.clone();
+        volume_slider.connect_value_changed(move |scale| {
```
kallupx, line 1070: This runs on every slider change, and set_preview_volume synchronously rewrites and fsyncs the preferences file. Dragging the slider produces a stream of fsyncs on the GTK thread. One improvement would be to debounce this or save once when the drag ends.

kallupx:
Very nice style! I added a few comments on perf and small bugs.

kallupx (changes requested):

JoeJoeflyn:
can u take another look @kallupx

src/ui/preview.rs
```diff
@@ -754,12 +884,222 @@ impl PreviewState {
...
+            if media_for_volume.is_muted() != muted {
+                set_preview_mute(
+                    &media_for_volume,
+                    &icon_for_volume,
+                    &preferences_for_volume,
+                    muted,
+                );
+            }
+            if let Some(prev) = save_slot.borrow_mut().take() {
+                prev.remove();
```
kallupx, line 1086: The timer ID stays stored after the timer has already finished. On the next volume change, this tries to remove the old source and aborts the app.

Repro: move the slider, wait about a second, then move it again. It crashes with Failed to remove source.

I think clearing the stored ID when the timeout fires should fix it.

src/ui/preview.rs
```diff
@@ -205,6 +208,71 @@ impl PreviewDrawer {
...
+                let volume = (preferences.preview_volume() + delta).clamp(0.0, 1.0);
+                set_preview_volume(&media, &preferences, &slider, icon, volume);
+                true
+            }
+            gtk::gdk::Key::m | gtk::gdk::Key::M => {
+                let muted = !media.is_muted();
+                if muted {
+                    media.set_volume(0.0);
+                    if let Some(slider) = slider.as_ref() {
+                        slider.set_value(0.0);
```
kallupx: Seems the M shortcut still saves the preferred volume as zero. Setting the slider to 0.0 fires value-changed, but this path doesn’t use the updating_slider guard from the button handler.

After the delayed save, restarting and unmuting will still restore 10% instead of the previous volume. Could this use the same guard as the button?

kallupx (changes requested):
The stream cleanup and text-field shortcut fixes look good, but I was able to reproduce a crash in the new volume debounce. Muting with M key also still overwrites the saved volume with zero. I left more details in the inline comments.

JoeJoeflyn:
@kallupx can you take another looks

kallupx (approved):
All comments have been addressed, tested that everything works manually.

PR #173, opened 2026-09-02
feat(browser): add manual refresh and auto-refresh interval

### LINKED ISSUE #172: Add manual refresh (F5 / button) and optional auto-refresh interval
### Problem or use case
Strata's main window has no manual refresh control. Nautilus offers F5 and a header button; Windows file managers offer selectable auto-refresh intervals. While Strata watches directories via GIO monitors, monitors can miss events on network shares or after errors, leaving the pane stale until navigation. Users expect F5/Ctrl+R and a visible refresh affordance next to search.

### Proposed solution
- Header: refresh button next to search (tooltip Refresh (F5), icon rotate-cw/refresh) that reloads the active pane(s).
- Keybinds: F5 and Ctrl+R reload current location(s). Also reconsider Ctrl+T for terminal vs refresh binding collision.
- Settings: General -> optional auto-refresh interval selector (Off, 1min, 5min, 10min) that periodically reloads visible columns when no monitor event arrives. Persisted in settings.toml.
- Use existing Browser::refresh / reload_column path so stale cancellation and batched loading remain bounded.

### Alternatives considered
- Always rely on monitors only — fails on flaky GVfs/network.
- Auto-refresh always 5min hard-coded — should be selectable/off.

### Additional context
Follow-up to #163/#164. Branch will be feat/refresh-pane.

Deoxizn:
## Description

Add manual refresh and optional auto-refresh, matching Nautilus F5 and Windows interval selector. Adds a header refresh button next to search (tooltip Refresh (F5), `strata-refresh` Lucide icon) that reloads the current pane(s), `F5` / `Ctrl+R` accelerators via `win.refresh` action + keyboard capture, and a General settings auto-refresh interval selector.

`Browser::refresh_all()` iterates all open columns and calls `refresh_column()` using the existing bounded `ColumnReloaded` / `request_directory` path so cancellation and batch loading remain correct. `BrowserView::refresh()` delegates, and `BrowserView::set_auto_refresh_interval(secs)` manages a `glib::timeout_add_local` that periodically calls `refresh_all()` (0 = off). Initial interval is loaded from `ThemeManager::auto_refresh_interval()` on window present and updated live from settings.

New `strata-refresh` icon (Lucide `rotate-cw`) is recolored via `assets` like other `strata-*` icons.

## Visual evidence

- Header: Search | Refresh (↻) | Views | Settings | Close — Refresh uses `strata-refresh` at 20px with `header-action` styling.
- Settings -> General -> REFRESH section with segmented control Off / 1 min / 5 min / 10 min.
- Keybindings page lists Refresh F5 / Ctrl+R.

## How to test

1. `cargo run`
2. Click header Refresh or press `F5` / `Ctrl+R` -> active pane shows loading skeleton then reloads; `cargo` logs show `refresh_all`.
3. Open Settings -> General -> REFRESH -> select 1 min -> wait 60s -> pane auto-reloads; set Off -> no auto reload.
4. Test in Columns (multiple panes), Grid, Explorer - all refresh correctly.
5. Check `~/.config/strata/settings.toml` persists `auto_refresh_interval = 60|300|600|0` across restart.
6. Navigate to `trash:///` or `smb://` -> Refresh still works via same path.

**Expected result:** Manual refresh always reloads; auto-refresh respects interval and can be disabled.

## Related issue

Closes #172


src/ui/settings.rs
```diff
@@ -392,6 +392,48 @@ fn general_page(browser: &BrowserView, manager: Rc<ThemeManager>) -> gtk::Widget
...
+    refresh_copy.append(&refresh_title);
+    refresh_copy.append(&refresh_desc);
+    refresh_row.append(&refresh_copy);
+    refresh_row.append(&control);
+    preferences.append(&refresh_row);
+    for (idx, button) in buttons.iter().enumerate() {
+        let manager = manager.clone();
+        let browser = browser.clone();
+        let secs = secs[idx];
+        let buttons_clone = buttons.clone();
```
kallupx: These buttons are already grouped, so they deactivate each other automatically. Capturing the whole list here also creates a reference cycle that can keep the browser and timer alive after closing the window. Can we remove this clone and the manual loop?

src/ui/browser.rs
```diff
@@ -978,6 +988,25 @@ impl ViewState {
...
+        if let Some(source) = self.auto_refresh.take() {
+            source.remove();
+        }
+        if secs == 0 {
+            return;
+        }
+        let weak = Rc::downgrade(&self.browser);
+        let source = glib::timeout_add_local(Duration::from_secs(u64::from(secs)), move || {
+            if let Some(browser) = weak.upgrade() {
+                browser.refresh_all();
```
kallupx, line 1019: Could we postpone auto-refresh while a rename or new-entry editor is open? Reloading clears the model but leaves the edit state active, so the editor and typed text can disappear.

src/app/browser.rs
```diff
@@ -1539,6 +1539,20 @@ impl Browser {
         }
     }
 
+    pub fn refresh_all(self: &Rc<Self>) {
+        let depths: Vec<usize> = {
+            let state = self.state.borrow();
+            (0..state.columns.len()).collect()
```
kallupx, line 1547: When switching from Columns to Grid or Explorer, the old columns remain in state, so this also refreshes hidden panes. Is that intentional? The issue mentions refreshing visible columns.

kallupx (changes requested):
Tests pass, and formatting is clean. Looks good overall, I left some comments.
 
Issues found during testing:
A focused GTK check confirmed the toggle reference cycle. I also traced auto-refresh through ColumnReloaded and found that the model is cleared while rename state remains active.

l0gicgate (changes requested):
Please add visual screenshots/videos when the UI changes.

Deoxizn:
Thanks @kallupx @l0gicgate for the review — pushed fix **ca72f35** addressing all three inline threads:

**1. Settings toggle reference cycle (r3915909383):**  
Removed `buttons_clone` and the manual `for (other_idx, other)… set_active(false)` loop. Buttons are already grouped via `set_group` (GTK handles exclusive deactivation), so we now just `if toggled.is_active() { manager.set…; browser.set…; }` with only `manager`/`browser`/`secs` captures. No longer keeps browser/timer alive after window close.

**2. Auto-refresh during rename (r3915911224):**  
Changed `ViewState::set_auto_refresh_interval` to `self: &Rc<Self>` and now captures `weak_state` + `weak_browser`. Timeout checks `active_rename`/`active_new_entry` and `mode_views.rename/new_entry_is_active()` before reloading — if an editor is open it returns `Continue` without `ColumnReloaded`, so the model isn’t cleared and typed text isn’t lost. Next interval will retry.

**3. Refreshing hidden panes (r3915913227):**  
`BrowserView::refresh()` is now mode-aware: Columns -> `refresh_all()`, Grid/Explorer -> `reload_active()`. Auto-refresh uses the same branch via `state.mode_views.borrow().mode()`. Hidden columns after a mode switch are no longer refreshed unnecessarily.

**Visual evidence (l0gicgate):** will attach before/after screenshots of the header Refresh button (`strata-refresh` next to Search) and General -> REFRESH segmented control (Off/1/5/10 min) plus a short clip of F5/Ctrl+R reload. The description already lists expected visuals; updating shortly.

Ready for re-review — `cargo fmt --check`, `clippy -D warnings`, `cargo test` 290 passed on `ca72f35`.

l0gicgate:
Heads-up for merge sequencing with #180: that PR adds relative modified-date labels (`just now`, `Xm ago`, etc.). Those labels are currently computed only when Explorer rows, the preview, or Properties are populated, so they become stale while a window remains open. Since this PR already owns refresh timing, could it also trigger a lightweight metadata-label repaint about once per minute, including when filesystem auto-refresh is Off? Re-reading every directory is not necessary for this and would leave the default-Off case stale.

l0gicgate:
Update: we decided to handle relative-date label repainting directly in #180, so no change is needed here for that integration.

l0gicgate:
Implemented the requested refresh placement in `357f02c`:

- Removed Refresh from the global window header.
- Added a pane-scoped Refresh action that reloads only the associated directory. In Explorer it sits left of Filter; in Grid it sits left of the thumbnail-size control; and in List/column mode it sits before the sorting controls.
- Added **Refresh (F5)** to the folder-background context menu, using the same pane-scoped reload.

### Explorer mode
pi-clipboard-c7c36473-3795-44b1-b7cb-1793759e86c4


### Grid mode

pi-clipboard-27f8db41-6f80-40ad-8358-7dce6f4d8767


### Folder context menu
image

l0gicgate (dismissed):
No blocking issues found in the updated diff.

The refresh controls now target their pane, the folder-background menu uses the same depth-specific reload, and the periodic refresh avoids active rename/new-entry editors without retaining the view. The merge resolution also preserves both auto-refresh and release-channel preferences. Local formatting, Clippy, and all 375 tests pass, and CI is green.

Non-blocking: the PR description still describes the old global header placement and should be updated before merge.

PR #183, opened 2026-09-02
build(arch): add stable and rc AUR packages

### LINKED ISSUE #50: build(arch): add an AUR package
## Summary

Provide an Arch Linux package for Strata, with the AUR as the initial distribution channel. Keep the packaging structure reusable enough that additional distro packages can follow without expanding this issue beyond Arch.

The first package should preferably be `strata-bin`, built from the signed upstream release archives. A source-built `strata` AUR package can be added separately if there is demand. Inclusion in Arch's official repositories is a downstream process and is not required for this issue.

## Scope

- Add and maintain an Arch `PKGBUILD` and `.SRCINFO` for stable Strata releases.
- Package the correct upstream archive for supported architectures (`x86_64` and `aarch64`) with pinned checksums.
- Declare required runtime dependencies accurately, with preview/RAW integrations represented as optional dependencies where appropriate.
- Install the binary, license files, desktop entry, and application icon into standard system paths.
- Coordinate desktop metadata with #49.
- Ensure package-managed installations do not attempt to replace `/usr/bin/strata` through the in-app updater; updates should be delivered by `pacman`/an AUR helper.
- Validate the package in a clean Arch environment and document the release/update process.
- Publish the package to the AUR, including any required maintainer setup, after review.

## Acceptance criteria

- `makepkg --cleanbuild --syncdeps` succeeds in a clean Arch environment.
- `namcap` reports no unexplained actionable errors for the package or `PKGBUILD`.
- Installing through `pacman -U` provides a working `strata` command and desktop launcher.
- Package-managed builds disable or hide self-update installation: Strata must not attempt to replace package-owned files, and users are directed to update through `pacman` or their AUR helper.
- Uninstalling the package removes all package-owned files cleanly.
- Both declared architectures use the matching release artifact and checksum.
- A documented maintainer workflow updates `pkgver`, checksums, `.SRCINFO`, and the AUR package for each stable release.
- The package is available from the AUR or the exact external publication blocker is documented.

## Future distro packaging

Strata can also be packaged for other distributions. Debian/Ubuntu (`.deb`), Fedora (`.rpm`), Flatpak, and AppImage each have different dependency, sandboxing, update, and review requirements, so they should be tracked as separate follow-up issues. Shared desktop metadata and install layout from #49 should be designed to support those formats.

spandan11106:
## Description

Adds official `strata-bin` and `strata-rc-bin` AUR packages generated from one template. Stable tracks the latest final release; the Preview package tracks the newest non-nightly release without downgrading when an older stable release follows a newer prerelease.

A package marker prevents in-place replacement of pacman-owned binaries and locks the displayed channel to the installed package. Update checks query the AUR first, so users are notified only after their package has actually been published. The marker is intentionally scoped to Strata's official AUR packages while remaining forward-compatible with unknown fields.

Non-nightly releases now regenerate package metadata and open a review PR through the protected `package-automation` environment. Merging reviewed package metadata publishes both AUR repositories through the protected `aur-production` environment.

## How to test

1. Run `python3 scripts/update_aur.py --stable 0.9.0 --preview 0.9.0 --check`.
2. Build either package with `makepkg --cleanbuild --syncdeps --force` and confirm its binary matches the corresponding release artifact.
3. Install a package, open **Settings → Updates**, and confirm the channel is locked and no update appears until a newer package version is present in the AUR.
4. Run the AUR publishing workflow against already-current package repositories and confirm it exits without creating duplicate commits.

Expected result: package metadata is reproducible, package-owned installs never self-update, notices follow AUR availability, and reviewed metadata is published by GitHub Actions.

## Visual evidence

N/A — the user-visible update state reuses the existing package-managed settings interface; this change primarily affects package availability and automation.

## Related issue

Closes #50


spandan11106:
## Why this doesn't replicate Omarchy's channel system, and why there are two packages

The channel model here was taken from Omarchy deliberately, but one half of it can't be reproduced from the AUR. Writing down which half, and why, so this isn't re-opened later.

### What Omarchy actually does

Omarchy's channels are **pacman repositories it owns**. `omarchy-channel-set` is, in full:

```bash
sudo cp -f "$OMARCHY_PATH/default/pacman/pacman-$channel.conf" /etc/pacman.conf
sudo cp -f "$OMARCHY_PATH/default/pacman/mirrorlist-$channel" /etc/pacman.d/mirrorlist
sudo pacman -S --needed --noconfirm --ask 4 "${packages[@]}"
```

| Channel | mirrorlist | `[omarchy]` repo |
| --- | --- | --- |
| stable | `stable-mirror.omarchy.org` | `pkgs.omarchy.org/stable/$arch` |
| rc | `rc-mirror.omarchy.org` | `pkgs.omarchy.org/rc/$arch` |
| edge | `mirror.omarchy.org` | `pkgs.omarchy.org/edge/$arch` |

Stable and rc ship **the same package name**, `omarchy`, from two different servers. That is the entire trick, and it requires operating those servers.

### Why we can't do that

The AUR is an index of build recipes, not a package repository. There is no Strata-controlled server for pacman to point at, and no `[strata]` section to swap, so there is nothing to rewrite in `pacman.conf`. Serving one package name from two channels would mean standing up and signing a real binary repo — hosting, keys, mirrors — which is a much larger commitment than publishing a `PKGBUILD`, and it would put a repo signing key next to the release workflow that already builds the binaries.

Without a repo to swap, **the package name is the only namespace available**, so the channel has to live in the name: `strata-bin` and `strata-rc-bin`.

Worth noting Omarchy hits the same wall itself. Its edge channel ships `omarchy-dev` / `omarchy-settings-dev` — differently *named* packages, with `--ask 4` in the switch command specifically to accept the `omarchy` ⇄ `omarchy-dev` replacement prompt. The name-based approach is Omarchy's own fallback whenever the repo trick isn't enough. We just have to use it for every channel rather than one.

### "Two binaries" is one binary from two tags

The two packages are not two builds and not a fork. Both render from the same `packaging/aur/PKGBUILD.in`, both download an official release archive, and both install it unmodified — `options=('!strip')` keeps the binary byte-identical to the artifact GitHub attested, so `gh attestation verify` still matches the file pacman installed. The only difference between them is which release tag they pin. They `provides=("strata=${pkgver}")` and `conflicts` with each other, so only one is ever installed and switching channels is an ordinary package install, which is the same user-facing shape as `omarchy-channel-set`.

### Why not one package with a channel option

`makepkg` has no user-selectable build option that pacman remembers across upgrades, so a channel choice would not survive `-Syu`. Worse, a single `pkgname` carrying both channels would have to move its `pkgver` backwards whenever a stable release lands after a newer RC, and pacman would stop offering upgrades to everyone already on the RC. Separate names avoid both problems: each package's version only ever moves forward.

### What we did keep

- **The channel is derived, not stored.** `omarchy-channel-current` has no config file; it reads `pacman -Q` and greps the mirror out of `pacman.conf`. Strata now does the equivalent — the packaging marker names the package's channel, the Settings selector is locked to it, and the persisted preference is made to match so the update check only asks for releases the installed package could actually deliver.
- **Defer rather than self-update.** Omarchy's updater is a system-level orchestrator (snapshot → `pacman -Syu` → migrations → `yay -Sua` → restart). Strata is one application and has no business owning a system upgrade, so the equivalent is to detect that a package manager owns `/usr/bin/strata` and hand the user the right command instead of writing over it.

One consequence worth flagging: because `omarchy-update` runs `yay -Sua` for foreign packages, an Omarchy user who installs `strata-bin` from the AUR already gets Strata updated by `omarchy update`, with no Strata-specific integration needed.

### Nightly

Not packaged, for a different and purely mechanical reason: `makepkg` calls `update_pkgver` *after* `download_sources` and `extract_sources`, and never re-downloads. A `pkgver()` that discovers the newest nightly would therefore bump the version after the old tarball had already been fetched, packaging the wrong binary — and GitHub has no "latest prerelease" redirect to point a static `source` at either (`/releases/latest/` resolves to the newest non-prerelease). Pinning each nightly instead would mean an AUR push per nightly. Nightly users install manually and update in-app, which already supports the channel.

spandan11106:
@l0gicgate please check this out whenever you are free. It would be nice to have a AUR package before we move to version 1.0.0

l0gicgate:
@spandan11106 it's on my todo list for tomorrow. Ryan just added us to OPR:
image

Have to do some work with the update mechanism, so if the package is managed by package manager it can't show the updates settings.

l0gicgate:
@cursor please review

PR #201, opened 2026-09-03
perf(hidden): toggle hidden files in-memory without re-enumerating

### LINKED ISSUE #200: perf: hidden files toggle freezes UI
## Summary

Toggling hidden file visibility causes a visible freeze because every open column is re-enumerated from disk and every directory monitor is torn down and recreated, all sequentially on the main thread.

## Reproduction

1. Open a directory with several columns visible
2. Toggle "Show Hidden Files"
3. UI freezes for a noticeable period

## Expected behavior

Toggle should be instant with no disk I/O.

JoeJoeflyn:
## Summary

- Store `is_hidden` on `FileEntry` at enumeration time
- Keep all entries in memory; filter hidden files at the UI level via `CustomFilter`
- `toggle_hidden` flips a boolean and calls `filter.changed()` — no disk re-enumeration, no monitor reinstallation
- Extracted shared `entry_filter()` helper to avoid duplicated filter closures across columns, grid, and explorer panes

## Why

Toggling hidden files re-enumerated every open column from disk AND tore down + reinstalled every directory monitor, all sequential on the main thread. For 5 open columns that's 5 disk reads + 5 monitor recreations — the freeze users see.

The fix keeps all entries (including hidden ones) in memory and lets GTK's virtualized `FilterListModel` hide/show them instantly. The `is_hidden` flag is encoded in the model value (`"dh\t.config"` vs `"dv\tREADME.md"`) so the filter closure can check it in O(1) without a side lookup.

Memory cost: ~132 bytes per hidden entry, bounded by the existing 100k `MAX_DIRECTORY_ENTRIES` cap. Negligible — a single thumbnail uses more RAM.

## Manual verification

1. Open a directory with several columns
2. Toggle "Show Hidden Files" on and off repeatedly
3. Toggle is instant with no freeze

## Expected result

Hidden files appear/disappear instantly on toggle with no UI freeze.

Closes #200

### DIFF
--- a/src/adapters/local_files.rs
+++ b/src/adapters/local_files.rs
@@ -122,6 +122,7 @@ fn entry_from_info(location: Location, info: gio::FileInfo) -> FileEntry {
             .modification_date_time()
             .map(|modified| MetadataValue::Known(modified.to_unix()))
             .unwrap_or(MetadataValue::Unavailable),
+        is_hidden: info_is_hidden(&info),
     }
 }
 
@@ -266,7 +267,6 @@ impl FileSource for LocalFileSource {
                     Ok(Ok(files)) => {
                         let mut entries: Vec<_> = files
                             .into_iter()
-                            .filter(|info| request.include_hidden || !info_is_hidden(info))
                             .filter_map(|info| {
                                 let child = directory.child(info.name());
                                 Some(entry_from_info(location_for_file(&child)?, info))
@@ -327,6 +327,7 @@ impl FileSource for LocalFileSource {
         include_hidden: bool,
         notify: Rc<dyn Fn(DirectoryChange)>,
     ) -> Option<LoadHandle> {
+        let _ = include_hidden;
         let path = location.native_path()?.to_path_buf();
         let file = gio::File::for_path(&path);
         let monitor = match file.monitor_directory(
@@ -405,7 +406,7 @@ impl FileSource for LocalFileSource {
             let cancelled = cancelled_for_change.clone();
             let source = glib::timeout_add_local_once(Duration::from_millis(100), move || {
                 timeout.take();
-                flush_monitor_changes(&pending, include_hidden, &notify, &cancelled);
+                flush_monitor_changes(&pending, &notify, &cancelled);
             });
             timeout_for_change.replace(Some(source));
         });
@@ -474,7 +475,6 @@ fn merge_pending_change(
 
 fn flush_monitor_changes(
     pending: &RefCell<HashMap<PathBuf, PendingMonitorChange>>,
-    include_hidden: bool,
     notify: &Rc<dyn Fn(DirectoryChange)>,
     cancelled: &Rc<Cell<bool>>,
 ) {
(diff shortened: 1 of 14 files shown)


src/ui/browser_modes.rs
```diff
@@ -998,15 +1010,10 @@ fn build_grid_pane(
...
-    let query = filter_query.clone();
-    let filter = gtk::CustomFilter::new(move |item| {
-        let Some(item) = item.downcast_ref::<gtk::StringObject>() else {
-            return false;
-        };
-        let query = query.borrow();
-        query.is_empty() || item.string().to_lowercase().contains(query.as_str())
-    });
+    let show_hidden = Rc::new(Cell::new(false));
+    let filter = super::browser::entry_filter(show_hidden.clone(), filter_query.clone());
```
kallupx, line 1022: Grid and Explorer still put raw display names into this model, but entry_filter() expects the encoded value. That leaves files like .config visible and can treat a normal name like photo as hidden. These paths should use the same encoding as Columns.

src/adapters/local_files.rs
```diff
@@ -266,7 +267,6 @@ impl FileSource for LocalFileSource {
                     Ok(Ok(files)) => {
                         let mut entries: Vec<_> = files
                             .into_iter()
-                            .filter(|info| request.include_hidden || !info_is_hidden(info))
```
kallupx, line 269: Now that hidden entries stay in NavigationState, j/down still walks the raw list. It can select an invisible dotfile, then Enter or Space acts on it. Keyboard movement should skip hidden entries when the setting is off.

src/ui/browser.rs
```diff
@@ -4067,17 +4034,8 @@ impl ViewState {
...
-            let Some(item) = item.downcast_ref::<gtk::StringObject>() else {
-                return false;
-            };
-            let query = query.borrow();
-            query.is_empty()
-                || model_display_name(&item.string())
-                    .to_lowercase()
-                    .contains(query.as_str())
-        });
+        let show_hidden = Rc::new(Cell::new(false));
```
kallupx: This always resets the filter to false, so a saved “show hidden” setting/a new column opened after enabling it doesn’t take effect. Could this start from the current column preference? Grid and Explorer have the same reset.

src/app/browser.rs
```diff
@@ -479,7 +482,6 @@ impl Browser {
                 id: request_id,
                 location,
                 batch_size: 128,
-                include_hidden: self.preferences.get().show_hidden,
```
kallupx, line 482: Peek lost the source-side filter here, but its model never got a UI hidden filter. With the default setting, peeking a folder can now show dotfiles. Lets filter peek batches before applying the item limit.

kallupx (changes requested):
Nice performance improvement overall. I traced the hidden-file flow end to end and reproduced the Grid/Explorer model mismatch, invisible keyboard selection, and unfiltered Peek results. I also confirmed that new panes don’t inherit the current preference. So I’m requesting a few changes, see the inline comments.

Once fixed and covered by regression tests, this should be good to go.

JoeJoeflyn:
@kallupx thanks for feedback, can you take another look

kallupx (approved):
All comments have been addessed, verified manually theres no noticeable freezing.

PR #202, opened 2026-09-03
feat(sidebar): allow pinned folders to be reordered

### LINKED ISSUE #184: Allow pinned folders to be reordered in the sidebar
## Problem or use case

Pinned folders are displayed in the order they were added, but users cannot reorganize them. This makes it difficult to keep frequently used folders in a preferred order as the pinned list grows.

## Proposed solution

Allow users to drag and drop folders within the **Pinned** section to reorder them, matching the existing reordering behavior for standard sidebar folders such as Documents, Downloads, Pictures, and Videos.

The custom order should be saved and restored across application restarts. Reordering should remain within the Pinned section and should not affect the standard folders.

## Alternatives considered

Users can unpin folders and pin them again in the desired order, but this is cumbersome and becomes increasingly inconvenient with longer lists.

## Visual examples

N/A — use the existing standard-folder drag-and-drop interaction for consistency.

## Additional context

Pinned folder order is already persisted in the GTK bookmarks data, so the feature should preserve that ordering when rows are rearranged.

spandan11106:
## Description

Pinned folders could only be arranged by unpinning and re-pinning them. They are now reorderable by drag and drop, using the same interaction, cursor, and drop styling as the standard folders (Desktop, Documents, Downloads, …).

Reordering stays inside the Pinned section: the drag payload is namespaced with a `pinned:` prefix, so a standard row rejects a pinned payload and a pinned row rejects a standard one. Rows carry their index into the stored bookmark list, so bookmarks hidden from the Pinned section (those that resolve to standard folders) keep their place in the file. The new order is written back to the GTK bookmarks file, so it is restored on the next launch.

## Visual evidence

N/A — no new elements or styling. The rows reuse the existing `.sidebar-row.reorderable` drag/drop appearance, so before/after screenshots are identical outside of the drag itself.

## How to test

1. Pin at least three folders (right-click a folder in the file list → Pin).
2. Drag a pinned row above and below other pinned rows; drop on the upper half of a row to place it before, the lower half to place it after.
3. Try dragging a pinned row onto a standard folder row, and a standard folder row onto a pinned row.
4. Restart Strata.

**Expected result:** Pinned rows reorder to the drop position and the Pinned section keeps its own order; cross-section drags do nothing; the order from step 2 is still in place after the restart.

## Related issue

Closes #184

### DIFF
--- a/src/ui/theme.rs
+++ b/src/ui/theme.rs
@@ -80,6 +80,8 @@ struct Preferences {
     browser_mode: String,
     #[serde(default = "default_browser_density")]
     browser_density: String,
+    #[serde(default = "default_sidebar_order")]
+    sidebar_order: Vec<String>,
     #[serde(default)]
     show_hidden: bool,
     #[serde(default = "default_enabled")]
@@ -109,6 +111,7 @@ impl Default for Preferences {
             reduce_motion: false,
             browser_mode: default_browser_mode(),
             browser_density: default_browser_density(),
+            sidebar_order: default_sidebar_order(),
             show_hidden: false,
             folders_first: true,
             sort_key: default_sort_key(),
@@ -138,6 +141,16 @@ fn default_video_preview_backend() -> String {
 
 fn default_browser_density() -> String {
     "compact".to_owned()
+}
+
+fn default_sidebar_order() -> Vec<String> {
+    vec![
+        "desktop".to_owned(),
+        "documents".to_owned(),
+        "downloads".to_owned(),
+        "pictures".to_owned(),
+        "videos".to_owned(),
+    ]
 }
 
 fn default_sort_key() -> String {
@@ -347,6 +360,15 @@ impl ThemeManager {
             super::browser_modes::BrowserDensity::Airy => "airy",
         }
         .to_owned();
+        self.save_preferences();
+    }
+
+    pub fn sidebar_order(&self) -> Vec<String> {
+        self.preferences.borrow().sidebar_order.clone()
+    }
+
+    pub fn set_sidebar_order(&self, order: Vec<String>) {
+        self.preferences.borrow_mut().sidebar_order = order;
         self.save_preferences();
     }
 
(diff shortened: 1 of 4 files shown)


l0gicgate:
@spandan11106 please include video recordings for features like this. Makes it easier to preview the changes.

Thank you for this contribution 🙏🏼

l0gicgate:
Reviewed the branch and applied two follow-up improvements to the sidebar reorder work:

- **Shared drag/drop wiring** — extracted the drag-source and drop-target setup from `append_reorderable_place` and `make_pinned_row_reorderable` into a single `make_reorderable` helper. Both row types now reuse one implementation (cursor swap, `dragging` class, drop skeleton), passing only their own payload and drop handler. Behavior is unchanged: pinned rows keep the `pinned:<index>` payload namespace, and the standard drop handler still rejects those payloads.

- **Documented the index remapping** in `reorder_pinned_places`, which is easy to misread because `source` is a pre-removal index while `target`/`destination` are computed against the shrunken vector.

Checks pass: `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --all-targets --all-features` (391 passed, 0 failed).

Refs #184

l0gicgate:
Also implemented persistence for the standard sidebar place order (pinned-places order already persisted to the GTK bookmarks file, but the standard rows were resetting to defaults on every launch).

- The reorderable standard rows now save their arrangement back to the `sidebar_order` preference in `~/.config/strata/settings.toml` whenever the order changes.
- On startup the sidebar restores that order, dropping any unknown IDs and appending any standard place missing from the stored list so the default set stays complete.
- The `ThemeManager` gained `sidebar_order()` / `set_sidebar_order()` accessors, following the existing preference read/write pattern.

Checks pass: `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --all-targets --all-features` (399 passed, 0 failed).

Refs #184

PR #203, opened 2026-09-03
feat(selection): start marquee drags from chrome beside file panes

### LINKED ISSUE #189: feat(selection): allow marquee drags to start outside file panes
### Problem or use case

Marquee selection currently has to begin on empty space inside a file pane. In List/Explorer mode, a directory with enough rows to fill and scroll the viewport leaves little or no empty space, making mouse-based selection of multiple files impractical without modifier-clicking or using the keyboard.

### Proposed solution

Explore whether a marquee drag can begin in adjacent empty space outside a file pane and continue into the pane to select files.

The implementation should determine:

- which surrounding regions can safely initiate selection without conflicting with navigation, column resizing, scrollbars, or other controls;
- whether GTK gesture capture can maintain the drag as it crosses into the file list;
- how coordinates and hit testing should be translated into the target pane;
- whether dragging near the pane edges should auto-scroll through a long list; and
- whether the behavior can work consistently across List/Explorer, Grid, and Columns modes.

At minimum, users should have a practical mouse-only way to begin marquee selection when rows fill the entire List/Explorer viewport. Existing Ctrl+Click, Shift+Click, keyboard selection, row dragging, and empty-space marquee behavior must continue to work.

### Alternatives considered

- Require a modifier while beginning a drag on a row. This can work technically but is less discoverable and still conflicts with dragging files.
- Rely on Ctrl+Click, Shift+Click, or keyboard range selection. These remain useful but do not replace direct marquee selection.
- Add permanent blank space inside the viewport. This wastes limited space and does not help users begin selection near a specific visible range.

### Visual examples

N/A — include a short recording of selection beginning outside a full, scrollable list if the interaction proves feasible.

### Strata version

Current development build.

### Additional context

This is partly an interaction and GTK event-routing exploration. If cross-pane marquee initiation is not reliable, document the constraint and implement the closest practical mouse-first alternative.

spandan11106:
## Description

Marquee selection had to begin on empty space inside a file pane. A directory long
enough to fill and scroll the viewport leaves no such space, so mouse-only
multi-selection was impractical in Explorer.

The two near-identical marquee implementations (columns view in `ui::browser`,
grid/explorer panes in `ui::browser_modes`) now live in a shared `ui::marquee`
module, which also supports starting a drag on inert chrome outside a pane and
carrying it into the pane:

- **Sidebar** — blank panel space, in every mode. This is the most reachable
  origin, since the sidebar sits beside all three views. It targets the pane
  nearest the start edge: the explorer or grid pane, or the first open column.
- **Explorer** — the pane header and the column-heading strip.
- **Grid** — the pane header.
- **Columns** — each column header, plus the blank strip beside the last open
  column; a drag there targets the last column.

A press landing on a button, entry, slider, or scrollbar is left to that control,
so navigation, sorting, column resizing, scrollbars, and the sidebar's own place
rows (which are buttons) keep their existing behaviour. GTK keeps the event
sequence on the originating widget once the gesture is claimed, so the drag
continues into the file list without further routing.

Two supporting changes make cross-pane drags behave:

- The anchor is stored in the view's own coordinates and the pointer in the
  scrolled window's, so the band stays glued to the content while the list moves.
- Holding the pointer near a viewport edge auto-scrolls the list, with the band
  and selection recomputed each frame. Axes the view cannot scroll contribute
  nothing, so the timer does not spin against a fixed edge.

The band is also clipped to its overlay instead of only being clamped at the
origin, so a drag beginning above or beside the view no longer paints outside it.

### Explorer now activates on double click

Please review this part specifically — it is a deliberate behaviour change.

Explorer set `single_click_activate(true)`. GTK bundles two behaviours into that
property: "activate rows on single click **and select them on hover**". The hover
half replaced the whole selection with whatever row the pointer crossed, which
discarded a marquee the instant the drag ended.

This was not caused by this branch and was not limited to the marquee — on `main`,
hover-select already made Ctrl+Click and Shift+Click multi-selection unusable in
Explorer, since moving the pointer collapsed the selection to one row. GTK offers
no way to keep single-click activation without hover selection, so Explorer now
activates on a double click, matching the grid and column views which already do.

## Visual evidence

N/A — the change is an interaction rather than a new visual element, and I have no
screen capture tooling on this machine. Happy to add a recording before merge.

## How to test

1. Open a directory with enough entries to fill and scroll the viewport, in Explorer.
2. Press blank sidebar space (the area below the last place) and drag right into the
   rows. Repeat from the pane header and from the column-heading strip.
3. Keep dragging past the bottom edge of the list.
4. Repeat step 2 in Grid, and in List from the blank strip right of the last column.
5. Ctrl+Click and Shift+Click several rows in Explorer, then move the pointer across
   the list.
6. Confirm the untouched paths: click a heading to sort, drag a column resize handle,
   drag the scrollbar, drag a row to move a file, click a sidebar place to navigate,
   drag a sidebar place to reorder it, and double-click a folder in Explorer to open it.

**Expected result:** a rubber band appears from the press point and selects every row
it crosses; step 3 scrolls the list and keeps extending the selection; the step 5
selection survives pointer movement; everything in step 6 behaves as before, except
that Explorer now opens on a double click.

## Related issue

Closes #189

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1283,6 +1283,15 @@ switch:checked slider {
   border-right: 1px solid alpha(@theme_border, 0.6);
 }
 
+.sidebar-scroll scrollbar.vertical {
+  min-width: 10px;
+}
+
+.sidebar-scroll scrollbar.vertical slider {
+  margin: 3px;
+  min-width: 4px;
+}
+
 .sidebar {
   background: @theme_surface;
   padding: 8px 7px;
--- a/src/style.css
+++ b/src/style.css
@@ -1283,21 +1283,35 @@ switch:checked slider {
   border-right: 1px solid alpha(@theme_border, 0.6);
 }
 
-.sidebar-scroll scrollbar.vertical,
-.sidebar-scroll scrollbar.vertical:hover,
-.sidebar-scroll scrollbar.vertical.hovering,
-.sidebar-scroll scrollbar.vertical.dragging {
+.fixed-scrollbar scrollbar,
+.fixed-scrollbar scrollbar:hover,
+.fixed-scrollbar scrollbar.hovering,
+.fixed-scrollbar scrollbar.dragging {
   background: transparent;
   border: none;
-  min-width: 5px;
   padding: 0;
   transition: none;
 }
 
-.sidebar-scroll scrollbar.vertical > range > trough > slider {
+.fixed-scrollbar scrollbar.vertical {
+  min-width: 5px;
+}
+
+.fixed-scrollbar scrollbar.horizontal {
+  min-height: 5px;
+}
+
+.fixed-scrollbar scrollbar > range > trough > slider {
   border-width: 1px;
   margin: 0;
+}
+
+.fixed-scrollbar scrollbar.vertical > range > trough > slider {
   min-width: 3px;
+}
+
+.fixed-scrollbar scrollbar.horizontal > range > trough > slider {
+  min-height: 3px;
 }
 
 .sidebar {
--- a/src/style.css
+++ b/src/style.css
@@ -1283,13 +1283,21 @@ switch:checked slider {
   border-right: 1px solid alpha(@theme_border, 0.6);
 }
 
-.sidebar-scroll scrollbar.vertical {
-  min-width: 10px;
+.sidebar-scroll scrollbar.vertical,
+.sidebar-scroll scrollbar.vertical:hover,
+.sidebar-scroll scrollbar.vertical.hovering,
+.sidebar-scroll scrollbar.vertical.dragging {
+  background: transparent;
+  border: none;
+  min-width: 5px;
+  padding: 0;
+  transition: none;
 }
 
-.sidebar-scroll scrollbar.vertical slider {
-  margin: 3px;
-  min-width: 4px;
+.sidebar-scroll scrollbar.vertical > range > trough > slider {
+  border-width: 1px;
+  margin: 0;
+  min-width: 3px;
 }
 
 .sidebar {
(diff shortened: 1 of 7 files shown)


spandan11106:
https://github.com/user-attachments/assets/b02da0b6-bcd1-406a-9d98-41ddc2387228

l0gicgate:
This is awesome @spandan11106 let's goooo! Thank you for this. I will try to get this in shortly

l0gicgate:
The single-click behavior was fixed as part of #220.

Explorer had used GTK's `single_click_activate(true)`. GTK bundles two behaviors into that property: activating rows on single click and selecting them on hover. The hover behavior replaced the whole selection with whichever row the pointer crossed, discarding a marquee as soon as the drag ended. This was not caused by this branch or limited to marquee selection: on `main`, hover selection also made Ctrl+Click and Shift+Click multi-selection unusable in Explorer.

#220 now keeps GTK's native single-click activation disabled and handles the configured click behavior with per-row gestures instead. That preserves marquee and modifier-based multi-selection while still supporting single-click activation when configured (with double click as Explorer's default, matching Grid).

l0gicgate:
Added a follow-up scrollbar polish pass:

- Kept GTK overlay scrolling enabled, so scrollbars still auto-hide normally.
- Prevented the scrollbar track and thumb from expanding on hover or while dragging.
- Applied the same fixed narrow treatment to the sidebar, every Miller column pane, Grid, and Explorer.
- Applied it to horizontal overflow scrollbars in Miller, Grid, and Explorer as well, for consistent behavior in both orientations.

The styling is shared through the `fixed-scrollbar` CSS class rather than duplicated per view.

PR #221, opened 2026-09-03
feat(preview): add SQLite database preview with table browsing

JoeJoeflyn:
## Description

- Add SQLite database preview with table browsing and schema viewer
- Modernize the preview with a clean data grid and native popover table picker
- Replace the `gtk::DropDown` with the app's standard `MenuButton`+`Popover`+`menu_option` pattern (same as the sort menu and video backend selector), including a `form-control` search entry that filters the scrollable table list
- Switch to GTK native column separators (`set_show_column_separators`) so header and body dividers align at the actual column boundary instead of the label text edge
- Prune dead and redundant CSS: `.preview-database-split` (zero references), duplicate pager label rules, and overridden background reset selectors
- Add `connect_map` to re-sync check marks when the table popover reopens

## Visual evidence

image

## How to test

1. Open a `.db` or `.sqlite` file in Strata
2. Click the table name button in the preview header
3. Verify the popover opens with a search field and scrollable table list
4. Type in the search field to filter tables
5. Click a table to switch — check mark updates and data grid reloads
6. Reopen the popover — check mark should be on the currently active table
7. Verify column dividers in the data grid align between header and body

**Expected result:** Table picker uses the same dropdown style as the rest of the app, search filters tables, column dividers are aligned, check marks sync on reopen.

## Related issue

N/A — new feature, no existing issue.


src/sandbox_helper.rs
```diff
@@ -600,3 +601,150 @@ fn bounded_output(command: &mut Command, max_bytes: u64) -> io::Result<Output> {
...
+
+const MAX_DATABASE_ROWS: i32 = 50;
+
+#[derive(Clone, Debug)]
+struct DbTableItem {
+    name: String,
+    is_view: bool,
+}
+
+fn render_database(path: &Path, encoded: i32) -> Result<Vec<u8>, String> {
```
kallupx, line 614: The sandbox only mounts the main database file, leaving its WAL behind. My test returned 2 committed rows normally but only 1 inside the sandbox. We should preserve the WAL state (https://www.sqlite.org/wal.html#the_wal_file) here so the preview doesn’t silently show outdated data.

src/ui/preview.rs
```diff
@@ -1454,5 +1987,150 @@ fn fmt_time(microseconds: i64) -> String {
...
+            match ch {
+                '"' => in_quotes = true,
+                ',' => {
+                    current_row.push(std::mem::take(&mut current_cell));
+                }
+                '\r' => {
+                    if chars.peek() == Some(&'\n') {
+                        chars.next();
+                    }
+                    current_row.push(std::mem::take(&mut current_cell));
```
kallupx: Could we preserve empty rows and keep empty text distinct from NULL? With three single-column rows containing NULL, '', and visible, only visible survives parsing. Empty strings in wider tables also get displayed as NULL.

src/sandbox_helper.rs
```diff
@@ -600,3 +601,150 @@ fn bounded_output(command: &mut Command, max_bytes: u64) -> io::Result<Output> {
...
+        path,
+        &format!("SELECT sql FROM sqlite_master WHERE name={quoted};"),
+    )
+    .unwrap_or_default();
+    let types = sqlite_scalar(
+        path,
+        &format!("SELECT name || char(9) || \"type\" FROM pragma_table_info({quoted});"),
+    )
+    .unwrap_or_default();
+    let count_str =
```
kallupx: Could we count the rows once when opening a table and reuse that total while paging? Every page currently runs COUNT(*) again, which can be expensive for views. Reusing it would keep the current row totals and page numbers in the UI while making page switching faster. We could refresh the count when the table is reopened or refreshed.

src/sandbox_helper.rs
```diff
@@ -600,3 +601,150 @@ fn bounded_output(command: &mut Command, max_bytes: u64) -> io::Result<Output> {
...
+
+#[derive(Clone, Debug)]
+struct DbTableItem {
+    name: String,
+    is_view: bool,
+}
+
+fn render_database(path: &Path, encoded: i32) -> Result<Vec<u8>, String> {
+    let tables = list_tables(path)?;
+    if tables.is_empty() {
```
kallupx, line 616: This returns  zero bytes, but sandbox::parse rejects that, so a valid database with no tables shows “Preview unavailable” instead of the empty-database message.

kallupx (changes requested):
I think we should address the WAL handling and misleading data/error states before merging. It would also be good to avoid repeating the full count when paging. Left a few comments.

src/sandbox.rs
```diff
@@ -428,6 +433,16 @@ fn sandbox_command(
         command.arg("--ro-bind").arg(executable).arg("/app/strata");
     }
     command.arg("--ro-bind").arg(input).arg(&sandbox_input);
+    if operation == ParseOperation::PreviewDatabase {
+        for suffix in ["-wal", "-shm", "-journal"] {
+            let mut companion = input.as_os_str().to_os_string();
+            companion.push(suffix);
+            command
+                .arg("--ro-bind-try")
```
kallupx: Could we open these companion files without following symlinks, verify they’re regular files, and mount them using --ro-bind-fd? Mounting unchecked paths can expose unrelated host files to the helper. Using validated descriptors also protects against the path being replaced before mounting.

kallupx (changes requested):
Added one security related comment

PR #225, opened 2026-09-03
fix(delete): offer permanent delete when a location doesn't support Trash

### LINKED ISSUE #179: Cannot delete on SMB/GVfs mounts: 'Move to Trash' fails, no 'Delete permanently' option
**Strata version:** 0.7.0 (GitHub release archive, x86_64-unknown-linux-gnu; source commit `f2594d106d8fdb878b786b445c61379d391e2c0f`)

**Installation method:** GitHub release archive

**Linux distribution and desktop environment:** Omarchy (Arch-derived), Hyprland on Wayland, glibc 2.44, GTK 4.22

## What happened?

On an SMB share mounted through GVfs (`smb://…`, backend `gvfsd-smb`), the file context menu offers only **Move to Trash** (Delete). Because GVfs SMB mounts do not implement GIO's Trash protocol, deleting fails with **"This location doesn't support Trash. Delete permanently instead."** There is no **Delete permanently** entry in the context menu, so files on SMB mounts cannot be deleted from the UI at all. A batch delete additionally produces a "Completed with errors" dialog whose only button is **Done** — there is no actionable way to complete the permanent deletion.

Worth noting: the release binary *does* contain a **Permanently delete** context-menu action (bound to Delete, marked `danger`) and calls `g_file_delete_async` for permanent deletion — but it is not surfaced on trash-less mounts.

## Steps to reproduce

1. In Strata, press **Ctrl+L**, enter `smb://<server>/<share>`, press Enter to open an SMB mount.
2. Right-click a file or folder on the share.
3. Observe the context menu lists only **Move to Trash** (Delete) — there is no **Delete permanently** option.
4. Press Delete (or trigger a batch delete).
5. The error appears: "This location doesn't support Trash. Delete permanently instead." (batch: "Completed with errors … Delete permanently instead." with only a **Done** button).
6. The item remains; there is no way to delete it from Strata.

## Expected behavior

Strata should offer a **Delete permanently** option (with a destructive-action confirmation) on filesystems that don't support Trash — for example, swap *Move to Trash* → *Delete permanently* via `g_file_delete_async` on such mounts, or include it as a distinct context-menu item. The "this location doesn't support Trash" dialog should also provide an actionable **Delete permanently** confirmation instead of only **Done**.

## Screenshots

- Context menu on an SMB mount showing only *Move to Trash* (no permanent-delete option).
- "Completed with errors" dialog: *1 item could not be deleted … This location doesn't support Trash. Delete permanently instead.* with only a *Done* button.

## Relevant logs

```text
WARN strata::ui::browser: unable to open file backend=native error_domain=g-io-error-quark error_code=15
```

## Additional context

- Reproducible on any GVfs SMB mount (tested against a Samba/Ubuntu server). Other non-local GVfs mounts that lack Trash are presumably affected too.
- Local filesystem trashing works fine — only remote/trash-less mounts are affected.
- Workaround (not ideal): delete from a terminal on the gvfs mount path, e.g. `rm -r "/run/user/1000/gvfs/smb-share:server=…,share=…/folder"`.
- The README's feature list includes a "permanent delete" file operation, so this appears to be a gap between the shipped capability and what's surfaced on trash-less mounts.

pdwaldrop:
## Description

On SMB/GVfs mounts (and other locations that don't support Trash), the only delete action available was "Move to Trash", which fails with "This location doesn't support Trash. Delete permanently instead." The resulting "Completed with errors" dialog only had a "Done" button — no way to actually delete the item short of dismissing the error, re-selecting the item, and manually choosing the separate "Permanently delete" context-menu action.

This tracks which failed entries were specifically trash-unsupported failures (as opposed to some other, unrelated error) through `OperationEvent::CompletedWithErrors` and `BrowserEvent::OperationCompletedWithErrors`, as a list of retryable locations. When that list is non-empty, the completion dialog now also offers a "Delete Permanently" button; choosing it reuses the existing permanent-delete confirmation dialog, scoped to just the entries that actually failed for that reason — not ones that already succeeded, or failed for some unrelated reason (e.g. a permission error).

## Visual evidence

Screenshots attached below showing the full flow against a real trash-unsupported mount:
1. The "Completed with errors" dialog now offers "Delete Permanently" alongside "Done", instead of being a dead end.
2. Clicking it chains into the existing permanent-delete confirmation dialog, scoped to just the retryable item.
3. Confirming actually deletes the file — Strata shows "This directory is empty" and the filesystem agrees.

## How to test

1. Find or create a location GIO refuses to trash into — e.g. `/tmp` itself reliably reproduces this (`gio trash <file-in-tmp>` fails with "Trashing on system internal mounts is not supported"), or use a real SMB/GVfs mount as in the original report.
2. In Strata, select a file there and press Delete (or trigger a batch delete).
3. Observe the "Completed with errors" dialog now includes a "Delete Permanently" button.
4. Click it, confirm the follow-up "Permanently delete N?" dialog.

**Expected result:** the item is actually deleted, with no dead-end error state.

I verified this exact flow end to end (not just via unit tests) against `/tmp` as the trash-unsupported destination.

## Related issue

Closes #179


pdwaldrop:
image

image

image

l0gicgate:
Review/testing update:

- Merged the latest `main` locally and resolved conflicts in the operations and browser UI tests/code.
- Adapted the retry flow to `main`'s new immediate Move to Trash behavior by snapshotting the requested entries before dispatching the trash operation.
- Refined the UX after manual testing: when every failure is caused by Trash being unsupported, Strata now transitions directly to the existing permanent-delete confirmation instead of first showing the Completed with errors dialog.
- Mixed failures still show Completed with errors and offer the permanent-delete action for only the Trash-unsupported entries, so unrelated errors are not hidden.
- Confirmed `/tmp` reports `access::can-trash: FALSE` and reproduces the unsupported-trash path.

The full local suite passes (formatting, Clippy with warnings denied, and all 502 tests). These follow-up changes are currently local and still need to be pushed to the PR branch.

PR #235, opened 2026-09-03
feat(browser): group Explorer and Grid entries by file type

### LINKED ISSUE #192: feat(browser): group files by type in Explorer and Grid modes
### Problem or use case

Directories with many different kinds of files are harder to scan when every item appears in one continuous list. Grouping items into labeled sections by file type would make related files easier to find at a glance.

### Proposed solution

Add an optional **Group by file type** view setting for Explorer and Grid modes.

When enabled:
- Place directories in a **Folder** group.
- Group files under user-facing type labels such as **JSON document**, **Markdown document**, **Plain text document**, or **Python script**.
- Preserve the selected sort order within each group.
- Show a clear heading for each group; Grid mode should use the same grouping with grid items beneath each heading.
- Keep the spatial Miller-column mode unchanged.

### Alternatives considered

Sorting by type puts similar files next to one another, but it does not provide labeled visual sections and makes boundaries between types less obvious.

### Visual examples

Follow the familiar file-manager pattern where the view is divided by type headings—for example, **Folder**, **JSON document**, **Markdown document**, **Plain text document**, and **Python script**—with matching items listed or tiled below each heading.

### Additional context

This should be scoped to Explorer and Grid modes only, not the Miller-column mode.

spandan11106:
## Description

Adds an optional **Group by file type** view setting for Explorer and Grid modes. When it is on,
directories go into a **Folder** group and files sit under labels from the shared MIME database
("JSON document", "Markdown document", "Python script", …), so they read the way they do elsewhere
on the desktop. Folders lead, the remaining groups are alphabetical, and the pane's own sort order
is preserved inside each group. The Miller-column mode is untouched, and the setting is stored with
the other view preferences.

Explorer grouping uses list-view sections, so it keeps one selection, keyboard navigation, and
marquee drags. Grid views have no section headings, so a grouped grid renders one heading and grid
per type: a pane can now hold several sections, selection is reported to the browser as the union
across them, one marquee drag runs through all of them, and a plain click clears the groups it did
not land in.

Two consequences worth flagging for a grouped grid: arrow-key movement and shift-click stay inside
the group they start in, and each group pins its column count to what the pane width allows,
because a grid only knows its row count once its columns are fixed.

Six commits, each self-contained and green on its own: the section and marquee refactors land
first with no behaviour change, then the setting, then Explorer grouping, then Grid grouping.

## Visual evidence

**Explorer**

Explorer grouped by file type

**Grid**

Grid grouped by file type

## How to test

1. Create a directory with several file types, e.g. `mkdir -p demo/{Projects,Notes}` and
   `touch demo/{config.json,package.json,README.md,notes.md,plain.txt,script.py,style.css}`.
2. Open it in Strata and switch to Explorer mode from the appearance menu.
3. Enable **Group by file type** in the same menu.
4. Switch to Grid mode; scroll, then select items with click, ctrl+click, and a marquee drag across
   two groups.
5. Sort by another column, type in the pane filter, and toggle hidden files.
6. Turn the setting off again, and switch to the list (Miller-column) mode.

**Expected result:** Explorer and Grid divide entries into labeled type sections with folders
first; sorting inside a group follows the column header or sort menu; filtering and hidden-file
toggles drop a group once it is empty; selection survives across groups; the setting is remembered
across restarts; and the ungrouped views and the Miller-column mode look and behave exactly as
before.

## Related issue

Closes #192


l0gicgate:
Updated the appearance-menu presentation based on review feedback:

- Moved **Group by file type** out of the mutually exclusive **View** choices and placed it beside **Hidden files**.
- Switched it to the distinct Lucide `list-checks` icon.
- Disabled the option in List/Miller mode, where grouping does not apply; the saved preference is preserved and becomes available again in Grid or Explorer.

Updated appearance menu

l0gicgate:
Manually verified the grouped context-menu selection fix:

1. Enabled **Group by file type** in Grid view.
2. Selected items across two different type groups with Ctrl-click.
3. Right-clicked an already selected item and confirmed the cross-group multi-selection was preserved.
4. Right-clicked an unselected item and confirmed all prior selections were cleared, only the clicked item remained selected, and the context menu targeted that item alone.
5. Repeated the context-selection check in Explorer and confirmed its existing behavior was unchanged.

All expected behavior passed.

PR #252, opened 2026-09-04
feat(appearance): unify menu typography and add app-wide text sizing

### LINKED ISSUE #155: feat(appearance): unify menu typography and add app-wide text sizing
### Problem or use case

Font sizes are inconsistent across Strata's interface. In particular, the Sort menu and the main Appearance menu in the top bar use text that is disproportionately large compared with the surrounding toolbar, file list, and other UI elements. This makes the interface feel visually inconsistent.

Strata also does not currently provide an app-wide text-size control, so users cannot adjust typography for readability or display density.

### Proposed solution

- Standardize menu typography so the Sort menu, top-bar Appearance menu, and similar menus use font sizes consistent with the rest of the UI.
- Audit related menus for the same inconsistency rather than fixing only the two known examples.
- Add an app-wide text-size setting under **Theme and Appearance**.
- Apply the selected size consistently across the app while preserving the intended hierarchy between headings, labels, menu items, and secondary text.
- Persist the setting and apply changes live.

### Alternatives considered

Relying on desktop-wide font or display scaling affects every application and does not address inconsistent relative sizes within Strata.

### Visual examples

Image

Image

### Additional context

The app-wide control should scale from a shared typography system rather than independently overriding individual widgets, so menus remain proportional to the rest of the UI.

pdwaldrop:
## Description

Two parts, both from #155:

**Menu typography audit.** The per-item right-click menu (`.item-context-menu`) already scales its text to `0.88em`, but three sibling popovers never got the same treatment and rendered at the unscaled root size: the Sort By popover (`.column-menu`, also reused by the Settings video-backend dropdown), the top-bar Appearance popover (`.appearance-menu`), and the folder-background context menu (`.folder-context-menu`, also reused for the Trash/Network/pinned-place sidebar menus). Brought all three to the same `0.88em` already established by `.item-context-menu`, rather than inventing a new value.

**App-wide text size.** Added a Text size control (Small/Medium/Large) under Theme & appearance. It's backed by a new persisted `text_size` preference, and applies by setting the root `window` font-size through `ThemeManager`'s existing higher-priority CSS provider (the same mechanism that already injects `@define-color` theme tokens on every theme change). Since virtually everything in `style.css` is sized in `em` relative to that root, one root font-size change scales the whole app proportionally — no per-widget overrides needed, and existing size hierarchy (headings vs. labels vs. secondary text) is preserved. Applies live, no restart required.

## Visual evidence

New Text size control:

typography settings

Home directory at each size (Small / Medium / Large), verified via a fully isolated dev build (separate D-Bus session, so it couldn't touch the running installed app) with each size persisted to config and the app relaunched fresh — confirms the setting is read correctly at startup, not just applied live:

Screenshots attached in a follow-up comment.

## How to test

1. Open Settings → Theme & appearance. A new TYPOGRAPHY section appears between the System card and the theme grid, with a Text size row (Small/Medium/Large), Medium selected by default.
2. Switch between the three — the whole app's text (sidebar, file list, breadcrumbs, settings itself) should resize immediately, no restart needed.
3. Quit and relaunch Strata — the chosen size should still be in effect.
4. Open the top-bar Appearance menu (list icon) and the Sort By menu (column header) — their text should now read at the same smaller, denser size as the per-item right-click menu (right-click a file), instead of looking larger than the rest of the UI.
5. Right-click empty folder space, and right-click Trash/Network/a pinned place in the sidebar — those context menus should also read at the smaller size now.

**Expected result:** text size setting persists and applies live app-wide; the four affected menus (Sort By, Appearance, folder-background context menu, sidebar context menus) all read at the same size as the already-correct per-item context menu.

## Related issue

Closes #155

image

image

image

image

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -100,6 +100,7 @@ headerbar menubutton.header-action > button:focus-visible {
 }
 
 .appearance-menu {
+  font-size: 0.88em;
   min-width: 220px;
   padding: 9px 8px;
 }
@@ -1887,6 +1888,7 @@ menubutton.column-header-action > button:active {
 }
 
 .column-menu {
+  font-size: 0.88em;
   min-width: 170px;
   padding: 8px;
 }
@@ -2221,6 +2223,7 @@ menubutton.column-header-action > button:active {
 }
 
 .folder-context-menu {
+  font-size: 0.88em;
   min-width: 230px;
   padding: 7px;
 }
--- a/src/ui/settings.rs
+++ b/src/ui/settings.rs
@@ -27,7 +27,7 @@ use super::{
     browser::{BrowserView, dismiss_modal_layer, modal_layer},
     browser_modes::{BrowserMode, ClickActivation, ClickCount},
     controls::{form_entry, menu_option, modal_layout, segmented_control},
-    theme::{Theme, ThemeManager, ThemeTokens},
+    theme::{TextSize, Theme, ThemeManager, ThemeTokens},
 };
 
 type ThemeCards = Rc<RefCell<Vec<(String, gtk::Button, gtk::Image)>>>;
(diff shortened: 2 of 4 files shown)


l0gicgate:
The original 12px **Small** setting showed glyph clipping in sidebar and folder labels (most visibly around “Trash”):

Small text-size glyph clipping

Updated Small to 11px. The options now use consistent 2px increments: **11px / 13px / 15px**.

l0gicgate (approved):
Reviewed after merging main and verifying the 11px, 13px, and 15px text-size scale. The implementation persists the selection, defaults legacy settings to Medium, and applies updates live.

PR #253, opened 2026-09-04
fix(operations): harden local permanent delete against symlink races

pdwaldrop:
## Description

#8 covers collision checks, traversal, deletion, and transfer all being separate path-based GIO calls — between any two of them, another process in a shared writable directory can swap a path component and redirect the operation, and recursion trusted the `FileEntry` type recorded when an item was originally listed rather than re-checking it.

This covers the **permanent-delete** half specifically, called out in the issue as the most important case. For local paths, deletion now walks descriptor-relative to each already-open directory file descriptor instead of re-resolving path strings:

- Opens each directory once via `openat2` with `RESOLVE_BENEATH | RESOLVE_NO_SYMLINKS | RESOLVE_NO_MAGICLINKS` (the exact flags the issue names), relative to the parent's fd — never by re-walking a path from `/`.
- Re-reads each entry's real type with `statat(..., SYMLINK_NOFOLLOW)` immediately before deciding whether to recurse or unlink, instead of trusting the `FileEntry.kind` the item was listed with. A directory that's since become a symlink is deleted as the symlink it now is, not opened as a directory.
- If a component changes after that check but before the corresponding open (the one window that's fundamentally irreducible), the `RESOLVE_NO_SYMLINKS` open fails closed with a clear error rather than following it.
- Mirrors the `openat`/`O_NOFOLLOW` walk the archive extractor (`ExtractionDestination`) already uses for the same reason, rather than introducing a second technique.
- Remote (GVfs) locations have no local descriptor to walk against, so they keep the existing path-based `permanently_delete` — explicit, not silently claiming an equivalent guarantee.
- Mid-delete cancellation responsiveness is preserved: each child recurses through its own async hop (checking `Cancellable::is_cancelled()` before each), matching the granularity the old GIO-async version had. The existing `cancelling_recursive_delete_leaves_the_unfinished_root_in_place` test (unchanged) continues to pass against the new backend and was run 8x back-to-back to rule out timing flakiness.

**Out of scope for this PR:** copy and move still walk GIO paths and have the same class of exposure. Extending this same descriptor-relative approach to them is a natural, separately-reviewable follow-up — the issue's acceptance criteria mention "copy/delete" together, so I'm leaving #8 open rather than closing it here.

## Visual evidence

Nothing in the UI changes, but here's a live reproduction of the fix itself: a folder (`target`) containing a real subdirectory plus a symlink (`decoy-link`) pointing outside the tree to a sentinel file, permanently deleted via Shift+Delete in Strata.

1. `target` selected, containing the decoy symlink. strata-pr253-1-selected

2. The permanent-delete confirmation. strata-pr253-2-confirm

3. `target` is gone after confirming. strata-pr253-3-after

4. The symlink's target folder, opened afterward — `sentinel.txt` untouched, same size and timestamp as before. strata-pr253-4-sentinel-intact

The same scenario is also covered deterministically by the two new adversarial unit tests (top-level and nested).

## How to test

1. Create a folder containing a real subdirectory with a file, plus a symlink pointing somewhere outside that folder (e.g. `ln -s /somewhere/else decoy`).
2. Select the folder in Strata and permanently delete it (Shift+Delete, or the "Permanently delete" context-menu item).
3. Confirm the folder and the symlink itself are gone, and whatever the symlink pointed to is untouched.
4. `cargo test` — two new tests (`permanent_delete_removes_a_symlink_standing_in_for_a_directory_without_following_it`, `permanent_delete_does_not_follow_a_symlink_nested_inside_the_tree`) cover this at both the top level and nested inside a recursive delete.

**Expected result:** a symlink anywhere in a deleted tree — whether it was always there or swapped in for a real directory — is removed as itself, never followed to whatever it points to.

## Related issue

Related to #8 (permanent-delete portion; copy/transfer hardening left as follow-up)

### DIFF
--- a/src/adapters/local_operations.rs
+++ b/src/adapters/local_operations.rs
@@ -644,11 +644,34 @@ enum LocalDeleteStep {
     },
 }
 
+/// Fails closed if an opened directory is no longer at its original name.
+fn ensure_local_delete_target_unchanged<ParentFd: AsFd, TargetFd: AsFd>(
+    parent: &ParentFd,
+    name: &OsStr,
+    target: &TargetFd,
+) -> Result<(), String> {
+    let named = rustix::fs::statat(parent, name, rustix::fs::AtFlags::SYMLINK_NOFOLLOW).map_err(
+        |error| {
+            format!(
+                "{} changed while it was being deleted: {error}",
+                name.to_string_lossy()
+            )
+        },
+    )?;
+    let opened = rustix::fs::fstat(target)
+        .map_err(|error| format!("Could not recheck {}: {error}", name.to_string_lossy()))?;
+    if named.st_dev != opened.st_dev || named.st_ino != opened.st_ino {
+        return Err(format!(
+            "{} changed while it was being deleted",
+            name.to_string_lossy()
+        ));
+    }
+    Ok(())
+}
+
 /// Inspects and, for non-directories, immediately deletes the entry named
 /// `name` inside `parent`. The type is re-read from disk here rather than
-/// trusted from any earlier listing, so a symlink swapped in for a
-/// directory is deleted as the symlink it now is instead of being opened as
-/// a directory.
+/// trusted from any earlier listing.
 fn open_local_delete_target<Fd: AsFd>(
     parent: &Fd,
     name: &OsStr,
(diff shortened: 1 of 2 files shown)


pdwaldrop:
strata-pr253-1-selected
strata-pr253-2-confirm
strata-pr253-3-after
strata-pr253-4-sentinel-intact

l0gicgate:
Maintainer security review and manual verification after merging current `main`:

### Review findings

The original implementation protected the final parent component with `O_NOFOLLOW`, but ordinary `open()` could still follow a symlink substituted into an earlier ancestor. It also kept traversing an opened directory if that directory was moved or replaced during deletion.

I pushed `f6a6b57` to address both findings:

- Resolve the complete parent path descriptor-relative from `/` with `openat2(RESOLVE_BENEATH | RESOLVE_NO_SYMLINKS | RESOLVE_NO_MAGICLINKS)`.
- Revalidate the opened directory's device/inode identity before each child deletion and before final removal, failing closed if its name now identifies another object or disappears.
- Add regression coverage for a symlinked parent component and for moving an opened directory during recursive deletion.

### Manual methodology and results

**Nested symlink test**

1. Created a target directory containing a regular nested file and a symlink to a sentinel directory outside the target tree.
2. Permanently deleted the target through Strata with Shift+Delete.
3. Checked the resulting filesystem directly.

Result: the selected target was removed; the outside sentinel remained present and unchanged.

**Symlinked parent-path test**

1. Created `parent-path/linked -> parent-path/actual`, with `actual/protected.txt` containing `must also survive`.
2. Navigated to the linked path through Strata and attempted to permanently delete `protected.txt`.
3. Checked both the UI result and underlying file directly.

Result: Strata reported `Could not safely open ... Too many levels of symbolic links (os error 40)`. `actual/protected.txt` remained present with its original contents.

The full local formatting, clippy, and test suite passed (531 tests). The moved-directory regression was also repeated 20 times locally without a failure.

l0gicgate (approved):
Reviewed the descriptor-relative deletion changes after the additional ancestor-resolution and directory-identity hardening. Automated and manual security scenarios pass.

PR #254, opened 2026-09-04
fix(operations): harden local copy against symlink races

pdwaldrop:
## Description

Follow-up to #253 (permanent-delete hardening), covering the other half of #8: copying. `copy_recursively` is the single function behind paste-copy, the cross-filesystem move fallback, and staged replace's copy-to-stage step — walking its source the same path-based way delete used to (enumerate, then re-resolve each child's own path next call), trusting whatever type GIO reported at listing time.

For fully local source/target pairs, `copy_recursively` now dispatches to a descriptor-relative walk, mirroring #253's approach:

- Each source directory is opened once via `openat2` with `RESOLVE_BENEATH | RESOLVE_NO_SYMLINKS | RESOLVE_NO_MAGICLINKS`, and every entry's real type is re-read immediately before deciding what to do with it — never trusted from an earlier listing.
- Symlinks are recreated directly (`readlinkat` + `symlink`), never followed.
- Regular files still go through GIO's own `copy_async` unchanged, except for one thing: instead of the original (re-resolvable) path, it's handed a `/proc/self/fd/<n>` reference to the fd already opened and verified via `openat2`, kept alive for the call's duration. That keeps GIO's metadata handling, progress reporting, and any reflink optimization completely untouched, while pinning what actually gets read to the exact object already checked — immune to the name being swapped afterward.
- The **destination** side stays on GIO throughout and needed no change: every call site here already writes to a private, freshly created staging name first and only becomes visible via the existing atomic `renameat2` commit, so it isn't exposed to the same class of race.
- Remote (GVfs) locations keep the existing path-based copy — explicit, not claiming an equivalent guarantee.
- Move's primary path is already a single atomic rename and needed no change; only its cross-filesystem fallback (when rename returns `WouldRecurse`) goes through this same hardened `copy_recursively`.

**One non-obvious wrinkle worth flagging for review:** `/proc/self/fd` entries are themselves reported as symlinks by `lstat`, so `NOFOLLOW_SYMLINKS` has to be dropped from the GIO copy call for the regular-file case specifically. Keeping it caused GIO to copy the magic-link's *target path text* as a new symlink instead of the file's actual bytes — caught immediately by the existing `moving_a_directory_falls_back_to_a_safe_copy_when_the_move_would_recurse` test, which started failing until this was found and fixed. Left a comment on that line explaining why.

## Visual evidence

N/A — this changes how local copies read the source tree internally; nothing in the UI changes. Two rounds of manual verification instead of screenshots (folder-listing screenshots of a copy operation don't show anything a picture can capture — the interesting part is on disk):

1. **Metadata preservation**: copied a folder where one file had unusual permissions (`754`) through the real app (Ctrl+C / Ctrl+V). The copy landed with `-rwxr-xr--` — exact match — confirming the `/proc/self/fd` handoff didn't regress GIO's metadata copying.
2. **Symlink safety**: copied a folder containing a symlink pointing outside the tree to a sentinel file, through the real app. Destination:
   ```
   to/source/nested/decoy -> .../outside-sentinel   (a real symlink, not a directory)
   ```
   `outside-sentinel/secret.txt` was never touched or duplicated into the destination.

New adversarial unit tests cover the same two scenarios (nested symlink, top-level symlink) deterministically.

## How to test

1. Create a folder containing a real subdirectory with a file, plus a symlink pointing somewhere outside that folder.
2. Copy it in Strata (Ctrl+C / Ctrl+V, or drag to another pane) into a new location.
3. Confirm the copy succeeds, the symlink is recreated as a symlink (not expanded into a real directory), and whatever it points to is untouched.
4. `cargo test` — `copy_recursively_does_not_follow_a_symlink_nested_inside_the_tree` and `copy_recursively_of_a_symlink_creates_a_symlink_not_a_recursive_copy`.

**Expected result:** a symlink anywhere in a copied tree is recreated as a symlink pointing at the same target, never followed into a real copy of its contents.

## Related issue

Related to #8 (copy half; delete covered by #253)

### DIFF
--- a/src/adapters/local_operations.rs
+++ b/src/adapters/local_operations.rs
@@ -10,7 +10,7 @@ use std::{
     future::Future,
     io,
     os::{
-        fd::{AsFd, OwnedFd},
+        fd::{AsFd, AsRawFd, OwnedFd},
         unix::{
             ffi::{OsStrExt, OsStringExt},
             fs::PermissionsExt,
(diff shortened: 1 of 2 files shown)


l0gicgate:
I exercised this locally and prepared a follow-up commit addressing four gaps found during review:

- reject FIFOs and other special filesystem objects before opening them (a read-only open of a FIFO otherwise blocked the transfer indefinitely and ignored cancellation);
- retain and revalidate `(st_dev, st_ino)` identities between directory enumeration, inspection, and descriptor opening so replacements fail clearly instead of copying a different object;
- open the complete local source parent path with `openat2` constraints, preventing intermediate symlink components from redirecting traversal;
- route staged replacement of non-directory sources through the descriptor-relative copy path too, while preserving overwrite behavior for copied symlinks.

Regression coverage was added for the FIFO hang, entry replacement, intermediate parent symlinks, and staged file-to-symlink replacement. The original nested and top-level symlink tests continue to pass.

Local checks:
- `cargo fmt --all --check`
- `cargo test --all-targets --all-features` (520 passed)

The follow-up is currently committed locally as `82a817d`; it has not been pushed to the contributor branch.

l0gicgate:
Manual testing of the follow-up commit also passed:

1. recursive local copy preserved nested contents and recreated an outside-pointing symlink without touching its sentinel;
2. regular-file mode `754` was preserved;
3. a directory containing a FIFO failed immediately with `Cannot copy special filesystem object named-pipe`, stayed responsive, and left no partial destination;
4. a source reached through an intermediate parent symlink failed closed with the expected safe-open error and created no destination;
5. staged regular-file replacement committed the new contents;
6. staged replacement of a regular file by a symlink produced the correct symlink and left its target unchanged;
7. a `/tmp` → `/dev/shm` cut/paste exercised the cross-filesystem move fallback, removed the source, preserved nested content and the symlink, and left the outside sentinel unchanged.

l0gicgate (approved):

PR #255, opened 2026-09-04
fix(operations): stop trusting stale types for replace/move cleanup

pdwaldrop:
## Description

**Depends on #253** (not yet merged) — built directly on that branch since this needs `permanently_delete_local_path` from it. GitHub is showing #253's commits as part of this diff for now; once #253 merges into `main`, this PR's diff will automatically narrow to just the commits below. The actual new content of this PR is the last commit only.

Four internal cleanup deletes still called the old GIO-based `permanently_delete`, each deciding directory-vs-file from a type determined once, well before the (potentially slow) operation it was cleaning up after:

- `replace_local_with`, after the atomic exchange: deleting whatever the old target turns out to be, using `target_is_directory` computed before the copy-to-stage step.
- `replace_local_with`, `move_source` cleanup: deleting the original source after a successful replace-as-move, using `source_is_directory` from the same early check.
- `move_local_with`'s `WouldRecurse` fallback: deleting the source after a cross-filesystem copy finishes.
- `copy_new_recursively`'s cancellation cleanup.

All five `permanently_delete` call sites (these four, plus the user-facing delete request handler from #253) now go through a new shared `permanently_delete_maybe_local` helper — local paths get the descriptor-relative walk, which re-reads each entry's real type itself and never accepts a caller-supplied hint at all; remote (GVfs) locations keep the existing GIO delete. This also de-duplicates the local-vs-remote dispatch that was written inline in #253's delete handler.

**Worth being direct about what this does and doesn't demonstrate**, since I want to flag this for review rather than let the commit message undersell or oversell it: GIO's `NOFOLLOW_SYMLINKS` already made a symlink standing in for a tracked directory at these specific call sites fail safe — I added a test for exactly that (`replacing_a_directory_cleans_up_a_symlink_in_the_old_contents_without_following_it`) and confirmed by temporarily reverting one call site that it **passes against the pre-existing code too**. So this isn't closing a bug that test demonstrates.

What it actually buys is protection the old path-based approach structurally cannot offer, and that a static symlink-substitution test can't exercise: once a directory is opened as a descriptor, nothing that happens to its *name* afterward — unlink-and-recreate, rename, anything short of touching the open file itself — can redirect an already-open fd to something else. That's a property of what a file descriptor is, not something provable by a single-shot unit test. This change closes that gap at every internal cleanup site instead of only the one path a user's delete click reaches directly, and stops trusting a type hint (`target_is_directory`/`source_is_directory`) that may be stale by the time it's acted on — matching the issue's "avoid trusting stale type metadata for recursion decisions" criterion, even though these aren't literally `FileEntry` values.

## Visual evidence

N/A — internal cleanup path, no UI change. Manually verified the replace-with-conflict flow still works correctly end-to-end in the app: pasted a folder onto an existing same-named folder, chose Replace, confirmed the new content landed and the old content was fully cleaned up (`old-item.txt` gone, `item.txt` has the new content).

## How to test

1. `cargo test` — the new `replacing_a_directory_cleans_up_a_symlink_in_the_old_contents_without_following_it`, plus the existing `staged_directory_replacement_does_not_merge_old_contents` / `staged_file_replacement_*` tests, which continue to pass unchanged against the new dispatch.
2. In the app: paste a folder onto an existing same-named folder and choose Replace. Confirm the new content lands and old content is fully gone.

**Expected result:** no behavior change visible to the user; internal cleanup after replace/move now shares the same race-safe, type-revalidating delete as the user-facing one.

## Related issue

Related to #8 (closes the remaining "stale type" gap across delete/copy/replace cleanup paths)

### DIFF
--- a/src/adapters/local_operations.rs
+++ b/src/adapters/local_operations.rs
@@ -565,6 +565,149 @@ fn permanently_delete(
     })
 }
 
+/// The outcome of resolving one delete target relative to its parent
+/// directory's file descriptor.
+enum LocalDeleteStep {
+    /// A non-directory entry (file, symlink, or other special file) that has
+    /// already been unlinked.
+    Removed,
+    /// A directory that was opened (not yet removed) along with its
+    /// immediate children, still to be deleted before the directory itself.
+    Directory {
+        handle: OwnedFd,
+        children: Vec<OsString>,
+    },
+}
+
+/// Inspects and, for non-directories, immediately deletes the entry named
+/// `name` inside `parent`. The type is re-read from disk here rather than
+/// trusted from any earlier listing, so a symlink swapped in for a
+/// directory is deleted as the symlink it now is instead of being opened as
+/// a directory.
+fn open_local_delete_target<Fd: AsFd>(
+    parent: &Fd,
+    name: &OsStr,
+) -> Result<LocalDeleteStep, String> {
+    let stat = rustix::fs::statat(parent, name, rustix::fs::AtFlags::SYMLINK_NOFOLLOW)
+        .map_err(|error| format!("Could not inspect {}: {error}", name.to_string_lossy()))?;
+    if !matches!(
+        rustix::fs::FileType::from_raw_mode(stat.st_mode),
+        rustix::fs::FileType::Directory
+    ) {
+        rustix::fs::unlinkat(parent, name, rustix::fs::AtFlags::empty())
+            .map_err(|error| format!("Could not delete {}: {error}", name.to_string_lossy()))?;
+        return Ok(LocalDeleteStep::Removed);
+    }
+    // RESOLVE_NO_SYMLINKS (stronger than O_NOFOLLOW) plus RESOLVE_BENEATH and
+    // RESOLVE_NO_MAGICLINKS: if `name` changed to a symlink (or a magic link)
+    // in the moment since the statat above, this fails closed instead of
+    // opening whatever it now points to.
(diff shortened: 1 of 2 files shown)


l0gicgate:
Merged the latest `main` into this branch and resolved the conflicts by retaining `main`'s newer parent-path and moved-directory protections while replaying this PR's cleanup dispatch changes.

During review I found that descriptor-relative deletion still opened the cleanup root only after the slow copy, so a process could substitute an unrelated object at the same path and have cleanup delete it. I pushed `d404662` to address that:

- capture local source/target identity (`st_dev` + `st_ino`) before slow replacement/move work;
- reject cleanup when the named object no longer has the expected identity;
- revalidate the replacement target immediately before the atomic exchange;
- preserve a substituted object after an exchange instead of allowing `tempfile`'s path-based `Drop` cleanup to remove it;
- retain the identity of copy roots created for cancellation cleanup; and
- add deterministic adversarial tests for source and target substitution.

Validation completed locally:

- `cargo fmt --all --check`
- `cargo test --all-targets --all-features` (621 passed)

The original manual folder-replacement fixture passed before the identity fix. Since the implementation changed afterward, I recommend one final manual replacement smoke test on the updated build before approval.

l0gicgate:
Repeated the manual replacement smoke test on the updated build after `d404662`: the incoming directory replaced the destination without merging, old-only entries and the old symlink were removed, the symlink's external sentinel remained intact, and the runtime log contained no errors. Manual verification passes.

PR #289, opened 2026-09-04
fix(browser): keep cut intent across windows and mount representations

### LINKED ISSUE #287: Cut/paste from context menu pastes as copy (cut state lost)
### Problem or use case
Cut via right-click context menu then Paste behaves like Copy: the source remains and a duplicate appears. Reported on /mnt NFS mount, same-window and cross-folder. Progress shows Copying instead of Moving, no error dialog.

### Where it breaks
- cut state is per-ViewState: src/ui/browser.rs cut_locations + paste_into same_locations check. Cross-window paste always sees empty cut list and degrades to copy. Clipboard only carries GDK FileList with no cut marker, so there is no cross-view source of truth.
- paste matching uses Location == (PathBuf) equality. Any normalization difference between stored FileEntry.location and clipboard round-trip via location_for_file (symlink /mnt, encoding, native vs URI) fails equality and degrades to copy.
- context-menu Cut/Copy does not call sync_mode_selection first (keyboard Ctrl+X/C does), so Grid/Explorer right-click can cut a stale selection.

### Proposed solution
- Share cut intent process-wide (thread-local) across all ViewStates, keep per-view list for dimming.
- Decide move_sources from local OR shared cut list, comparing via gio::File::equal as well as Location ==.
- Sync ModeViews selection inside cut_entries/copy_entries so context and keyboard paths match.
- Clear shared cut on copy/clear/successful move.

### Alternatives considered
- Full x-special/gnome-copied-files interop: larger change, follow-up for external Nautilus cut/paste. This fix stays internal but removes silent copy degradation.

Deoxizn:
## Description

Fixes Cut via right-click context menu pasting as Copy (source left behind, duplicate created). Three defects combined to cause the silent degradation:

- Cut intent lived only in per-`ViewState` `cut_locations`, while the GDK clipboard carries just a `FileList` with no cut marker. A cut in one window always pasted as a copy in another. Added a process-wide thread-local cut list as the source of truth for "was this a cut" (per-view list retained for dimming).
- `same_locations` used strict `Location ==` (`PathBuf`) equality. Clipboard round-trips through `location_for_file` can yield an equivalent but non-identical representation (native path vs `file://` URI, normalization on mounts such as NFS), failing the match and degrading to copy. Matching now also accepts GIO-level `equal`, with multiset semantics preserved for duplicates; `retain_untransferred` uses the same comparison.
- Context-menu Cut/Copy never called `sync_mode_selection` (keyboard Ctrl+X/C does), so Grid/Explorer right-click could operate on a stale selection. Sync moved inside `cut_entries`/`copy_entries` so both paths match.

No new dependencies. Full `x-special/gnome-copied-files` interop with external file managers is intentionally left as a follow-up.

## Visual evidence

N/A - no visual changes. Cut-row dimming behavior unchanged.

## How to test

1. Same window, local dir: Cut a file via right-click -> row dims -> Paste into another folder -> progress shows "Moving items", source removed.
2. Cross-window: Cut in window A -> Paste in window B -> moves, not copies.
3. Mount-equivalent: Cut `/mnt/share/file` (or any native path) -> Paste elsewhere -> moves; clipboard `file://` URI form still matches.
4. Grid/Explorer: multi-select files -> right-click Cut -> Paste -> all move, no stale selection.
5. Copy still copies: Copy -> Paste leaves source; follow-up Cut in another window cancels the first cut (paste copies).

**Expected result:** Cut+Paste always moves (with existing confirmation/collision flow); only Copy duplicates.

## Related issue

Closes #287

### DIFF
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -8,7 +8,7 @@ use std::{
     path::Path,
     pin::Pin,
     process::{Command, Stdio},
-    rc::Rc,
+    rc::{Rc, Weak},
     time::{Duration, Instant},
 };
 
@@ -293,7 +293,6 @@ pub(super) struct ViewState {
     mode_views: RefCell<ModeViews>,
     columns: RefCell<Vec<ColumnView>>,
     hovered_column: Cell<Option<usize>>,
-    cut_locations: RefCell<Vec<Location>>,
     horizontal_scroll_generation: Rc<Cell<u64>>,
     peek: RefCell<Option<PeekView>>,
     pending_peek: RefCell<Option<glib::SourceId>>,
@@ -449,7 +448,6 @@ impl BrowserView {
             mode_views: RefCell::new(mode_views),
             columns: RefCell::new(Vec::new()),
             hovered_column: Cell::new(None),
-            cut_locations: RefCell::new(Vec::new()),
             horizontal_scroll_generation: Rc::new(Cell::new(0)),
             peek: RefCell::new(None),
             pending_peek: RefCell::new(None),
@@ -480,6 +478,8 @@ impl BrowserView {
 
         // Columns are laid out from the start edge, so the blank strip beside the last
         // one is the natural place to begin a marquee that runs into it.
+        register_cut_view(&state);
+
         let weak_state = Rc::downgrade(&state);
         super::marquee::install_shared_origin_surface(&state.scroller, move |surface, _, x, _| {
             let state = weak_state.upgrade()?;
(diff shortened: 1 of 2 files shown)


Deoxizn:
Verified locally (toolchain installed since CI on fork PRs needs approval to run): `cargo fmt --all --check` clean, `cargo clippy --all-targets --all-features -- -D warnings` clean, `cargo test --all-targets --all-features` 531 passed including the 2 new cut-equivalence tests. The force-push fixes two issues the local build caught: nonexistent `Iterator::positions` replaced with `enumerate findings`, and doc-comments on the `thread_local!` macro changed to plain comments (`unused_doc_comments` deny).

l0gicgate:
Pushed follow-up commit 9be0192 addressing two review findings:

- Context-menu actions now synchronize Grid/Explorer selection before `context_entries` reads it, so Cut/Copy cannot capture stale browser selection.
- Removed per-window cut intent as a paste fallback. The process-wide cut list is now the sole source of truth for move-vs-copy and cut styling, and registered windows refresh together whenever a cut is set, cleared, or partially completed. This prevents a Copy in one window from being interpreted as an older Cut in another and keeps dimming synchronized.
- Added a regression test confirming that clearing shared cut intent cannot be revived by stale state.

The full local format, clippy, and test suite passes (532 tests).

l0gicgate:
Manual validation completed successfully on commit 9be0192:

- Context-menu multi-file Cut/Paste moved correctly in Grid, Explorer, and Columns.
- Cross-window Cut/Paste moved rather than copied.
- Copy remained a copy.
- Copying the same file from another window superseded an older Cut, cleared dimming across windows, and pasted as a copy.
- A newer Cut superseded an older Cut across windows and updated dimming in both.
- Collision Skip preserved the destination and kept the cut pending for a later successful paste; Replace moved the source and replaced destination contents.
- A temporary localhost SFTP/GVFS backend verified remote URI Cut/Paste as a move, with the source removed and destination content preserved.

Temporary fixtures and the test server were removed after verification.

PR #296, opened 2026-09-04
feat(sidebar): add eject and unmount for removable media

### LINKED ISSUE #295: Add sidebar eject/unmount for removable media
### Problem or use case
Removable media (USB sticks, second hard drives, optical discs) and plain disk mounts appear under DEVICES in the sidebar, but there is no eject icon and no right-click Eject/Unmount anywhere except SMB Disconnect. Nautilus shows an eject affordance per device; without one there is no safe-removal path and users resort to `umount` in a terminal (or worse, unplug live media).

### Proposed solution
- Show an inline eject button on sidebar device rows (volumes + volumeless mounts) whenever GIO reports the action is available, plus right-click Eject/Unmount entries reusing the existing sidebar context-menu pattern.
- Prefer Eject when the volume/drive reports `can_eject`, else Unmount when the mount reports `can_unmount`; hide the affordance entirely when neither is available (internal fixed disks).
- Add the Lucide `eject` icon as `strata-eject` following the existing namespaced-asset convention.
- After a successful release, navigate home if the active location was inside the removed device; surface failures (except user-cancelled) in the standard error dialog.

### Alternatives considered
- Terminal-only `umount`: undiscoverable and unsafe for optical/USB (no eject ioctl).
- Always showing Unmount even when GIO reports it unavailable: leads to guaranteed-failure buttons on internal disks.

Deoxizn:
## Description

Removable media (USB sticks, second hard drives, optical discs) and plain disk mounts appear under DEVICES with no safe-removal path — only SMB shares have Disconnect. Adds Nautilus-style ejection: an inline ⏏ button on sidebar device rows plus a right-click Eject/Unmount entry, shown only when GIO reports the action (`can_eject` preferred, else `can_unmount`; hidden on fixed internal disks).

Details: new `strata-eject` icon from Lucide (geometry intact, theme-recolored like other `strata-*` icons); volume eject uses `Volume::eject`, mount release uses mount eject/unmount; after success navigates home if the active location was inside the device; failures other than user-cancel show the standard error dialog. Device rows wrap the row button + eject sibling so row clicks never misfire, and active-row highlighting accounts for the wrapper.

## Visual evidence

- Sidebar DEVICES rows for USB/HDD/optical now show ⏏ at the row end; right-click offers Eject/Unmount with the eject icon. (Screenshot to be attached after a device-present run.)

## How to test

1. `cargo run`, plug in a USB stick -> DEVICES shows it with an ⏏ button.
2. Click ⏏ (or right-click -> Eject) -> device ejects; if browsing inside it, view returns home.
3. Right-click a fixed internal disk row -> no Eject/Unmount entry.
4. Unmountable plain mount (non-SMB) -> Unmount entry; cancel in the auth dialog -> no error shown; real failure -> "Unable to unmount device" dialog.
5. `cargo test --all-targets --all-features` (618 passed incl. 3 new release-matrix tests).

**Expected result:** Ejectable media can be safely removed from the sidebar; internal disks unchanged.

## Related issue

Closes #295

### DIFF
--- a/data/icons/scalable/actions/strata-eject.svg
+++ b/data/icons/scalable/actions/strata-eject.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="#8bc9eb" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M4 13a1 1 0 0 1-.72-1.695l7.257-7.668a2 2 0 0 1 2.926 0l7.256 7.668A1 1 0 0 1 20 13z"/><rect x="3" y="17" width="18" height="4" rx="1"/></svg>
--- a/data/strata.gresource.xml
+++ b/data/strata.gresource.xml
@@ -14,6 +14,7 @@
     <file>icons/scalable/actions/strata-copy.svg</file>
     <file>icons/scalable/actions/strata-corner-down-left.svg</file>
     <file>icons/scalable/actions/strata-download.svg</file>
+    <file>icons/scalable/actions/strata-eject.svg</file>
     <file>icons/scalable/actions/strata-eye-off.svg</file>
     <file>icons/scalable/actions/strata-eye.svg</file>
     <file>icons/scalable/actions/strata-external-link.svg</file>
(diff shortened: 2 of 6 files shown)


l0gicgate:
Reviewed and manually tested this with a mounted Kingston USB device. The inline eject button and context-menu Eject action both worked, navigation returned Home after ejecting from inside the device, and fixed internal volumes remained unchanged.

I also pushed follow-up commit 81cc5b0 addressing two review findings:

- `EjectMount` now invokes the mount's eject operation rather than incorrectly calling eject on a volume that reported it unsupported.
- Inline and context-menu release actions now share an in-flight guard, preventing repeated clicks from starting concurrent eject/unmount operations; the guard resets after completion so failures can be retried.
- Added regression coverage for the in-flight guard.

The full local checks pass, including all 619 tests.

PR #301, opened 2026-09-04
feat(operations): undo completed moves with Ctrl+Z

### LINKED ISSUE #298: feat(operations): undo completed moves with Ctrl+Z
### Problem or use case

After completing a Cut and Paste move, pressing Ctrl+Z does not move the files back to their original locations. Ctrl+Z currently only restores the latest move to Trash. Users expect the most recent reversible file operation to be undone.

### Proposed solution

Track the source and final destination of each successfully completed move, including partial and collision-resolved transfers. Make Ctrl+Z undo the latest reversible operation process-wide by moving those entries back to their original locations. Preserve retry state for partial failures and avoid overwriting new conflicts without confirmation.

### Alternatives considered

Keep Ctrl+Z limited to Trash restoration or add a separate shortcut for move undo. A unified latest-operation undo better matches common file-manager behavior, but it should be implemented separately from the pending-cut fix because it requires operation history, reverse-transfer conflict handling, partial-failure behavior, and tests.

### Visual examples

N/A — behavior and operation-state change; existing progress and conflict UI can be reused where appropriate.

### Strata version

0.9.1 / PR #289 test build

### Additional context

Raised during manual verification of #289. Requested behavior is specifically undo after Paste has completed, not cancellation of a pending Cut.

spandan11106:
## Description

`Ctrl+Z` previously only restored the latest move to Trash, so a completed Cut and Paste could not be reversed. Each completed move now records where its items started and where they landed, and `Ctrl+Z` undoes whichever reversible operation happened last — a move or a move to Trash — process-wide across windows.

- Partial, failed, and cancelled moves record the items that actually transferred, and an undo that only partly succeeds leaves the remainder pending so the next `Ctrl+Z` retries it.
- An item whose original location is occupied again asks for confirmation before overwriting, reusing the existing transfer conflict dialog (Replace / Skip / apply to all); items that no longer exist where the move left them are dropped.
- An undo does not itself register a new undo, so `Ctrl+Z` never toggles an item back and forth, and it leaves any pending Cut untouched.

## Visual evidence

N/A — behavior-only change. The undo reuses the existing move progress overlay and the existing "File already exists" conflict dialog, so there is nothing new to show.

## How to test

1. Create two folders, put a file in the first one, and open the first folder in Strata.
2. Select the file, press `Ctrl+X`, navigate to the second folder, and press `Ctrl+V`.
3. Once the move finishes, press `Ctrl+Z`.
4. Repeat steps 1–2, then recreate a file with the same name in the original folder before pressing `Ctrl+Z`.
5. Move a file to Trash and press `Ctrl+Z`.

**Expected result:** step 3 moves the file back to the folder it was cut from, and a second `Ctrl+Z` does nothing. Step 4 asks whether to replace the newer file, and skipping it leaves both files where they are. Step 5 still restores from Trash.

## Related issue

Closes #298

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ strata                 # home directory
 strata ~/Documents     # a specific directory
 ```
 
-Useful shortcuts include <kbd>Ctrl</kbd>+<kbd>K</kbd> for recursive search, <kbd>Ctrl</kbd>+<kbd>L</kbd> for a path or URI, <kbd>Ctrl</kbd>+<kbd>F</kbd> to filter the current pane, <kbd>Ctrl</kbd>+<kbd>Z</kbd> to undo the latest move to Trash, <kbd>Space</kbd> for preview, <kbd>F2</kbd> to rename, and <kbd>Alt</kbd>+arrow keys for history and parent navigation.
+Useful shortcuts include <kbd>Ctrl</kbd>+<kbd>K</kbd> for recursive search, <kbd>Ctrl</kbd>+<kbd>L</kbd> for a path or URI, <kbd>Ctrl</kbd>+<kbd>F</kbd> to filter the current pane, <kbd>Ctrl</kbd>+<kbd>Z</kbd> to undo the latest move or move to Trash, <kbd>Space</kbd> for preview, <kbd>F2</kbd> to rename, and <kbd>Alt</kbd>+arrow keys for history and parent navigation.
 
 ### Desktop entry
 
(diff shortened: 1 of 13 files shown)


spandan11106:
Video proof


https://github.com/user-attachments/assets/10be630f-d49f-4a23-8f41-3de19072ea4a

l0gicgate:
Review follow-up:

- Merged the latest `main` and resolved the `src/ui/window/tests.rs` conflict while retaining both the undo-shortcut and removable-media test coverage.
- Fixed partial move-undo retry state so items intentionally skipped or found missing before transfer are not offered again if another item fails.
- Preserved byte-safe filenames when deriving URI transfer targets, so non-UTF-8 names are not silently omitted from move undo history.
- Added regression coverage for the partial retry-state case.

Manual verification completed successfully for move undo, overwrite confirmation/skip behavior, and Trash undo. Local formatting, Clippy (`-D warnings`), and all 647 tests pass.

PR #307, opened 2026-09-04
feat(search): start file search by typing

### LINKED ISSUE #304: feat(search): start file search by typing
## Problem or use case
Opening search currently requires Ctrl+K or clicking the search button. Keyboard-first navigation benefits from beginning a search as soon as the user types while the file list has focus.

## Proposed solution
Add an opt-in Settings preference named “Type to search”. When it is enabled and the file list has focus, typing a printable character opens search and places that character in the query field. Reopening after dismissing should work repeatedly, and the caret should remain after the initial character.

## Alternatives considered
Continue using Ctrl+K for every search. This remains available when the preference is disabled or as the explicit shortcut.

## Visual examples
N/A — existing search UI is reused; only its keyboard entry point changes.

## Strata version
0.9.1 (built from source)

## Additional context
Manually verified locally: direct typing searches correctly, repeated dismiss-and-search cycles work, and Ctrl+K behavior remains intact.

GavidetDoliath:
## Description
Adds an opt-in **Type to search** preference. Typing with a file view focused opens the active pane’s inline filter (Ctrl+F), not the global search dialog. Ctrl+K remains available for global search.

- Merged `main` and resolved conflicts.
- Preserved the initial character by focusing the seeded entry without selecting its text; subsequent typing appends instead of replacing it. Removed the ineffective deferred-focus workaround.
- Enabled recursive local-folder search in Columns, Explorer, and Grid. Explorer/Grid previously filtered only loaded entries; they now show an in-pane result list and restore their normal presentation when the query is dismissed.
- Kept results scoped to the pane’s folder and descendants. Added full-path subtitles and tooltips so identically named results can be distinguished.
- Preferred nearby matches over deeper duplicates without letting proximity outweigh stronger matches.
- Fixed results stopping after a pause in typing by keeping the result listener alive across query changes. Search work is cancelled when the query is cleared.
- Updated the setting description and added regression coverage for initial-character preservation, scope, ranking, nested results, query changes, reopening, and keyboard activation.

## Visual evidence
Searching `ext` from `Documents/demo` shows direct and nested matches with their origins:

Recursive inline search with full result paths

## How to test
1. Enable **Settings → General → Type to search**.
2. In each of Columns, Explorer, and Grid, focus a local folder’s file view and type `LICENSE`. Confirm the complete query is preserved, including `L`.
3. From Documents, type `stra` with a `Documents/github/strata` fixture. Confirm the nested folder appears with its full path and that matches outside Documents are excluded.
4. Search for duplicate filenames at different depths. Confirm nearer duplicates appear first; pause and edit the query to confirm results continue updating.
5. Use Up/Down and Enter to navigate results. Clear or dismiss the query with Escape, then type again. In Explorer/Grid, confirm the normal presentation is restored on dismissal.
6. Verify Ctrl+F still opens the inline filter and Ctrl+K still opens global search. Disable Type to search and confirm ordinary browser keyboard behavior is restored.

**Expected result:** typing starts a scoped recursive search without losing characters, results identify their origins, and repeated searches work across all three views. Opening a folder result navigates into it; opening a file result navigates to its containing folder.

## Related issue
Closes #304


l0gicgate:
@GavidetDoliath thank you for this contribution, can you make sure to include video evidence in PRs that include behavioral changes. Thank you 🙏🏼

l0gicgate:
I'm not sure if we should open the global search, instead this should probably open the in pane filter search and start typing there.
image

l0gicgate:
Merged the latest `main` and resolved the resulting conflict.

Addressed the review feedback by changing type-to-search to open the active pane’s inline filter (the same filter used by Ctrl+F) and seed it with the typed character, rather than opening global search. I also updated the setting description to match the new behavior.

Reviewed the updated diff and found no remaining issues. The full local checks pass: formatting, Clippy with warnings denied, and all 694 tests.

l0gicgate:
Added full-path subtitles to inline search results (up to two lines, with the complete path in a tooltip). Matching filenames nearer the search root now rank above deeper duplicates, while exact matches retain priority over weaker matches. Also fixed the result listener being invalidated when the query changes after a pause.

Manually exercised an isolated Videos fixture with matching files directly in Videos, in archive, and in archive/deep. Typing screenrecording.mp4 shows those results in that order with their paths below the filenames. Added automated coverage for proximity ranking and excluding results outside the search root.

Inline search paths and nearby-first duplicates

l0gicgate:
Fixed the missing recursive search in Explorer and Grid: these previously filtered only loaded entries, unlike Columns. Local-folder queries now show an in-pane recursive result list with full-path subtitles and the shared proximity ranking. Clearing or dismissing the query restores the original presentation; Enter opens the selected result.

Verified with Documents/github/strata: type stra from Documents in either mode, confirm the nested folder appears, press Enter to open it, and dismiss another query with Escape to restore the normal view. Added a GTK regression test covering nested results, scope, query changes, reopening, and keyboard activation.

Explorer recursive results
Grid recursive results

l0gicgate:
Added the supplied demo screenshot to the Visual evidence section and rewrote the PR description to summarize the complete changes: merging main, switching typing to the inline filter, fixing the selected first character, recursive search in all three views, full result paths, proximity ranking, query-listener lifetime, and regression coverage. The description also includes updated manual testing steps. All local checks pass, including 698 tests.

PR #308, opened 2026-09-04
feat(explorer): add Mode column showing file permissions

JoeJoeflyn:
## Summary
- Adds a **Mode** column to the Explorer view, placed between Name and Size
- Displays the symbolic unix permission string (e.g. `rw-r--r--  644`) using the existing `format_permissions` helper from the properties dialog
- The column is non-sortable (no sort arrow or click handler)
- Queries `unix::mode` via GIO alongside the existing directory attributes
- Adds `mode: MetadataValue<u32>` to `FileEntry`

#### Test plan
- [x] Switch to Explorer view — Mode column appears after Name, before Size
- [x] Permission strings display correctly for files and folders
- [x] Column is resizable like the others
- [x] Clicking the Mode header does not trigger sorting
- [x] Other columns still sort correctly
image

### DIFF
--- a/src/adapters/local_files.rs
+++ b/src/adapters/local_files.rs
@@ -25,8 +25,8 @@ use crate::{
 
 const LIST_ATTRIBUTES: &str =
     "standard::display-name,standard::name,standard::type,standard::is-hidden,standard::is-symlink";
-const FULL_ATTRIBUTES: &str = "standard::display-name,standard::name,standard::type,standard::is-hidden,standard::is-symlink,standard::size,time::modified";
-const METADATA_ATTRIBUTES: &str = "standard::type,standard::size,time::modified";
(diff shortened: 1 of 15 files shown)


l0gicgate:
Follow-up review fixes pushed:

- Populated `mode` in the fast native scanner and propagated it through asynchronous metadata updates.
- Distinguished unsupported mode metadata from a valid zero mode.
- Requested mode metadata only for Explorer rows, avoiding extra work in Grid view.
- Repainted bound Explorer rows when metadata arrives; this also fixes Size and Modified remaining blank after asynchronous fills.
- Added coverage for native enumeration, metadata fills, state updates, and Explorer metadata requests.

Manually verified that Mode now displays correctly. The full local suite passes: formatting, Clippy with warnings denied, and all 623 tests.

PR #309, opened 2026-09-04
fix(navigation): keep the address-bar scrollbar from expanding

### LINKED ISSUE #297: fix(navigation): keep the address-bar scrollbar from expanding
## Strata version

0.9.1 / current development version.

## Installation method

Built from source.

Omarchy, Hyprland, Wayland.

## What happened?

When the breadcrumb path overflows the address bar, its horizontal scrollbar expands on hover or interaction. This changes the height/spacing of the address-bar area and creates a distracting layout shift.

The browser panes already keep their scrollbars at a fixed narrow size, so the address bar behaves inconsistently with the rest of the interface.

## Steps to reproduce

1. Navigate to a sufficiently deep path so the address-bar breadcrumbs overflow horizontally.
2. Move the pointer over the horizontal scrollbar or interact with it.
3. Observe that the scrollbar expands and changes the address-bar layout.

## Expected behavior

The address-bar scrollbar should remain a fixed size in its normal, hover, and dragging states, matching the non-expanding scrollbar behavior used by the browser panes. It should remain usable without causing the address bar or surrounding toolbar to resize.

## Screenshots or video

N/A — the behavior is the scrollbar expanding during hover/interaction.

## Relevant logs

No relevant logs.

## Additional context

The breadcrumb `ScrolledWindow` does not currently receive the shared `fixed-scrollbar` CSS class used by pane scrollers. Reusing the established treatment should keep this behavior consistent without introducing separate styling.

spandan11106:
## Description

The breadcrumb `ScrolledWindow` was the only scroller in the interface without the shared
`fixed-scrollbar` class, so its horizontal scrollbar grew on hover and while dragging and shifted
the address-bar layout. Adding the class reuses the treatment the browser panes already have and
keeps the scrollbar at a constant size in every state.

## Visual evidence

Verified on Hyprland with a temporary probe build that forced the breadcrumb scrollbar to stay
visible, so the hover geometry could be compared directly:

- Before: hovering the address-bar scrollbar drew a full-height trough with a fat slider
  (~13 px tall), pushing the breadcrumb row.
- After: the same hover keeps a 5 px scrollbar with a 3 px slider and no trough, identical to the
  pane scrollers, and the address-bar height does not change.

Screenshots could not be attached from the CLI; the comparison above is what the probe build showed.

## How to test

1. Open a path deep enough that the address-bar breadcrumbs overflow horizontally.
2. Hover the pointer over the breadcrumb scrollbar and drag it.

**Expected result:** the scrollbar stays the same narrow size while hovering and dragging, remains
usable for scrolling the breadcrumbs, and the address bar and toolbar do not change height.

## Related issue

Closes #297

### DIFF
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -409,6 +409,7 @@ impl BrowserView {
             .vscrollbar_policy(gtk::PolicyType::Never)
             .hexpand(true)
             .build();
+        breadcrumb_scroller.add_css_class("fixed-scrollbar");
         let location_stack = gtk::Stack::builder()
             .hhomogeneous(false)
             .vhomogeneous(false)


spandan11106:
Before :

https://github.com/user-attachments/assets/c38fbee9-2d91-472a-9a37-dfea4e372405

After : 

https://github.com/user-attachments/assets/7bdeff65-39a2-43b5-a4ec-9b67ee60a9a7

PR #310, opened 2026-09-04
perf(browser): keep large listings scrolling after mode and filter ch…

### LINKED ISSUE #305: perf(browser): Keep large Grid and Explorer listings scrolling after mode and filter changes
**Problem or use case**

In a large directory, Grid and Explorer scrolling already feels heavy. After switching view modes or changing the filter, it gets worse instead of staying the same: each visit is more sluggish, and a mode switch often shows only one row until you scroll. Columns was already using a position map; Grid and Explorer still walked the source list for every visible row, and leftover widget cycles kept old panes alive.

**Proposed solution**

Map Grid/Explorer source items to indexes in O(1) instead of walking the list on bind. Tear down collection views and break widget cycles on mode switch (marquee, miller handles, row controllers, thumbnail settle groups) so RSS does not grow across visits. Show the target mode before rebuilding it so ListView measures a real viewport. Classify filter edits as more/less strict so Grid and Explorer can notify the filter instead of rebuilding.

**Alternatives considered**

Leave Grid/Explorer on a linear source-index walk: that is the current cost and it grows with directory size. Rebuild the whole model on every filter keystroke: simpler, but slower than a MoreStrict/LessStrict notify. Keep strong widget refs in gestures and factories: easier to write, but it pins the old pane after a mode switch.

**Visual examples**

https://github.com/user-attachments/assets/f2a9af48-f99c-45a5-8df6-bedcebd5bf70

**Strata version**
0.9.1 (source, main 708eedb)

**Additional context**

Please check:
- All three modes, including after several Grid ↔ Explorer ↔ Columns switches
- Filter, hidden files, and type-group changes
- Selection, context menu, drag, peek, new item, rename
- Live add/remove while Grid or Explorer is open
- After a mode switch, more than one row is visible without scrolling first

Failure mode to watch: activating or previewing the wrong item (clicked A, opened B) after a filter or sort change.

ogarza:
## Description

Three things stacked in large Grid/Explorer folders:

1. **Bind was O(visible × n).** Every visible row walked the `StringList` to turn a filter/sort index into a source index. Columns already had a position map; Grid and Explorer did not.
2. **Mode switches leaked the old pane.** Gestures and factories held strong widget refs, so each Grid ↔ Explorer ↔ Columns visit left the previous listing alive. Scrolling then got worse, not better.
3. **ListView measured while hidden.** Rebuild + `scroll_to` ran before the stack page was shown (or before the pane was parented), so GTK thought the viewport was one row until you scrolled.

**Where to look**

| What | File | Start here |
| --- | --- | --- |
| O(1) source index for Grid/Explorer bind, activate, peek, drag, selection | `src/ui/browser_modes.rs` | `SourceIndexMap` (`watch` / `apply` / `rebuild`). `watch()` rebuilds immediately: `StringList::new` does not emit `items_changed` for existing rows. The `items_changed` handler must use the signal’s `source`, not a cloned list, or the list stays pinned. |
| Show the new mode, then rebuild, then drop the old one | `src/ui/browser.rs` | `BrowserView::set_view_mode` |
| Parent the pane, then apply the snapshot (avoids the one-row viewport) | `src/ui/browser_modes.rs` | `rebuild_grid`, `rebuild_explorer` |
| Idle `scroll_to` / `grab_focus` after layout | `src/ui/browser_modes.rs` | `scroll_collection_to`, `focus_visible_pane` |
| Drop factory + model when a pane goes away | `src/ui/browser.rs`, `src/ui/browser_modes.rs` | `detach_collection_view`, `detach_pane_models` |
| Marquee must not pin the collection after a mode switch | `src/ui/marquee.rs` | `MarqueeState` is `WeakRef`; drag handlers take the widget from the gesture |
| Miller resize handle / reveal tick must not pin the column shell | `src/ui/browser.rs` | `shell.downgrade()` on the overlay handle and the miller scroller tick |
| Row controllers must not cycle `ListItem` ↔ child | miller factory in `src/ui/browser.rs`; `install_preview_click`, `install_grid_peek`, `install_explorer_drag_drop` in `src/ui/browser_modes.rs` | `item.downgrade()` / `row.downgrade()` |
| Dead hooked thumbnail groups must drop | `src/ui/thumbnail.rs` | `cancel_thumbnail` retain: keep only if pending work remains, or the viewport is still alive |
| Filter keystrokes notify MoreStrict/LessStrict instead of rebuilding | `src/ui/browser.rs` | `filter_change_for`, `notify_filter_query` (Grid/Explorer). Do not copy miller’s empty-query `set_filter(None)`: that would drop the hidden-files predicate. |

Tests match those pieces: `SourceIndexMap` (filter/sort/placeholder + unpin), miller `rebuild_position_map`, marquee unpin, `filter_change_for` / `notify_filter_query`, thumbnail dead-viewport drop. Mode-switch “one row until scroll” still needs the manual check below.

## Visual evidence

https://github.com/user-attachments/assets/a0fe28cd-10e1-4ec4-b218-9ee6b8f632f6

## How to test

Use a large local directory (tens of thousands of entries helps).

1. Open it in Grid, fling-scroll, then switch Grid → Explorer → Columns and back. Repeat several times.
2. Change the filter, hidden-files toggle, and type grouping in Grid and Explorer.
3. Select, open the context menu, drag, peek, create, and rename items.
4. Add or remove files in the folder while Grid or Explorer is open.
5. After a mode switch, confirm more than one row is visible without scrolling first.

**Expected result:** Scrolling stays usable after mode and filter changes and does not get worse on each visit. A mode switch shows a full viewport immediately. Activating or previewing an item still opens the one you clicked (not a neighbor after filter/sort).

## Related issue

Closes #305

### DIFF
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -4439,14 +4439,11 @@ impl ViewState {
             scroll_column_to(column, position);
         }
         column.list.grab_focus();
-        let list = column.list.clone();
+        let list = column.list.downgrade();
         glib::idle_add_local_once(move || {
-            if let Some(position) = position
-                && position < list.model().map_or(0, |model| model.n_items())
-            {
-                list.scroll_to(position, gtk::ListScrollFlags::FOCUS, None);
+            if let Some(list) = list.upgrade() {
+                list.grab_focus();
             }
-            list.grab_focus();
         });
     }
 
@@ -4905,15 +4902,26 @@ impl ViewState {
                 gtk::gdk::FileList::static_type(),
                 gtk::gdk::DragAction::COPY | gtk::gdk::DragAction::MOVE,
             );
-            let highlighted_row = row.clone();
-            let highlight = move |target: &gtk::DropTarget, _, _| {
-                highlighted_row.add_css_class("drop-destination");
+            let highlighted_row = row.downgrade();
+            drop.connect_enter(move |target, _, _| {
+                if let Some(row) = highlighted_row.upgrade() {
+                    row.add_css_class("drop-destination");
+                }
                 file_drop_action(target)
-            };
-            drop.connect_enter(highlight.clone());
-            drop.connect_motion(highlight);
-            let highlighted_row = row.clone();
-            drop.connect_leave(move |_| highlighted_row.remove_css_class("drop-destination"));
+            });
+            let highlighted_row = row.downgrade();
+            drop.connect_motion(move |target, _, _| {
+                if let Some(row) = highlighted_row.upgrade() {
+                    row.add_css_class("drop-destination");
+                }
+                file_drop_action(target)
+            });
+            let highlighted_row = row.downgrade();
+            drop.connect_leave(move |_| {
+                if let Some(row) = highlighted_row.upgrade() {
+                    row.remove_css_class("drop-destination");
+                }
+            });
             let weak_state_for_accept = weak_state.clone();
             let accepted_item = item.downgrade();
             let map_for_accept = map_for_hover.clone();
@@ -4937,8 +4945,11 @@ impl ViewState {
             let weak_state_for_drop = weak_state.clone();
             let dropped_item = item.downgrade();
             let map_for_drop = map_for_hover.clone();
-            let dropped_row = row.clone();
+            let dropped_row = row.downgrade();
             drop.connect_drop(move |target, value, _, _| {
+                let Some(dropped_row) = dropped_row.upgrade() else {
+                    return false;
+                };
                 dropped_row.remove_css_class("drop-destination");
                 let Some(state) = weak_state_for_drop.upgrade() else {
                     return false;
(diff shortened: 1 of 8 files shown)


ogarza:
1. Miller columns leak autoscroll markers on the window overlay

install_autoscroll parents the marker on self.overlay (the long-lived browser overlay). truncate() already removes each column’s marquee band, but never the autoscroll marker. Each opened column leaves a hidden 22×22 overlay child behind.

Fix: pass the per-column overlay (column_overlay) into install_autoscroll, or remove_overlay the marker when the column is destroyed — same pattern as the marquee band.

l0gicgate:
Manual review and testing update:

- Reviewed the source-index mapping, pane teardown, weak widget references, deferred layout scroll, and filter-change paths. I did not find a blocking correctness issue.
- Exercised the 100,000-entry fixture across repeated Grid → Explorer → Columns → Grid cycles. Full viewports appeared immediately, scrolling did not degrade across visits, and no blank panes, freezes, crashes, stale cards, or incorrect rendering appeared.
- Grid scrollbar-thumb dragging was noticeably slower than Page Up/Down and the list-based views. This is not a regression. We tested an 80 ms scroll-settle path that defers Grid thumbnail and metadata work during continuous adjustment changes; it made release-build thumb dragging substantially smoother and rendered the correct content after settling. That change is included in `a58bb85`.
- Grid still does not fully match Explorer/Columns during large random thumb jumps, so the remaining GtkGridView binding/layout work is tracked separately in #322.

Local formatting, Clippy (`-D warnings`), and the full 636-test suite passed after the change.

PR #311, opened 2026-09-04
feat(navigation): add fast scrolling across browser views

### LINKED ISSUE #293: feat(navigation): add fast scrolling across browser views
## Problem or use case

Navigating large directories is unnecessarily slow in Grid, Explorer, and Columns views:

- The scrollbar is too difficult to grab and drag for quickly jumping through a long directory.
- Mouse-wheel scrolling only advances incrementally; there is no middle-click autoscroll for sustained, speed-controlled movement.
- `Page Up` and `Page Down` do not provide page-sized keyboard navigation.

Users are therefore forced to repeatedly scroll the wheel or step through items when they want to move a substantial distance.

## Proposed solution

Provide consistent fast-scrolling behavior across all browser views:

- Make visible scrollbars reliably draggable, with a practical pointer hit target even if the visual thumb remains narrow.
- Support middle-click autoscroll in file-view scrolling areas. Movement away from the anchor should control direction and speed, and another click or `Escape` should stop it.
- Bind `Page Up` and `Page Down` to move by approximately one viewport while preserving a clear focused/selected item and keeping it visible.
- Apply scrolling to the active pane or column so shortcuts and pointer actions do not move an unrelated view.
- Ensure these gestures do not conflict with existing mouse-history buttons, selection, drag-and-drop, or text-field middle-click behavior.

## Alternatives considered

- Increase wheel speed globally: this may make precise scrolling harder and still does not provide direct scrollbar or keyboard navigation.
- Only enlarge the scrollbar: improves pointer dragging but leaves mouse and keyboard fast-scrolling gaps.
- Rely on search for large directories: search helps when the destination is known, but does not replace rapid browsing.

## Visual examples

N/A — this concerns pointer hit targets and scrolling interactions across existing views.

## Strata version

0.9.1 development build (commit `b662e3f`).

## Additional context

The browser uses very narrow fixed scrollbars (approximately 5 px with a 3 px slider), which likely contributes to the difficulty grabbing them. The interaction area can be widened independently of the visual treatment if necessary.

spandan11106:
## Description

Adds the three fast-scrolling gestures the issue asks for, in Grid, Explorer, and Columns:

- **Scrollbar grab target.** The file views' scrollbar troughs are now 14 px wide instead of 5 px, with the slider margins keeping the visible bar the same 3 px it was. Pressing anywhere in the trough jumps and drags, so the bar is practical to grab without changing how it looks.
- **Middle-click autoscroll.** A new `ui::scrolling` module installs autoscroll on each file view's scrolling area. The press point is marked with a themed anchor, movement away from it sets direction and speed (dead zone, then quadratic ramp to a cap, both axes where the view can scroll), and another click anywhere or `Escape` stops it. A press over a text field is left alone so middle-click paste still works, and mouse buttons 8/9 keep driving history.
- **`Page Up` / `Page Down`.** They move the focus by about one viewport of the view that has focus — so the active pane or column only — keeping one row of overlap, one selected item, and scrolling it back into sight. Grid mode counts the items per row, so a page is rows × columns.

## Visual evidence

N/A. The scrollbar's visual treatment is deliberately unchanged (only the pointer target grows), and the other two changes are pointer/keyboard interactions. The one new drawn element, the autoscroll anchor, is a themed circle at the press point.

## How to test

1. Open a directory with several hundred entries and switch between Columns, Grid, and Explorer.
2. Press and drag the scrollbar at the right edge of a file list; also press once part-way down the empty trough.
3. Middle-click in the file area, move the pointer away from the anchor and back, then stop with `Escape`; repeat and stop with a left click. Middle-click in the filter field to confirm it still pastes.
4. Focus a file list and press `Page Down` and `Page Up` several times, including at the top and bottom of the directory. Repeat with a second column open so only the focused one moves.

**Expected result:** the scrollbar can be grabbed and dragged from anywhere in its track while looking the same as before; middle-click scrolls continuously in the direction of the pointer, faster the further away it is, and stops on `Escape` or a click; `Page Up`/`Page Down` move the selection about a screenful at a time in the focused pane or column, stopping at the first and last entry, with the selected item always visible.

## Related issue

Closes #293


spandan11106:
https://github.com/user-attachments/assets/be39cd15-4e3e-443a-bf03-1280380f4290

l0gicgate:
Merged the latest `main` into this branch and resolved the `src/ui/window.rs` / `src/ui/window/tests.rs` conflicts by preserving the Page Up/Page Down handling while adopting main’s generalized undo shortcut.

During review I found and fixed one theming issue: the autoscroll anchor shadow used a static black color, so it now uses the semantic `@theme_border` token and continues to update with live theme changes. No other issues found. The full local format, clippy, and test suite passes.

l0gicgate:
Updated the scrollbar styling from the review feedback:

- Active/held scrollbar thumbs now use the live semantic `@theme_accent` color globally, including the more-specific theme catalog and settings scrollbars.
- Browser scrollbar thumbs are doubled from 3 px to 6 px while retaining the 14 px grab target.

Pushed as `1d51094`; the full local format, clippy, and test suite still passes.

PR #313, opened 2026-09-04
feat(omarchy): offer one-time default file manager setup

### LINKED ISSUE #299: feat(omarchy): offer one-time default file manager setup
## Problem or use case

Installing Strata on Omarchy does not make it the complete default file manager. Users must separately:

1. Set `io.github.lgse.Strata.desktop` as the XDG handler for `inode/directory`.
2. Replace Omarchy's Nautilus file-manager shortcuts with Strata-compatible bindings.

This setup is easy to miss and differs between Omarchy 3.x and Lua-based Omarchy Quattro. Strata should detect whether it is already configured and offer to perform the setup automatically.

## Proposed solution

On the first eligible launch on Omarchy, show a one-time dialog asking whether the user wants to make Strata the default file manager.

### Prompt behavior

- Show the prompt only on supported Omarchy installations (3.x and Quattro).
- Before showing it, detect whether Strata is already both:
  - the XDG `inode/directory` handler; and
  - the target of Omarchy's file-manager shortcuts.
- Do not prompt if setup is already complete.
- Persist that the prompt was answered so declining does not cause it to reappear on every launch.
- Provide a Settings action/status so users can apply or repair the integration later without resetting first-run state.
- Explain that accepting changes both folder associations and Omarchy's file-manager shortcuts.

### Automatic setup

When accepted:

- Set `io.github.lgse.Strata.desktop` as the default handler for `inode/directory`, preferably through GIO rather than parsing MIME configuration directly.
- Update only the user's Hyprland configuration, never files under `/usr/share/omarchy/`:
  - **Omarchy 3.x:** handle `~/.config/hypr/bindings.conf`.
  - **Omarchy Quattro:** handle `~/.config/hypr/bindings.lua`.
- Replace/override both the normal file-manager shortcut and the “file manager at terminal cwd” shortcut where that generation supports them.
- Preserve unrelated custom bindings, create a backup before editing, and use an identifiable managed block so setup is idempotent and can be repaired or reverted.
- Reload Hyprland, check `hyprctl configerrors`, and report actionable errors. If validation fails, restore the backup rather than leaving broken configuration.
- Verify the MIME association and active bindings after setup before reporting success.

If the desktop entry is unavailable or the Omarchy generation/configuration cannot be identified safely, do not partially apply changes; explain what could not be automated and link to manual instructions.

## Alternatives considered

- Keep README-only instructions: safe, but many users will never discover that XDG associations and Omarchy shortcuts are separate.
- Set only the XDG association: application-launched folders would use Strata, while Omarchy's keyboard shortcut would continue opening Nautilus.
- Modify Omarchy's packaged defaults: those files are package-owned and would be overwritten by updates.
- Prompt on every launch until accepted: intrusive and contrary to a one-time onboarding prompt.

## Visual examples

N/A — use Strata's existing themed dialog patterns. Proposed actions: **Make Default** and **Not Now**.

## Strata version

Current development version (commit `9be0192`).

## Additional context

Automating the XDG association is feasible. The main implementation risk is safely detecting and editing two generations of user-owned Omarchy bindings without damaging customizations.

Current manual setup is documented in the README. Quattro uses `~/.config/hypr/bindings.lua`; earlier Omarchy 3.x releases use `~/.config/hypr/bindings.conf`.

spandan11106:
## Description

Installing Strata on Omarchy left two separate pieces of state pointing elsewhere: the XDG `inode/directory` association, and Omarchy's own file-manager shortcuts, which launch Nautilus directly and never consult that association. Setting one without the other gives inconsistent behaviour, and the manual steps differ between Omarchy 3.x and Quattro.

Strata now detects both on the first eligible launch and offers to do the setup. The same action lives under **Settings → General → Omarchy integration**, so the integration can be applied or repaired later without resetting first-run state. Declining is persisted, so the prompt does not come back.

Setup resolves the generation from the installed Omarchy version rather than from which files happen to exist, since an upgraded Quattro system still has `bindings.conf` on disk. Only the user's own bindings are edited — `~/.config/hypr/bindings.lua` on Quattro, `~/.config/hypr/bindings.conf` on 3.x — through a marked block that is replaced rather than duplicated on re-run, after taking a backup. `hyprctl configerrors` gates the result: if Hyprland rejects the new bindings, the previous file is restored and the failure is reported with a link to the manual instructions. Nothing under `/usr/share/omarchy/` is touched.

## Visual evidence

The first-run dialog, on Omarchy Quattro 4.0.0.alpha (Tokyo Night):

> **Make Strata your file manager?**
> Folder associations and Omarchy Quattro's file-manager shortcuts still point elsewhere.
>
> Strata will become the default application for folders and will take over Omarchy's file-manager shortcuts, which currently launch Nautilus. Only your own Hyprland configuration is edited, and it is backed up first.
>
> Shortcuts: ~/.config/hypr/bindings.lua
>
> `Not Now`  `Make Default`

Screenshot attached separately — it uses the existing themed modal (`modal_layout` + `modal_layer`) and semantic `@theme_*` colors throughout, so it follows theme changes live.

## How to test

Run against a throwaway home so nothing real is modified:

1. `mkdir -p /tmp/fakehome/.config/hypr && cp ~/.config/hypr/{hyprland.lua,bindings.lua} /tmp/fakehome/.config/hypr/`
2. `HOME=/tmp/fakehome XDG_CONFIG_HOME=/tmp/fakehome/.config XDG_DATA_HOME=~/.local/share ./target/debug/strata /tmp/fakehome`
3. Wait ~2s for the prompt, then press **Make Default**.
4. Inspect `/tmp/fakehome/.config/hypr/bindings.lua` and `/tmp/fakehome/.config/mimeapps.list`.
5. Restart the same command, then open **Settings → General**.

**Expected result:** step 4 shows a `strata:file-manager-bindings` block appended after the pre-existing bindings, which are unchanged, no `.strata.bak.*` file left behind, and `inode/directory=io.github.lgse.Strata.desktop` in `mimeapps.list`. Step 5 does not prompt again, and the **Omarchy integration** row reads "Strata is the Omarchy Quattro file manager." with a **Repair** button that is idempotent — pressing it does not add a second block.

On a non-Omarchy system, no prompt appears and the Settings row is absent.

## Related issue

Closes #299

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,9 +194,13 @@ When building from source, `make install-local` installs the binary, icon, and d
 
 ### Make Strata the Omarchy file manager
 
-The XDG association above handles folders opened by applications. On current Lua-based Omarchy releases, also override the stock Nautilus shortcuts in `~/.config/hypr/bindings.lua` so Omarchy launches Strata directly.
+The XDG association above handles folders opened by applications. Omarchy's own file-manager shortcuts launch Nautilus directly and never consult that association, so both have to be changed.
 
-First inspect the active bindings and back up your user configuration:
+On a supported Omarchy installation Strata offers to do this for you the first time it starts, and the same action is always available under **Settings → General → Omarchy integration**. Accepting sets the `inode/directory` handler and adds a marked block of file-manager bindings to your own Hyprland configuration — `~/.config/hypr/bindings.lua` on Omarchy Quattro, `~/.config/hypr/bindings.conf` on Omarchy 3.x. Your file is backed up first, the rest of it is left alone, and if `hyprctl configerrors` reports a problem the previous bindings are restored. Re-running the action repairs the block rather than adding a second copy; deleting the block restores Omarchy's defaults.
+
+Files under `/usr/share/omarchy/` are never touched, so these overrides survive Omarchy updates.
+
+To do it by hand instead, first inspect the active bindings and back up your user configuration:
 
 ```bash
 omarchy menu keybindings --print | grep -i "file manager"
@@ -228,7 +232,16 @@ hyprctl configerrors
 omarchy menu keybindings --print | grep -i "file manager"
 ```
 
-`hyprctl configerrors` should produce no errors. These user overrides survive Omarchy updates; do not edit files under `/usr/share/omarchy/`.
+`hyprctl configerrors` should produce no errors. Do not edit files under `/usr/share/omarchy/`; they are package-owned and are replaced on update.
+
+On Omarchy 3.x the same overrides go in `~/.config/hypr/bindings.conf` using Hyprland's own syntax:
+
+```ini
+unbind = SUPER SHIFT, F
+unbind = SUPER ALT SHIFT, F
+bindd = SUPER SHIFT, F, File manager, exec, uwsm-app -- strata
+bindd = SUPER ALT SHIFT, F, File manager (cwd), exec, uwsm-app -- strata "$(omarchy-cmd-terminal-cwd)"
+```
 
 ### Network shares
 
--- a/src/services/mod.rs
+++ b/src/services/mod.rs
@@ -2,6 +2,7 @@
 
 mod file_source;
 mod install_source;
+mod omarchy_setup;
 mod operations;
 mod preview;
 mod release_channel;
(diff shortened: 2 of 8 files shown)


spandan11106:
image

image

l0gicgate:
So I built the interactive installer instead. We can close this issue for now as it shouldn't be as widespread of a problem.

PR #314, opened 2026-09-04
fix(browser): hide Move to Trash when the location cannot support it

### LINKED ISSUE #284: Hide “Move to Trash” when trashing is unsupported
## Strata version

0.9.1 (source commit 62e00ba)

## Installation method

Built from source

Omarchy, Hyprland 0.56.2 (Wayland)

## What happened?

The item context menu offers **Move to Trash** for entries in `/tmp`, even though GIO does not support trashing entries on this system-internal mount. Choosing it starts an operation that cannot succeed and produces an avoidable error/fallback flow.

## Steps to reproduce

1. Open `/tmp` in Strata.
2. Select a file or directory and open its context menu.
3. Observe that **Move to Trash** is available.
4. Choose **Move to Trash**.
5. Observe that the trash operation is rejected as unsupported.

## Expected behavior

Do not show **Move to Trash** when the selected entries cannot be trashed, including entries under `/tmp` on system-internal mounts. Keep **Permanently delete** available as the valid destructive action.

Ideally this should be based on the location's trash capability rather than a path-only `/tmp` special case, so other mounts that do not support trash behave consistently.

## Screenshots or video

N/A — the behavior is described by the context-menu state and error below.

## Relevant logs

```shell
$ gio trash /tmp/example
 gio: file:///tmp/example: Trashing on system internal mounts is not supported
```

## Additional context

The operation was independently reproduced with `gio trash` on `/tmp`, which is mounted as `tmpfs` in this environment. Strata already exposes **Permanently delete** via Shift+Delete, so suppressing the unsupported trash action does not remove the valid deletion path.

pdwaldrop:
## Description

"Move to Trash" was always shown regardless of whether the current location actually supports trashing, so on a location that doesn't (e.g. a tmpfs mount, or certain removable/network filesystems), choosing it just fails.

This resolves it proactively instead of reactively: each directory load now runs a single `access::can-trash` GIO query against the directory itself, and the cached result is read synchronously when building the item context menu. An unresolved check (query times out, etc.) still defaults to showing the option — never hides the only delete path — and Trash itself always shows "Move to Trash" regardless of the check, since restoring from Trash is really a move back to the original location.

## Visual evidence

Right-click context menu on a file, comparing a trash-unsupported location (tmpfs) against a trash-supported one (btrfs): 
image

image

## How to test

1. Create a file on a location that doesn't support trash, e.g. `touch /tmp/no-trash-file.txt` (tmpfs generally doesn't support `access::can-trash`).
2. In Strata, navigate to that location and right-click the file.
3. Create a file on a normal trash-supported location, e.g. your home directory, and right-click it there too.

**Expected result:** the tmpfs file's context menu shows "Permanently delete" but not "Move to Trash"; the home-directory file's context menu shows both. Trash itself continues to always show "Move to Trash" (used for restoring items).

## Related issue

Closes #284

### DIFF
--- a/src/adapters/local_files.rs
+++ b/src/adapters/local_files.rs
@@ -46,6 +46,7 @@ enum NativeEnumeration {
         entries: Vec<FileEntry>,
         truncated: bool,
         metadata_complete: bool,
+        can_trash: Option<bool>,
     },
     Failed(String),
     Cancelled,
@@ -251,6 +252,16 @@ fn scan_native_directory(
         Ok(children) => children,
         Err(error) => return NativeEnumeration::Failed(error.to_string()),
     };
+    // A single stat-speed query on the directory itself, not per entry -- this is cheap
+    // enough to always run rather than threading a separate opt-out through the request.
+    let can_trash = gio::File::for_path(path)
+        .query_info(
+            gio::FILE_ATTRIBUTE_ACCESS_CAN_TRASH,
+            gio::FileQueryInfoFlags::NONE,
+            Some(cancellable),
+        )
+        .ok()
+        .map(|info| info.boolean(gio::FILE_ATTRIBUTE_ACCESS_CAN_TRASH));
     let hidden_names = native_hidden_names(path);
     let mut entries = Vec::new();
     let mut truncated = false;
@@ -320,6 +331,7 @@ fn scan_native_directory(
         entries,
         truncated,
         metadata_complete,
+        can_trash,
     }
 }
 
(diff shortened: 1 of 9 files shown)


l0gicgate:
Addressed the review findings in `b6f07af`:

- Trash capability is now derived from an actual listed child rather than the directory being viewed. This avoids false negatives for locations such as `$HOME`, where GIO can report the directory itself as non-trashable while its contents are trashable.
- Remote enumeration now requests `access::can-trash` with the existing child metadata and caches the first resolved value. This removes the separate serialized capability query, so it can no longer consume the directory enumeration deadline or produce an empty truncated listing on timeout.
- Updated the capability documentation and added a regression test that verifies the emitted value matches the child entry's GIO capability.

The full local formatting, Clippy, and test suite passes.

PR #321, opened 2026-09-04
feat(browser): add a Tree view mode and align view names with Finder

### LINKED ISSUE #315: feat(browser): add Tree view mode and align view names with Finder
## Problem or use case

The current view mode labels don't match what users expect from a file manager:

- **"List"** is actually Miller columns (click a folder → next column appears horizontally)
- **"Grid"** is actually an icon grid (thumbnails in a grid layout)
- **"Explorer"** is actually a list/table view (Name, Mode, Size, Type, Modified columns)

macOS Finder uses these names:
- **Icons** — files as icons in a grid
- **List** — files in a table with sortable columns
- **Columns** — Miller columns (horizontal drilling)

The naming mismatch makes it hard for users coming from Finder or other file managers to find the view they want.

Additionally, there is no **Tree view** — a single-pane vertical list with expandable/collapsible folders (disclosure triangles, indentation by depth). This is common in Windows Explorer, Nautilus (list view with expandable rows), Thunar, and `tree` output. It lets users see deep directory structures without horizontal scrolling, which Miller columns require.

## Proposed solution

1. **Rename the UI labels** to match Finder terminology:

   | Current label | What it does | New label |
   |---|---|---|
   | "List" | Miller columns | "Columns" |
   | "Grid" | Icon grid | "Icons" |
   | "Explorer" | Table with columns | "List" |

   The internal enum names (`BrowserMode::Columns`, `Grid`, `Explorer`) stay unchanged — only the user-facing strings in the view switcher menu change.

2. **Add a 4th view mode: Tree**

   A single-pane vertical list where:
   - Folders have disclosure triangles to expand/collapse
   - Children are indented by depth
   - Clicking a folder toggles expansion (or navigates into it, depending on click activation settings)
   - Arrow keys expand/collapse and navigate up/down the tree
   - Files show their icon + name, same as other modes
   - Sorting and hidden-file preferences apply per folder

   This would be a new `BrowserMode::Tree` variant with its own rendering path.

## Alternatives considered

- **Rename only, no tree view**: Fixes the naming confusion but doesn't add the tree navigation paradigm that some users prefer over Miller columns.
- **Tree in the sidebar instead of a view mode**: A sidebar tree is a different feature (navigation aid, not a full file view). This request is for a full view mode.
- **Rename the enums too**: Would touch many files for zero user-visible benefit. Not worth the churn.

## Visual examples

N/A — the existing views are already implemented; the tree view would follow the same visual language (icons, row styling, theme colors).

## Strata version

`main` at 58fc3a9

## Additional context

The tree view would complement the existing Miller columns: columns for horizontal drilling, tree for vertical overview of deep structures.

spandan11106:
## Description

The view switcher named the Miller columns "List", the icon grid "Grid" and the table "Explorer", none of which is what those views are called elsewhere. The labels are now **Columns**, **Icons** and **List**. The `BrowserMode` variants keep their names, so this is a user-facing rename only.

Adds a fourth presentation, **Tree**: one vertical pane whose folders expand in place with a disclosure chevron, indented by depth. Its root level mirrors the browser's active column, so selection, inline rename, the item context menu, cut styling and drag and drop behave exactly as they do in List. Rows inside an expanded branch belong to directories the navigation state does not own, so the view lists them through `Browser::list_directory` and orders them with `Browser::sorted_entries` — listing budgets, sort order and the hidden-file rule stay in the application layer. Every branch is dropped when the pane's column reloads, so changing the sort or toggling hidden files never leaves a stale listing behind.

Keyboard: <kbd>→</kbd>/<kbd>l</kbd> expands the focused folder (and activates it when there is nothing to expand), <kbd>←</kbd>/<kbd>h</kbd> collapses it, then walks to the parent row, then leaves the directory. <kbd>↑</kbd>/<kbd>↓</kbd> and <kbd>k</kbd>/<kbd>j</kbd> move through the flattened tree rather than skipping expanded children.

Known limits, both because a row below the root level has no position in the browser's active column: rows inside an expanded branch are not part of the multi-selection the rest of the app acts on, and right-clicking one shows the folder menu rather than the item menu. Clicking, expanding, opening and quick preview all work on them. Worth a follow-up once selection can address entries by location.

## Visual evidence

Tree mode, collapsed:

Tree view with folders collapsed

Tree mode, with `alpha` and `alpha/nested` expanded:

Tree view with nested folders expanded

## How to test

1. Build and run Strata over a directory with a few levels of nesting, for example `cargo run -- ~/Projects`.
2. Open the appearance menu (the toolbar's list icon) and confirm the modes now read **Columns**, **Icons**, **List**, **Tree**, with the check mark on the mode you are in.
3. Choose **Tree**, then click the chevron beside a folder, and the chevron beside a folder inside it.
4. Focus a row and press <kbd>→</kbd>, <kbd>←</kbd>, <kbd>↓</kbd> and <kbd>↑</kbd>.
5. Toggle hidden files with <kbd>Ctrl</kbd>+<kbd>H</kbd> and change the sort from the pane header, both while a branch is expanded.
6. Quit and reopen Strata.

**Expected result:** folders expand and collapse in place with their children indented one level further and sorted with folders first; arrow keys expand, collapse and move through the flattened tree; hidden files appear and disappear at every level, and a sort change reorders expanded branches too; Strata reopens in Tree mode. Switching to Columns, Icons and List still works as before.

## Related issue

Closes #315


spandan11106:
Sorry the above description does not have proper visuals. Here are some better one. 

The name changes for the views : 
image

Tree view does not support group by file types. 


https://github.com/user-attachments/assets/7d99184d-6bd5-4e55-b984-2cbe5f8a06e5

l0gicgate:
@spandan11106, keyboard navigation crashes the changes here. Please fix.

spandan11106:
Fixed in 4cc8bb1.

Root cause: `BrowserView::activate_focused` held a shared borrow of `mode_views` while the tree called `Browser::activate_in_place`, which emits synchronously. The observer then wanted a mutable borrow of the same cell, so it panicked with `RefCell already borrowed` inside a GTK callback — a `extern "C"` frame that cannot unwind, so the process aborted rather than unwound. It hit every activation in Tree mode: <kbd>Enter</kbd> always, and <kbd>→</kbd>/<kbd>l</kbd> whenever there was nothing to expand (a file, or an already-expanded folder).

```
ViewState::handle → mode_views.borrow_mut()   <- panicked here
Browser::emit ← Browser::select ← activate_in_place ← tree::activate_position
```

The tree now returns what activating a row should do (`TreeActivation`) instead of doing it, so the caller drops its borrow before the browser runs and emits back into the mode views. Mouse activation goes through the same path, so there is one behaviour to reason about.

Verified by driving `tree_step` (Next, Expand, Next, Previous, Collapse, Collapse) and `activate_focused` directly in a running instance in Tree mode: it aborted at the first activation before the change and completes the whole sequence after it. Added unit coverage for the activation intent — root rows resolve to a column position, rows inside an expanded branch to their location.

PR #327, opened 2026-09-04
fix(browser): align Explorer row hover and click hit targets

### LINKED ISSUE #251: fix(browser): align Explorer row hover and click hit targets
## Problem or use case

In Explorer (list) view, resting the pointer exactly on the border between two rows highlights the lower/next item, but a click at that same spot does **not** select it. Hover and click are using different hit geometry, so there is a thin dead zone where the UI promises one row and the click does nothing (or hits the wrong target).

## Steps to reproduce

1. Open a folder in Explorer / list view with several adjacent files (e.g. `AGENTS.md`, `CLAUDE.md`, `CONTRIBUTING.md`).
2. Move the pointer slowly along the horizontal divider between two rows until the lower row’s hover highlight appears.
3. Without moving the pointer, click.
4. Observe that the highlighted row is not selected (selection unchanged / click ignored).

## Expected behavior

Hover highlight and click selection must share the same row hit target:

- If a row shows hover, a click at that pointer position selects that row.
- Row borders have no gap where hover and click disagree.
- Applies to Explorer list rows at minimum; Grid / Miller should use the same rule if they share the widget.

## Proposed solution

Drive hover and button-press from one geometry (the row’s allocation / pick function), including separator/border pixels. Assign the boundary consistently to one adjacent row (or expand the clickable area to match the painted hover rect) so CSS/theme separators cannot create a hover-only strip.

## Alternatives considered

- Ignoring hover on separators: removes the lie, but still leaves a hard-to-click strip.
- Expanding only the hover rect: makes the mismatch worse.

## Visual examples

https://github.com/user-attachments/assets/7eb2e4e6-cabe-4841-8ac7-21886cf7bc51

## Strata version

0.9.0

## Additional context

Related but not duplicates: #169 (hover color tokens), #86 / #104 (selection contrast), #213 (click activation count), #84 (double-click activation).

wmfeht:
## Description

Explorer list rows inherited GTK theme padding on the `ListView` row node. That padding made the hover background paint onto the 1px strip between rows, while click selection still used the inner allocation, so hovering a divider could highlight the next row without selecting it.

This sets `.explorer-list row { padding: 0; }` so hover highlight and click selection share the same geometry.

## Visual evidence

Before: the recording attached to #251 shows the hover/click mismatch on the row divider.

After: the hover highlight and click target agree on every pixel of the row, including the border between adjacent files.

## How to test

1. Open a folder in Explorer / list view with several adjacent files (for example `AGENTS.md`, `CLAUDE.md`, `CONTRIBUTING.md`).
2. Move the pointer slowly along the horizontal divider between two rows until the lower row’s hover highlight appears.
3. Without moving the pointer, click.

**Expected result:** The highlighted row is selected. There is no divider strip where hover and click disagree.

## Related issue

Closes #251

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -3473,6 +3473,7 @@ menubutton.grid-thumbnail-menu > button:focus-visible {
   color: @theme_text;
   outline: none;
   transition: background 160ms ease-out, color 160ms ease-out;
+  padding: 0;
 }
 
 .explorer-list row:hover {


wmfeht:
https://github.com/user-attachments/assets/c16fa511-5097-4375-b1f0-838b98a81617

ogarza (approved):
Thanks, this works. Good catch!

PR #328, opened 2026-09-05
perf(browser): keep Grid scrolling and thumbnail resizing smooth

### LINKED ISSUE #322: perf(browser): make Grid scrollbar-thumb dragging as smooth as list views
### Problem or use case

Dragging the vertical scrollbar thumb through a very large Grid listing remains choppier than Explorer or Columns, even after the source-index and pane-lifetime improvements in #310. The difference is most visible with about 100,000 entries. Page Up/Down and ordinary scrolling are fast. This is not a regression.

### Proposed solution

Profile release builds while dragging the Grid scrollbar thumb, then reduce the remaining synchronous Grid card binding and GTK layout work. Consider coalescing work per frame, simplifying the active-scroll card path, or introducing a fixed-size virtual grid if GtkGridView random-access layout is the dominant cost.

### Alternatives considered

A scroll-settle prototype defers thumbnail and metadata work until 80 ms after adjustment changes. Manual testing showed a clear improvement and correct rendering, but Grid still did not reach Explorer/Columns smoothness. A custom virtual grid should be considered only if smaller binding/layout optimizations are insufficient.

### Visual examples

N/A — performance-only behavior.

### Strata version

0.9.1 source build, PR #310 (`884787a`) plus the uncommitted scroll-settle prototype.

### Additional context

Manual fixture: `scripts/generate-fixture.sh`, directory `target/fixtures/100000`. Tested in an optimized release build. The prototype kept item rendering correct and made thumb dragging much smoother, while the two list-based views remained smoother still.

ogarza:
## Description

Grid thumb-drags through large listings still rebuilt every visible card: mixed 26px fallbacks vs thumbnail-sized images, wrapping labels, and extra layout. Cards now keep a fixed icon slot and a two-line `GtkInscription` name. While the scrollbar thumb is moving, bind only updates the name; icons, tooltips, cut styling, and opacity wait for the existing 80 ms settle.

The thumbnail slider is also less jumpy: it scales the current texture in place instead of flashing a generic icon, drops GTK's long-press fine-tune so the thumb does not snap to 64 or 256, and moves 1px per wheel notch. Scrolling over the listing closes the popover and moves the grid; scrolling over the sidebar or other chrome just closes it.

## Visual evidence

https://github.com/user-attachments/assets/64163761-2b2b-4e13-9165-69af196abd61

## How to test

1. `./scripts/generate-fixture.sh target/fixtures` if needed, then `cargo build --release && target/release/strata target/fixtures/100000`.
2. Switch to Grid. Drag the vertical scrollbar thumb from top to bottom and back.
3. Repeat in Explorer on the same directory, and try Page Up/Down in Grid.
4. After the thumb stops, confirm names are correct and thumbnails appear.
5. Open Thumbnail size. Drag the slider slowly at both ends; it should ease off 64 and 256 without snapping.
6. With the popover still open, roll the wheel over the slider (1px per notch), then over the listing (popover closes, grid scrolls), then over the sidebar (popover closes, grid does not move).
7. Drag the slider across a folder of images and confirm existing thumbnails grow in place without flashing generic icons.

**Expected result:** Grid thumb-drag is close to Explorer/Columns. Names follow the thumb. Thumbnails appear after settle. Short names leave empty space in the two-line slot instead of shrinking the cell. The slider stays controllable at the ends, wheel-scrolls in 1px steps, dismisses when scrolling the listing or sidebar, only moves the grid when the pointer is over the listing, and resizes pictures without an icon flash.

## Related issue

Closes #322

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -107,6 +107,7 @@ pub mod icons {
 }
 
 const FONT_VERSION: &str = "2.304";
+const ICON_TEXTURE_PX: i32 = 96;
 const ICON_TEXTURE_CACHE_LIMIT: usize = 256;
 const JETBRAINS_MONO: &[u8] = include_bytes!("../data/fonts/JetBrainsMono[wght].ttf");
 
@@ -278,7 +279,7 @@ fn primary_icon_texture(name: &str, color: &str) -> Option<gdk::Texture> {
             1,
         );
     }
-    texture_from_svg(name, color, source)
+    texture_from_svg(name, color, svg_at_texture_size(source))
 }
 
 fn folder_decoration_texture(decoration: &str, color: &str) -> Option<gdk::Texture> {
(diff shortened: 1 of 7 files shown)


ogarza:
I noticed that dragging on the thumbnail size slider without letting go of the mouse occasionally displays the fallback icon... I don't like that

ogarza:
using scrollwheel on the slider causes it to jump almost the whole slider in 1 'click'

ogarza:
If the thumbnail slider is on the page and the user moves the mouse and tries to use the scroll-wheel, nothing happens until you click to dismiss the slider

PR #333, opened 2026-09-05
feat(transfer): default cross-volume drops to copy

### LINKED ISSUE #248: feat(transfer): default cross-volume drops to copy, with modifiers for move
## Problem or use case

Dragging a file or folder onto a different physical volume or network location currently behaves like a same-volume move: the drop relocates the item and removes it from the source. Across devices that is surprising and easy to get wrong. Users often expect a cross-volume drop to **copy** (leave the original) or at least to **confirm** before the source is deleted.

Same-volume drops can stay as move. Cross-volume and remote drops need a safer default, with modifier keys to override.

## Expected behavior

Detect whether the drop destination is on the same filesystem / volume as the source (including local mounts and network / GVfs locations). Then:

| Gesture | Same volume | Different volume / network |
|---|---|---|
| Plain drop | Move (current) | **Copy** (preferred), or prompt Copy / Move / Cancel |
| Modifier force-copy (e.g. `Ctrl`) | Copy | Copy |
| Modifier force-move (e.g. `Shift`) | Move | Move (copy-then-delete only after success) |
| Modifier force-link if supported (e.g. `Alt`) | Symlink / optional | Disabled or no-op with a clear reason |

Also:

- Cursor / drop badge should show the pending action (copy vs move) before release.
- A forced cross-volume move must not delete the source until the destination copy fully succeeds; cancel / failure leaves the source intact (same integrity bar as #150 / #7).
- Local↔remote and remote↔remote drops follow the same policy once the backends support the action (#66).

## Proposed solution

1. Classify each drag as same-volume vs cross-volume (and local vs remote) before committing the drop.
2. Default plain cross-volume drops to **copy**.
3. Honor modifier keys to force copy or move; document them in the keyboard shortcuts / help UI.
4. Optionally offer a one-shot confirmation dialog the first time a cross-volume **move** is requested (settings toggle later if needed).
5. Reuse the existing transfer pipeline for progress, conflicts, and cancellation.

## Alternatives considered

- Always move, even across volumes: matches some older file managers, but silently removes the only copy from a USB / network share.
- Always prompt: safest, but noisy for routine USB copies; modifiers cover power users if the default is copy.
- Same-volume copy by default: breaks the expected rearrange-within-folder workflow.

## Visual examples

N/A — observed when dragging between home and another mounted volume / network location: plain drop relocates instead of copying.

## Additional context

Related but not duplicates:

- #150 — cross-filesystem directory *move* currently fails (`Can't recursively copy directory`); this issue is about the **default action** for DnD, not making move work.
- #66 — capability-gate remote copy/move (including drag/drop).
- #7 — failure-safe replacement / staging.
- #239 — Miller same-folder duplicate and preview-as-drag-source.

wmfeht:
## Description

Dragging within Strata onto another volume currently relocates the source because drop handling treats any same-process drag as a move. This change classifies source and destination volumes, then defaults a plain cross-volume or unknown-volume drop to copy.

Same-volume drops still move. Ctrl forces copy, Shift forces move (using the existing copy-then-delete path). Cut+Paste is unchanged and still moves. Destination directory symlinks are followed so a `~/usb` link onto a stick copies instead of deleting the original.

## Visual evidence

N/A — the visible change is the compositor copy vs move cursor and whether the source is removed, not layout or theming.

## How to test

1. Drag a file within `$HOME` onto another folder on the same volume.
2. Drag a file from `$HOME` onto a USB volume, `/tmp` if it is tmpfs, or another mounted disk.
3. Repeat (1) while holding Ctrl, and (2) while holding Shift.
4. Drag onto a directory symlink that points at another device.
5. Hover an empty mount-point folder, mount a USB there, and drop without leaving the row.

**Expected result:**

1. Move (source gone).
2. Copy (source remains).
3. Ctrl copies on the same volume; Shift moves onto the other volume only after the copy succeeds.
4. Copy (source remains).
5. Copy, not move. The cursor should switch to copy after the mount without needing a pointer wiggle.

Also confirm Cut then Paste across volumes still moves, and a cancelled Shift-move of a large directory leaves the source intact.

On Hyprland/Omarchy, check whether the cursor distinguishes copy vs move. If it does not, the default is still copy; a badge follow-up should be filed rather than treating the cue as shipped.

## Related issue

Closes #248

### DIFF
--- a/src/adapters/mod.rs
+++ b/src/adapters/mod.rs
@@ -3,8 +3,13 @@
 mod local_files;
 mod local_operations;
 mod local_preview;
+mod volume;
 
 pub use local_files::LocalFileSource;
 pub(crate) use local_files::location_for_file;
 pub use local_operations::LocalOperationProvider;
 pub use local_preview::LocalPreviewProvider;
+pub(crate) use volume::{
+    cached_volume_identity, flush_volume_identity_cache, prefetch_volume_identity,
+    query_volume_identity, register_volume_restatus,
+};
--- a/src/adapters/volume.rs
+++ b/src/adapters/volume.rs
@@ -1,0 +1,276 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#[cfg(test)]
+mod tests;
+
+use std::{
+    cell::RefCell,
+    collections::{HashMap, VecDeque},
+    os::unix::fs::MetadataExt,
+    rc::Rc,
+};
+
+use gtk::{gio, glib, prelude::*};
+
+use crate::{model::Location, services::VolumeIdentity};
+
+const LRU_LIMIT: usize = 64;
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+struct CacheKey {
+    location: Location,
+    follow_symlinks: bool,
+}
+
+struct RestatusHook {
+    run: Rc<dyn Fn() -> bool>,
+}
+
+struct VolumeCache {
+    epoch: u64,
+    entries: HashMap<CacheKey, VolumeIdentity>,
+    order: VecDeque<CacheKey>,
+    inflight: HashMap<CacheKey, u64>,
+    restatus: Vec<RestatusHook>,
+}
+
+impl VolumeCache {
+    fn new() -> Self {
+        Self {
+            epoch: 0,
+            entries: HashMap::new(),
+            order: VecDeque::new(),
+            inflight: HashMap::new(),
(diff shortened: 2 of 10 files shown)


l0gicgate:
Thanks for the PR @wmfeht I will review this shortly!

wmfeht:
Oop this isn't ready sorry, agent went off the rails and pushed

l0gicgate:
Tag me when it's ready @wmfeht 🙏🏼

PR #338, opened 2026-09-05
feat(transfer): default cross-volume drops to copy

wmfeht:
## Description

This evaluates whether the target of a drag and drop operation is on a separate volume and does a copy instead of a move for different volumes. It should also setup infrastructure for other future enhancements like shift-drag to copy (not implemented, as it would require rewriting list view selection). This should be similar to how Dolphin or Nautilus etc. implement the same functionality.

## Visual evidence

https://github.com/user-attachments/assets/eea39c32-888b-42a7-a1b2-2ab00df00ee2

## How to test

1. Drag a file from one folder to another folder on the same disk and drop it.
2. Drag a file from the local disk onto a folder on a USB stick or a mounted SMB/SFTP share and drop it.

**Expected result:** Step 1 moves the file. Step 2 copies it, leaving the source in place.

## Related issue
Partially addresses #248 - recommend splitting shift/ctrl-drag operations into a separate issue as there is a limitation with the list view that prevents dragging items with modifier keys without reimplementing selection.


wmfeht:
Local automated testing on `0e01cff` (`feat/248-cross-volume-copy-default`) passed.

- `./scripts/check.sh`: `cargo fmt --all --check`, `cargo check --all-targets --all-features`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --all-targets --all-features` (display variables cleared), `cargo deny check`, and `typos` all succeeded. Cargo tests: 903 passed in the library crate, plus the remaining `--all-targets` binaries/integration tests; 0 failed. 18 tests were ignored by design (GTK/display/D-Bus cases that `check.sh` skips).
- `./scripts/e2e.sh -p no:cacheprovider`: 189 passed (harness unit tests and GUI scenarios, including visual baselines) in 3m45s.

No failures.

PR #350, opened 2026-09-05
fix(drag-drop): accept sidebar drops and compact explorer previews

### LINKED ISSUE #285: Fix sidebar file drops and explorer drag previews
## Strata version

Current development version (commit `62e00ba`).

## Installation method

Not provided in the original report.

Linux; exact distribution and desktop environment were not provided by the reporter.

## What happened?

File drag and drop has two related usability problems:

1. Files cannot be moved by dragging them onto folder destinations in the sidebar.
2. In Explorer mode, the drag preview captures the entire selected row, including its metadata columns, rather than showing a compact representation of the dragged file or selection.

The first problem prevents a common file-management workflow. The second produces an oversized, awkward drag preview.

### Sidebar drop target

1. Open a directory containing a file.
2. Drag the file onto a folder or pinned location in the sidebar.
3. Observe that the sidebar destination does not accept the drop and the file is not moved.

### Explorer drag preview

1. Switch to Explorer mode.
2. Select a file and begin dragging it.
3. Observe that the drag preview contains the full Explorer row and all visible metadata columns.

## Expected behavior

- Valid folder destinations in the sidebar should accept file drops and perform the appropriate move or copy operation using the same semantics as directory drop targets in the browser.
- Explorer mode should use a compact drag preview representing the dragged file or multi-file selection, not a snapshot of the entire metadata row.

## Screenshots or video

The original report was provided via a chat screenshot describing both behaviors; no application recording is available.

## Relevant logs

No logs were provided.

## Additional context

The current Explorer drag source creates its icon from the source widget, which appears to explain why the whole row is captured. Sidebar drag-and-drop currently supports reordering pinned locations, but file-list drops also need to be recognized without interfering with pinned-item reordering.

Hank-dev:
## Description

File drags from Grid and Explorer could be claimed by the sidebar's pinned-place reorder target before the `FileList` target saw them. Local sidebar rows now prioritize file payloads while retaining string payloads for place reordering, so Home, standard places, pinned folders, and mounted native devices accept files from every browser mode.

Explorer drag previews now render only the filename instead of the full metadata row. Grid previews remain unchanged. Trash, Network, and other URI locations remain non-droppable.

## Visual evidence

Not captured — the review environment does not provide pointer-injection tooling for holding and moving a GTK drag. This changes only the transient Explorer drag preview and drop highlight.

## How to test

1. Open `~/Documents` with a file and pin a local `demo` folder.
2. In Columns, Grid, and Explorer modes, drag the file onto `Pinned/demo`.
3. In Explorer mode, inspect the preview while dragging.
4. Try dropping onto Trash or Network.

**Expected result:**

`Pinned/demo` outlines as a drop target in every mode and runs the normal move/copy transfer. Explorer drags show only the filename. Trash and Network do not highlight as drop targets.

## Related issue

Closes #285


l0gicgate (dismissed):
I am reworking this PR, because it doesn't work in Grid or Explorer mode. Only miller column mode works

l0gicgate:
Follow-up review fixes pushed after reproducer feedback:

- File-list drops now run in capture phase, and reorder targets explicitly reject file-list payloads. This prevents Grid/Explorer drags from being claimed as pinned-place reorder data before the sidebar file target can highlight and transfer them.
- Explorer drag icons now paint only the filename label rather than the full metadata row.

The complete local fmt, clippy, and test suite passes (771 passed, 18 display-dependent tests ignored by the default suite). I could not perform or record pointer-driven drag verification because this environment has no usable pointer-injection tooling; no screenshot/video was uploaded.

l0gicgate (dismissed):
Reviewed the follow-up fixes and updated diff. File payloads are separated from reorder payloads, and Explorer now uses the filename widget for its drag icon. Full local checks pass.

l0gicgate (approved):
Reviewed final follow-up including the sidebar payload regression test. Full local checks pass.

PR #352, opened 2026-09-05
fix(remote): compare remote locations hierarchically in is_within

### LINKED ISSUE #351: fix(remote): navigating away or refreshing after unmount does nothing for remote locations
### Strata version

main (commit 65e2e65 / v0.10.0)

### Installation method

Built from source

Arch Linux, Hyprland

### What happened?

After ejecting or disconnecting a mounted remote location (SFTP tested, but the same root cause affects SMB and any other URI-backed mount) while it's the active browsing location, Strata does nothing: the Miller columns keep showing the now-stale directory contents instead of navigating back to Home.

Root cause: `Location::is_within` (`src/model/mod.rs`) only compares native filesystem paths:

```rust
pub fn is_within(&self, other: &Self) -> bool {
    self.native_path()
        .zip(other.native_path())
        .is_some_and(|(path, parent)| path.starts_with(parent))
}
```

For a `Location::Uri` (SMB/SFTP/FTP/WebDAV), `native_path()` always returns `None`, so this always returns `false`. That breaks `navigate_home_if_within` in `src/ui/window.rs`, which uses it to decide whether the active browser location should return Home after an eject/unmount succeeds, and also affects the `DirectoryChange::Remove` handling in `src/app/navigation.rs` that truncates stale Miller columns.

### Steps to reproduce

1. Connect to a remote location (e.g. `sftp://user@host/path`).
2. Browse a few levels deep into it.
3. In the sidebar, eject/disconnect that device.
4. Observe the browser view.

### Expected behavior

The browser should navigate back to Home (matching what already happens for a local removable drive), since the active location no longer exists.

### Screenshots or video

N/A, behavioral bug not visually distinct beyond "nothing happens."

### Additional context

Fix: use `gio::File::has_prefix` (already used elsewhere in this file) to compare URI locations hierarchically instead of relying on native paths alone. A PR with the fix and unit tests follows.

mrocha-montes:
## Description

`Location::is_within` only compared native filesystem paths, so it always returned `false` for URI locations (SMB, SFTP, FTP, WebDAV). This broke `navigate_home_if_within` in `src/ui/window.rs` after ejecting or disconnecting a remote mount that was actively being browsed (the Miller columns kept showing stale content instead of returning Home), and also affects the `DirectoryChange::Remove` handling in `src/app/navigation.rs` that truncates stale Miller columns.

Fix: use `gio::File::has_prefix` (already used elsewhere in this file) to compare URI locations hierarchically, falling back to the existing native-path comparison when both locations are native.

## How to test

1. Connect to a remote location (SMB or SFTP).
2. Browse a few levels deep into it.
3. In the sidebar, eject/disconnect that device.

**Expected result:** the browser navigates back to Home, matching what already happens when a local removable drive is ejected while active.

I verified this exact flow end to end against a real SFTP host, both before the fix (view stayed stale after eject) and after (view returns Home).

## Related issue

Closes #351

### DIFF
--- a/src/app/navigation/tests.rs
+++ b/src/app/navigation/tests.rs
@@ -288,6 +288,21 @@ fn external_removals_close_affected_descendant_columns() {
 }
 
 #[test]
+fn remote_external_removals_close_the_exact_open_column() {
+    let root = Location::uri("sftp://user@host/mnt/share");
+    let removed = Location::uri("sftp://user@host/mnt/share/removed");
+    let mut state = NavigationState::default();
+    state.navigate(root.clone(), RequestId(1));
+    assert!(state.descend(0, removed.clone(), RequestId(2)));
+
+    let path = state
+        .path_after_external_change(0, &DirectoryChange::Remove(removed))
+        .expect("the removed remote column should be closed");
+
+    assert_eq!(path.locations(), &[root]);
+}
+
+#[test]
 fn selecting_a_sibling_replaces_deeper_columns() {
     let mut state = NavigationState::default();
     state.navigate(location("/home"), RequestId(1));
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -120,7 +120,9 @@ impl Location {
         let (Some(uri), Some(parent_uri)) = (self.uri_value(), other.uri_value()) else {
             return false;
         };
-        gio::File::for_uri(uri).has_prefix(&gio::File::for_uri(parent_uri))
+        let file = gio::File::for_uri(uri);
+        let parent = gio::File::for_uri(parent_uri);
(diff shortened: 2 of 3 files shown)


l0gicgate:
I pushed `dda3346` to address the review finding.

`gio::File::has_prefix` excludes equality, unlike the native `Path::starts_with` behavior used by `Location::is_within`. The URI branch now checks `file.equal(&parent) || file.has_prefix(&parent)`, so exact remote locations and descendants are both treated as contained.

I also added regression coverage for remote self-containment and for removing the exact open remote column. The full local suite passes: formatting, clippy with warnings denied, and all-target/all-feature tests (705 passed, 1 ignored).

Issue #66, opened 2026-08-31
feat(remote-ops): capability-gate operations and harden recovery


l0gicgate:
## Summary

Capability-gate remote file operations and harden cancellation, refresh, and disconnect recovery across GIO/GVfs backends.

The current GIO operation provider can perform many operations on remote URIs, and #31 added explicit refreshes for several no-monitor cases. However, the UI generally presents mutation actions without querying backend capabilities, unsupported remote Trash fails only after invocation, and safe replacement currently requires a local destination.

## Scope

- Query the relevant GIO attributes/capabilities before presenting or invoking:
  - create file/folder;
  - rename;
  - write/upload;
  - copy and move;
  - duplicate;
  - permanent delete;
  - Trash.
- Apply capabilities consistently to context menus, keyboard shortcuts, toolbar/background actions, clipboard paste, and drag/drop.
- Hide or disable unsupported actions with a useful explanation.
- Do not offer remote Trash unless the backend explicitly reports support.
- Never silently substitute permanent deletion for Trash.
- Verify local-to-remote, remote-to-local, and remote-to-remote copy/move behavior.
- Define safe collision/replacement behavior for remote destinations; the current replacement staging path intentionally rejects non-local destinations.
- Refresh every affected remote parent when live monitoring is unavailable.
- Preserve URI-native destinations and never route operations through GVfs FUSE paths.
- Harden cancellation and stale-result suppression during slow enumeration and mutation.
- On disconnect, cancel stale work and leave a retryable unavailable state.
- Provide explicit unavailable/fallback behavior for remote recursive search, thumbnails, and rich previews rather than relying on a FUSE path or unbounded download.
- Keep opening remote files through the system default application supported.

## Acceptance criteria

- [ ] Unsupported operations cannot be invoked through menus, shortcuts, clipboard, or drag/drop.
- [ ] Supported create, rename, duplicate, copy, move, upload/download, and permanent-delete operations work through GIO.
- [ ] Remote Trash appears only where explicitly supported.
- [ ] Choosing Trash never silently performs permanent deletion.
- [ ] Collision handling is safe and produces a clear result on backends that cannot provide safe replacement semantics.
- [ ] Local↔remote and remote↔remote transfers refresh all affected views.
- [ ] Cancelling slow enumeration or mutation leaves a consistent, refreshable view and ignores stale completion events.
- [ ] Disconnect during browsing or mutation produces a recoverable state rather than resetting navigation or crashing.
[...]

PR #361, opened 2026-09-05
feat(operations): capability-gate the permanent delete action

mrocha-montes:
## Description

First slice of #66. "Permanently delete" was always offered in the context menu regardless of backend support, unlike "Move to Trash" which already hides itself when `access::can-trash` reports false (#179/#225).

This extends the same query and plumbing (`DirectoryEvent::Finished`, per-column state, `can_trash_at`) to `access::can-delete`, adding a parallel `can_delete`/`can_delete_at`, and hides "Permanently delete" the same way "Move to Trash" already is, except while already browsing Trash, where "Move to Trash" serves as permanent delete under a shared label.

## Scope

In scope: capability-gating for the "Permanently delete" context-menu action specifically.

Out of scope, left for follow-up PRs against #66: gating create, rename, copy/move, and duplicate; applying gating to keyboard shortcuts, toolbar actions, clipboard paste, and drag/drop (today only the context menu is gated, matching how "Move to Trash" itself was first introduced).

## How to test

1. Create a directory and remove write permission from it (e.g. `chmod 555 some-dir`), so `access::can-delete` and `access::can-trash` both report false for its contents.
2. Right-click a file inside it in Strata.

**Expected result:** neither "Move to Trash" nor "Permanently delete" appear in the context menu.

I verified this exact scenario locally (both options correctly disappear together), plus the existing automated suite (4 new unit tests covering the visibility function directly).

## Related issue

Refs #66

### DIFF
--- a/src/adapters/local_files.rs
+++ b/src/adapters/local_files.rs
@@ -23,8 +23,8 @@ use crate::{
     },
 };
 
-const LIST_ATTRIBUTES: &str = "standard::display-name,standard::name,standard::type,standard::is-hidden,standard::is-symlink,access::can-trash";
-const FULL_ATTRIBUTES: &str = "standard::display-name,standard::name,standard::type,standard::is-hidden,standard::is-symlink,standard::size,time::modified,unix::mode,access::can-trash";
+const LIST_ATTRIBUTES: &str = "standard::display-name,standard::name,standard::type,standard::is-hidden,standard::is-symlink,access::can-trash,access::can-delete";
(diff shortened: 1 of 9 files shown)


l0gicgate:
Review follow-up:

- Added regression coverage that verifies directory enumeration resolves `access::can-delete` from a child entry.
- Added coverage that reload clears the previously resolved delete capability, preventing stale context-menu decisions during refresh.
- Trimmed redundant comments while retaining the non-obvious Trash/permanent-delete rationale.

Verification: `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --all-targets --all-features` all pass locally (715 tests passed, 1 ignored). These review-only changes do not alter UI behavior, so no new screenshot was needed; the PR description retains the original manual read-only-directory verification steps and result.

Issue #291, opened 2026-09-04
feat(keyboard): close keyboard-first workflow gaps


l0gicgate:
## Problem or use case

Feedback from a keyboard-focused Yazi user is that Strata currently feels mouse-first despite being positioned as keyboard-first. Their first-use observations were:

- Strata felt slow on lower-powered hardware compared with Yazi.
- Pressing `/` did not start a fuzzy search scoped to the current directory.
- Keyboard focus and the active selection were not immediately obvious, and focus sometimes appeared to return to the sidebar/quick-bookmark area.
- `hjkl` navigation felt like arrow-key substitution rather than a complete keyboard workflow.
- Common file-manager actions and jumps were not available through familiar direct or leader-key sequences. Examples they expected included view/preview, paste, delete, top/bottom navigation, create/add, and open.
- Features such as filtering and view controls still felt as though they required mouse interaction or hard-to-discover shortcuts.
- They could not tell whether keybindings could be customized through Strata's configuration.

The result was that they felt they needed to enable their trackpad to use Strata effectively.

## Proposed solution

Audit and polish Strata's end-to-end keyboard-only workflow, then split implementation into focused follow-up issues where appropriate. The audit should cover:

- A persistent, unmistakable visual indication of which pane, sidebar item, or file owns keyboard focus.
- Stable focus behavior that does not unexpectedly move back to the sidebar.
- A discoverable current-directory fuzzy search/filter action, with `/` considered as a default binding.
- Keyboard access to common file actions, view/preview controls, filtering, creation, opening, deletion, paste, and first/last-item navigation.
- Whether a leader-key model or direct single-key bindings provide the best workflow without breaking text entry.
- User-configurable keybindings with conflict detection, plus an accurate shortcut reference in Settings.
- Profiling keyboard navigation, directory loading, and mode transitions on lower-powered hardware so interaction remains responsive.

Exact default bindings should be decided as a coherent keymap rather than copying Yazi one key at a time. Text fields and other editing contexts must continue to receive ordinary typing.

## Alternatives considered

- Keep the current fixed shortcuts: avoids configuration complexity, but does not address discoverability or differing keyboard workflows.
- Copy Yazi's complete keymap: familiar to Yazi users, but may conflict with Strata's GUI conventions and existing shortcuts.
- Add isolated shortcuts without a broader audit: helps individual actions, but risks an inconsistent and conflicting keymap.

## Visual examples

N/A — the original feedback was supplied as a text report rather than an application screenshot.

[...]

PR #363, opened 2026-09-05
fix(browser): resolve delete targets from the focused pane

### LINKED ISSUE #300: fix(delete): honor the focused list pane when resolving deletion targets
## Strata version

0.9.1 development build (commit `2e0931b`).

## Installation method

Built from source.

Omarchy, Hyprland, Wayland.

## What happened?

In a single-pane list view, entering a folder and pressing `Delete` when no child item is selected targets the folder that was just entered. Confirming the operation therefore deletes the directory currently being viewed from its parent rather than acting on an explicitly selected item in the focused list.

This is especially surprising when keyboard focus is inside the opened folder while an older mouse/parent state still identifies the folder used to enter it. Destructive keyboard actions should follow keyboard focus and explicit selection, not pointer hover or an inferred parent entry.

## Steps to reproduce

1. Switch to a single-pane list view (Explorer or Grid).
2. Open a child folder from its parent directory.
3. Ensure no item inside the child folder is selected and keyboard focus is in the child list.
4. Press `Delete`.
5. Observe that the delete confirmation targets the folder currently being viewed (the entry from the parent directory).
6. Confirm the operation and observe that the opened folder is moved to Trash.

## Expected behavior

- `Delete` should resolve its target from the pane that owns keyboard focus.
- In Explorer/Grid mode, only explicitly selected entries in that focused pane should be eligible for deletion.
- If the focused pane has no selected entry, `Delete` should do nothing rather than falling back to the folder opened from the parent.
- Mouse hover or stale selection in another pane must not override keyboard focus for a destructive action.
- Columns-mode behavior should be reviewed separately and preserved only where the parent target is explicit and unambiguous.

## Screenshots or video

N/A — the incorrect target is visible in the delete confirmation dialog.

## Relevant logs

No relevant logs.

## Additional context

The behavior is confirmed in the current implementation. `Browser::deletion_entries()` deliberately falls back to `active_child_position()` in the parent depth when the active depth has no selection. The test `deletion_targets_the_entered_folder_when_the_child_has_no_selection` currently codifies that behavior.

The fix should add regression coverage at the UI/mode boundary so target resolution accounts for the focused view instead of applying the model fallback indiscriminately. Related: #236 established that pane-scoped actions should honor the focused pane.

pdwaldrop:
## Description

In Columns mode, `confirm_delete` picked its target column from the model's `active_column` bookkeeping instead of the pane that actually holds GTK keyboard focus. Those two can diverge, and Columns already leaves a breadcrumb-style highlight on the folder row in every ancestor column you drilled through, so several columns can look "selected" at once with no clear signal of which one Delete will act on.

`confirm_delete` now resolves the target depth from real keyboard focus first (`focused_column_depth()`), and only falls back to the model's active depth when nothing is focused (single-pane Explorer/Grid). The existing behavior of falling back to the navigated-into folder when the focused pane has no explicit selection is unchanged, just anchored to the real focused depth instead of the model's.

Per @l0gicgate's note on the issue, the `#291` branch will separately add a guard so an *empty* focused column doesn't fall back to its parent at all — that's out of scope here.

## Visual evidence

N/A — this changes which pane a keyboard shortcut resolves against internally; the dialog that appears looks identical either way. The behavior difference only shows up when the model's bookkeeping and actual keyboard focus disagree, which isn't something a screenshot can distinguish from the correct case.

## How to test

1. Open a folder with at least two subfolders in Columns view.
2. Use the arrow keys to descend into a subfolder (`Right`), then move focus back to the parent column (`Left`) without closing the child column.
3. Press `Delete`.

**Expected result:** the confirmation dialog names the row that has actual keyboard focus in the parent column, not whatever the child column happens to show highlighted.

The regression this guards against is covered directly by two new tests that force the model's active depth and the caller-supplied focused depth to disagree:
- `deletion_entries_prefer_an_explicitly_focused_pane_over_the_model_active_depth` (`src/app/browser/tests.rs`)
- `delete_prefers_the_focused_pane_then_falls_back_to_the_active_depth` (`src/ui/browser/tests.rs`)

## Related issue

Closes #300

### DIFF
--- a/src/app/browser.rs
+++ b/src/app/browser.rs
@@ -1087,14 +1087,24 @@ impl Browser {
         self.state.borrow().selected_entries()
     }
 
-    pub fn deletion_entries(&self) -> Vec<FileEntry> {
+    /// `depth` should be the pane that actually holds keyboard focus, when the caller
+    /// knows it; the model's own `active_depth` can lag behind it (e.g. a stale
+    /// selection left over in another column), which would otherwise point a
+    /// destructive action away from where the user is actually looking.
+    pub fn deletion_entries(&self, depth: Option<usize>) -> Vec<FileEntry> {
         let state = self.state.borrow();
-        let selected = state.selected_entries();
-        if !selected.is_empty() {
-            return selected;
+        let depth = depth.or_else(|| state.active_depth());
+        if let Some(depth) = depth {
+            let positions = state.selected_positions(depth);
+            if !positions.is_empty() {
+                return positions
+                    .into_iter()
+                    .filter_map(|position| state.entry_at(depth, position))
+                    .collect();
+            }
         }
 
-        let Some(parent_depth) = state.active_depth().and_then(|depth| depth.checked_sub(1)) else {
+        let Some(parent_depth) = depth.and_then(|depth| depth.checked_sub(1)) else {
             return Vec::new();
         };
         let Some(position) = state.active_child_position(parent_depth) else {
--- a/src/app/browser/tests.rs
+++ b/src/app/browser/tests.rs
@@ -306,6 +306,39 @@ impl FileSource for FakeFileSource {
     }
 }
 
+/// Unlike `FakeFileSource`, returns a "leaf" entry whose location is derived from the
+/// requested directory, so tests can tell which column's entries they actually got back.
+struct DepthAwareFileSource;
+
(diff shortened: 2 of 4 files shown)


l0gicgate:
Superseded by #358, which fixes the focused-column deletion targeting and also prevents deletion from an empty focused Columns pane.

PR #365, opened 2026-09-05
fix(operations): harden local move against symlink and TOCTOU races

pdwaldrop:
## Description

The last piece of #8's acceptance criteria: move's primary path called GIO's `move_async` directly -- plain path-based, only guarding the final path component from being a symlink. Every ordinary move (cut/paste, drag/drop, undo, restore from trash) went through this, unlike copy (#254) and delete (#253), which already route through a descriptor-relative walk.

Moves between two native local paths now resolve with a single `renameat2`, opening each side's parent directory fresh (walking from the filesystem root with `RESOLVE_BENEATH`/`RESOLVE_NO_SYMLINKS`) right before the call, so a symlink or path-component swap can't redirect it, and `NOREPLACE` fails closed instead of silently overwriting whatever a race put at the destination name. A rename the kernel can't do atomically -- into the source's own subtree, or across filesystems -- reports the same `WouldRecurse` signal GIO's move would, so the existing copy-then-delete fallback (added in an earlier PR) still handles it unchanged.

While building this I found two related gaps in the same area and folded them in rather than opening two more PRs for the same acceptance criteria:

- The parent-directory helper this reuses (previously delete-only, `open_local_delete_parent`) is now shared and renamed `open_local_parent_directory`. Copy's own parent-open was weaker than delete's -- a single `O_NOFOLLOW` open that only protects the immediate parent's own name, not an ancestor further up the path. `copy_rejects_a_symlink_higher_in_the_sources_parent_path` demonstrates a symlink swap one level higher than the immediate parent was not caught before this change.
- `replace_local`'s single-file staging copy now goes through `copy_recursively` like the directory case already did, instead of a separate raw GIO copy call. This is mostly a consistency cleanup -- `replace_local_with` already re-validates the source's identity (via the same hardened helper) before either branch runs, so a static-symlink test can't distinguish old from new behavior here. The real effect is narrowing the window between that upfront check and the actual read. I've been explicit about this in the test's doc comment rather than overselling it as newly-demonstrated coverage.

## Visual evidence

N/A -- this is internal file-operation plumbing with no UI surface. Verified manually in a running instance instead: cut/paste of both a file and a directory between two real folders, confirming correct content, an empty source afterward, and (via `ls`) that mtimes were preserved (a real rename, not a copy).

## How to test

Automated coverage (all in `src/adapters/local_operations/tests.rs`):
- `a_plain_move_relocates_the_entry_via_the_hardened_rename_path` -- ordinary move via the new fast path
- `moving_a_directory_into_its_own_child_fails_instead_of_deleting_it` -- the kernel's `EINVAL` correctly triggers the fallback, and the (equally impossible) fallback fails without deleting the original
- `move_rejects_a_symlink_in_the_destinations_parent_path` -- new protection; fails against the pre-fix implementation
- `move_rejects_a_symlink_in_the_sources_parent_path` -- regression coverage for the pre-existing source-identity check (see its doc comment)
- `copy_rejects_a_symlink_higher_in_the_sources_parent_path` -- new protection; fails against the pre-fix implementation
- `replace_rejects_a_symlink_in_the_sources_parent_path` -- regression coverage (see its doc comment)

Manual:
1. Create two folders, each with a file and a subfolder with content in it.
2. Cut the file from one and paste it into the other; then cut the subfolder and paste it too.

**Expected result:** both land in the destination with their original content and mtimes; the source is left empty. No fallback dialog or error.

## Related issue

Part of #8 -- I believe this closes out the acceptance criteria across copy, delete, move, and replace, but I'll let @l0gicgate make that call.


l0gicgate:
Review updates pushed:

- Preserved symbolic-link replacement semantics after routing file staging through the hardened recursive copier. Overwritten links are now staged descriptor-relatively and published atomically.
- Added regression coverage for replacing one symlink with another.
- Removed verbose comments that narrated test behavior rather than documenting implementation constraints.

Manual verification: N/A — this is internal file-operation plumbing with no UI change; the new regression test verifies both the replaced link target and unchanged source link.

PR #371, opened 2026-09-05
perf(browser): defer explorer list bind work until scroll settles

### LINKED ISSUE #368: perf(browser): make explorer view scrolling as fast as grid
**Strata version**

0.10.0 (main @ c19f8b7)

**Installation method**

Built from source

Omarchy, Hyprland

**What happened?**

List (Explorer) feels janky when flinging or dragging the scrollbar through a large folder. Grid stays smooth. PR #310 already fixed the shared listing-pool leaks and added scroll-settle for Grid; Explorer still does thumbnail, metadata-fill, live date, cut-style, and file-appear work on every recycled row.

**Steps to reproduce**

1. Open a large local folder (the 10,000-entry fixture is enough).
2. Switch to Grid and fling or thumb-drag the scrollbar.
3. Switch to List and repeat the same motion.
4. Observe List hitching while Grid stays even.

**Expected behavior**

List scrolling should stay as continuous as Grid. Names and metadata stay correct while moving; icons and other expensive paint can catch up after the scrollbar settles.

**Additional context**

Regression in feel, not a crash. After #310 / #328, Grid defers bind work for 80ms after the last adjustment change. Explorer never got that path, so later bind cost (thumbnails, custom folder icons, live modified dates) shows up only in List.

ogarza:
## Description

Extract Grid's 80ms adjustment settle into a shared `install_scroll_settle` and attach it to Explorer's inner `ListView` scroller.

While `value-changed` is firing, Explorer bind still writes the five column strings (and skips `set_label` when the text is unchanged) so recycled rows keep the right name/size/date. It no longer starts `set_thumbnail_or_icon`, `request_metadata_fill`, cut CSS, `set_modified_date` timer registration, or `file-appear`. Those run once on the bound pool after the timeout via `refresh_explorer_section`. Placeholder / new-item bind stays on the unfiltered branch and is not gated.

Row chrome is built by `assemble_explorer_row`; bind, settle refresh, and metadata updates share `explorer_row_parts` so that walk cannot drift. `.explorer-fast-scroll` disables row transitions the same way `.grid-fast-scroll` does.

A clamp from hiding files near the bottom of the list can trip the same settle flag as a fling; icons catch up on the timeout. Clipboard, undo, and rename stay on the model/selection path.

## Visual evidence

https://github.com/user-attachments/assets/74170168-f8f3-4b90-816e-cc73e96f569f

## How to test

1. Open a large listing and fling / thumb-drag in List, then the same motion in Grid.
2. Stop scrolling and confirm icons, cut fade, and live dates fill in on the visible rows.
3. Create a file, rename, copy/paste, undo, and toggle hidden files while scrolled near the bottom.

**Expected result:** List bind stays on text during movement; expensive work runs after settle. New-item field still appears. Hidden toggle may pop icons in ~80ms if the scrollbar clamped.

## Related issue

Closes #368


l0gicgate:
Review follow-up:

- Merged the latest `main` into the PR branch.
- Guarded the deferred fast-scroll CSS update so an idle callback cannot restore the transition-suppression class after the 80 ms settle timeout has already completed. This avoids leaving row transitions disabled when the main loop is delayed.

Manual verification: built with all features and opened the deterministic 10,000-entry fixture in Explorer using isolated XDG state. The visible rows populated names, modes, types, modified dates, and icons correctly. The PR's existing video remains representative of the scrolling behavior. I also captured a current Explorer screenshot locally at `/tmp/pr371-explorer-review.png`; the available GitHub CLI has no attachment-upload support, so I could not attach it to this comment.

l0gicgate:
Pair-testing completed on the fresh `aeb938c` build with the staged 10,000-entry Explorer fixture. Fast scrolling and settle behavior were confirmed working smoothly with no reported row, icon, or operation regressions.

PR #376, opened 2026-09-05
fix(search): index hidden files when the show-hidden preference is on

### LINKED ISSUE #366: fix(browser): Search skips hidden files even when they are visible
**Strata version**  
`c19f8b7` (source)

**Installation method**  
Built from source

Omarchy, Hyprland

**What happened?**  
With hidden files shown, filtering or searching for e.g. `.steam` (Ctrl+F or Ctrl+K) returns nothing. The listing already has the match; search replaces it with an index that never includes dotfiles.

**Steps to reproduce**  
1. Open any folder that contains a visible hidden file, e.g. `.steamid`.  
2. Turn on hidden files (Ctrl+H or Ctrl+.). Confirm the file is in the listing.  
3. Press Ctrl+F and type a prefix of that name (e.g. `.steam`).  
4. Repeat with Ctrl+K.  
5. Repeat in Columns, Grid, and Explorer.

**Expected behavior**  
If hidden files are shown, search and filter should return those names. `.steamid` should match `.steam`.

**Screenshots or video**  
N/A — not a visual glitch. A short clip of the listing vs empty results helps if you have one.

**Relevant logs**  
N/A

**Additional context**  
Every search path uses `index_tree` (`src/services/search.rs`), which walks with `ignore::WalkBuilder` and `.hidden(true)` plus `standard_filters(true)`. That skips names starting with `.` and ignores the show-hidden preference.

- Columns Ctrl+F: non-empty query swaps the pane onto that index (`src/ui/browser.rs`).  
- Grid/Explorer Ctrl+F: `inline_search::wrap` hides the listing and uses the same index (`src/ui/inline_search.rs`). The pane filter would match `.steamid`, but the search overlay covers it.  
- Ctrl+K: `SearchDialog` always indexes `$HOME` through the same walker (`src/ui/window.rs`).

Not intermittent. Happens in all three view modes.

pdwaldrop:
## Description

`index_tree`'s `ignore::WalkBuilder` always called `.hidden(true)`, unconditionally excluding dotfiles from the index regardless of whether the browser was actually showing hidden files. This affected every search surface: Columns and Grid/Explorer `Ctrl+F`, the `Ctrl+K` search dialog, and (found while fixing this, same root cause) the Move/Copy destination path search.

`index_tree` and `index_tree_with_budget` now take a `show_hidden` flag and pass `.hidden(!show_hidden)` to the walker. One subtlety worth calling out: that call has to come **after** `.standard_filters(true)` in the builder chain -- `standard_filters` enables its own `.hidden(true)` internally, so if `.hidden(!show_hidden)` were called first (a natural first attempt), `standard_filters` would silently override it back to always-hidden. I caught this because my first version of the fix passed its own test but still failed the "does this differ from the buggy code" check, which is what led me to look at the bundle's implementation.

Each call site now reads the current show-hidden preference at search time: per-column preferences for Columns' recursive search (matching how that search is already scoped to one column's subtree), the global preference everywhere else.

## Visual evidence

Before (hidden files shown in the browser, `.steamid` present, but searching for it returns nothing):

before

After (same folder, same query, same hidden-files-shown state):

after

## How to test

1. Open a folder containing a visible hidden file (e.g. `.steamid`).
2. Turn on hidden files (`Ctrl+H` or `Ctrl+.`) and confirm the file appears in the listing.
3. Press `Ctrl+F` and type a prefix of the name (e.g. `steam`).
4. Repeat with `Ctrl+K`, and in Columns, Grid, and Explorer.

**Expected result:** the hidden file matches. Turning hidden files back off and repeating the search should return no matches again.

Automated coverage: `hidden_files_are_indexed_only_when_show_hidden_is_enabled` in `src/services/search/tests.rs` exercises both directions directly against the walker.

## Related issue

Closes #366


l0gicgate (approved):
Reviewed against the latest main. No merge-blocking issues found; the hidden-file preference is propagated across the recursive search surfaces and covered in both states.

PR #379, opened 2026-09-05
fix(browser): focus first item after grid navigate

### LINKED ISSUE #374: bug(browser): in Grid view, enter on a sidebar place does not focus files or allow Right into the listing
**Strata version:**
`fix/291-keyboard-focus-ownership` 1c526bde82ac6c7ff229d0c39911192d65a99f80

**Installation:**
Built from source

**Environment:**
Omarchy, Hyprland

**What happened?**  
In Grid, moving to a sidebar place with the arrows and pressing Enter opens that folder, but keyboard focus does not land on the first item. Right also does not enter the listing. Left, then Right twice, is required before Grid arrows work.

**Steps**
1. Open Strata in Grid with the sidebar visible.
2. Left-arrow into the sidebar.
3. Up/Down to a different place (e.g. Documents).
4. Enter.
5. Try Right or the other arrows without clicking.

**Expected**  
Focus the first item, or at least let Right enter the current file view (as in `docs/keyboard-navigation.md`).

**Actual**  
The listing opens; arrows do nothing until Left, then Right twice.

**Additional context**  
Root cause is the Enter → `navigate()` → `FocusChanged { position: None }` → `focus_visible_pane()` path: it steals sidebar focus onto a Grid with no selected item, so both “auto-focus first item” and “Right returns to the current file view” miss. `select_first_on_load` is never set for sidebar `navigate()`. Explorer likely shares the same `navigate()` path.

Hakira-Shymuy:
## Description

Keyboard focus was staying on the sidebar (or otherwise leaving the file pane) after opening a folder in grid/explorer. Navigate now requests the first entry when a column finishes loading, scrolls with FOCUS|SELECT, and refocuses the visible pane so Enter/arrow keys operate on files.

## Visual evidence

N/A — this is a keyboard/focus behavior change, not a layout or visual styling change.

## How to test

1. Open a folder in grid view with the sidebar visible and the file pane focused.
2. Press Enter on a directory (or otherwise navigate into it).
3. Repeat in explorer/column view.
4. After the destination loads, press an arrow key or Enter without clicking the file list.

**Expected result:** The first item in the opened folder is selected, the file pane keeps keyboard focus, and further keys act on that view instead of the sidebar.

## Related issue

Closes #374

### DIFF
--- a/src/app/browser.rs
+++ b/src/app/browser.rs
@@ -648,6 +648,10 @@ impl Browser {
         self.state.borrow_mut().focus_column(depth);
     }
 
+    pub fn select_first_on_load(&self, depth: usize) {
+        self.state.borrow_mut().select_first_on_load(depth);
+    }
+
     pub fn focus_active(&self) {
         let focus = self.state.borrow().active_focus();
         if let Some((depth, position)) = focus {
--- a/src/app/browser/tests.rs
+++ b/src/app/browser/tests.rs
@@ -1710,6 +1710,21 @@ fn activating_an_open_list_item_closes_its_child_column() {
 }
 
 #[test]
+fn requesting_first_selection_during_navigate_selects_the_first_entry() {
+    let browser = Browser::new(Rc::new(FakeFileSource));
+    let loader = browser.clone();
+    browser.observe(move |event| {
+        if matches!(event, BrowserEvent::ColumnAdded { depth: 0, .. }) {
+            loader.select_first_on_load(0);
+        }
+    });
+
+    browser.navigate(Location::local("/fixture"));
+
+    assert_eq!(browser.selected_positions(0), [0]);
+}
+
+#[test]
 fn explorer_activation_replaces_the_directory_instead_of_adding_a_column() {
     let browser = Browser::new(Rc::new(FakeFileSource));
     let events = Rc::new(RefCell::new(Vec::new()));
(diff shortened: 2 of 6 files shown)


l0gicgate:
@Hakira-Shymuy did you observe this on main? We merged #358 which reworked a ton of stuff with keyboard focus/navigation.

ogarza:
I saw this behavior in `fix/291-keyboard-focus-ownership` right before the merge

Hakira-Shymuy:
> @Hakira-Shymuy did you observe this on main? We merged #358 which reworked a ton of stuff with keyboard focus/navigation.
@l0gicgate @ogarza 

Uh, seems fixed lmao, im 20 minutes late to the party :/

when i cloned was not fixed, that was like 1 hour ago
switched to main and seems to work nice on make run-dev

feel free to discard the pr :/

ogarza:
@l0gicgate is too fast for us

l0gicgate (approved):
Reviewed the updated diff against main. The focus handoff and first-item selection are covered in both Grid and Explorer, and the full local pre-push suite passes.

ogarza:
oops missclick

l0gicgate:
Nono! This PR fixes it! You probably didn't go back to main @Hakira-Shymuy to test. I just switched between main and this and your PR fixes it.

ogarza:
one thing I noticed is once in the grid or explorer, how do you go back to the left navigation bar?

Hakira-Shymuy:
> one thing I noticed is once in the grid or explorer, how do you go back to the left navigation bar?

so far the same way as before, 
you just arrow your way out of the browser

One behavior i noticed and most likely would change is the auto focus

If i change from list to grid, and i step into the arrows, the navigation is still focused on the Header, as a user i would expect to be auto focused already inside the browser

ogarza:
yes, was an issue on my end, all good! thanks!

PR #389, opened 2026-09-05
fix(browser): restore column rows when the miller filter is cleared

### LINKED ISSUE #377: bug(browser): in Column view clearing the filter leaves fallback icons until the listing is rebuilt
### Strata version
0.10.0 (built from source, commit f584b93)

### Installation method
Built from source

Omarchy (Arch), Hyprland

### What happened?
In Column (Miller) view, after using a pane filter and then backspacing it empty, rows in that column lose their thumbnails (fallback icons) and the file-size overlay that normally appears on hover.

Scrolling and resizing do not restore either. Changing the column’s sort order restores both.

Names stay correct. Explorer and Grid do not appear to do this.

### Steps to reproduce
1. Open Strata in Column (Miller) view.
2. Open a folder that already shows image thumbnails.
3. Hover a file and confirm the size overlay appears.
4. Focus that column and open its filter (`Ctrl+F`, or the funnel button).
5. Type a query and wait briefly so the listing actually changes (~200ms; once the query settles, the column filter searches subfolders).
6. Backspace until the filter is empty.
7. Confirm thumbnails are fallback icons, and hovering a file no longer shows its size.
8. Scroll the column and resize the window: both stay wrong.
9. Change the sort field or sort direction: thumbnails and hover sizes come back.

### Expected behavior
Clearing the filter should restore the directory listing with the same thumbnails and hover sizes as before. Those should not stay missing until a later listing rebuild.

### Screenshots or video
https://github.com/user-attachments/assets/3dada8a2-046b-4005-9147-9b9a1b7ca55e

### Relevant logs
N/A (no crash).

### Additional context
The stuck state is a one-shot bad row bind when the filter is cleared, not the thumbnail settle/viewport gate. Scroll and resize do not rebind rows that stay on screen. Sort does: it emits `EntriesReplaced` → `EntryListModel::replace` → `items_changed(0, n, n)`, which unbinds and rebinds every visible row.

Miller columns are different from Explorer/Grid here. A settled non-empty query swaps the column `FilterListModel` onto a recursive-search `StringList`; clearing the query swaps it back to the directory model (`src/ui/browser.rs`, miller `debounce_filter_entry` handler). On empty query the handler:

1. Clears `search_results` and empties the search model
2. Calls `set_model` back to the directory list (GTK binds visible rows here)
3. Only then sets `recursive_search_active` to false

The factory bind still sees search as active during step 2, looks up the already-empty `search_results`, and takes the `entry is None` branch. That path:

- paints the generic documents fallback icon
- hides folder chevrons
- clears `.file-size` (`column_size_text(None)` → empty label, `set_visible(false)`), so hover CSS has nothing to fade in

If `set_filter(None)` is then a no-op (search already removed the filter), nothing rebinds until the next model rebuild.

Losing hover size together with the thumbnails is the missing-entry bind, not a thumbnail-cache miss.

Also restores (same class of rebuild as sort): toggling folders-first, reloading the column, navigating away and back. Does not restore: scroll, window resize.

Related: subfolder search from the miller filter landed in #275. Not a crash; no core dump.

spandan11106:
## Description

In Column (Miller) view, clearing a settled recursive-search filter left the column's rows with generic fallback icons, no folder chevrons, and no hover file-size overlay until an unrelated listing rebuild.

The empty-query branch of the column filter handler swapped the `FilterListModel` back to the directory listing *before* clearing `recursive_search_active`. GTK binds the visible rows synchronously inside `set_model`, so the row factory still believed search was active, looked up the already-emptied `search_results`, and took its missing-entry path — fallback icon, hidden chevron, and an empty `.file-size` label for hover CSS to fade in. Scroll and resize do not rebind rows that stay on screen, so the state persisted until something rebuilt the model (sort change, folders-first toggle, reload, navigating away and back).

The reset now lives in `deactivate_recursive_search`, which drops the flag before the model swap so those rows rebind against the directory entries. A regression test asserts the flag is already clear when the filtered model announces the re-added rows; it fails on the previous ordering.

## Visual evidence

N/A — the fix restores the thumbnails and hover sizes shown in the issue's video; there is no new or changed UI.

## How to test

1. Open Strata in Column (Miller) view and enter a folder that shows image thumbnails.
2. Hover a file and confirm the size overlay appears.
3. Focus the column, open its filter (`Ctrl+F`), type a query, and wait ~200ms for the subfolder search to settle.
4. Backspace until the filter is empty.
5. Hover files and scroll the column.

**Expected result:** thumbnails and hover file sizes come back immediately when the filter empties, with no need to change the sort order or reload the column.

## Related issue

Closes #377

### DIFF
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -5107,20 +5107,19 @@ impl ViewState {
             if query.is_empty() {
                 search_gen_for_changed.set(search_gen_for_changed.get().saturating_add(1));
                 search_handle_for_changed.borrow_mut().take();
-                search_results_for_changed.borrow_mut().clear();
-                search_model_for_changed.splice(0, search_model_for_changed.n_items(), &[]);
-                if filtered_model_for_search.model().as_ref()
-                    != Some(model_for_search.upcast_ref::<gio::ListModel>())
-                {
-                    filtered_model_for_search.set_model(Some(&model_for_search));
-                }
+                deactivate_recursive_search(
+                    &search_active_for_changed,
+                    &search_results_for_changed,
+                    &search_model_for_changed,
+                    &filtered_model_for_search,
+                    &model_for_search,
+                );
                 apply_filter_query(
                     &filtered_model_for_search,
                     &filter,
                     &filter_query,
                     text.to_lowercase(),
                 );
-                search_active_for_changed.set(false);
                 return;
             }
             *filter_query.borrow_mut() = text.to_lowercase();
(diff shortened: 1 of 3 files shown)


l0gicgate:
Review complete. I merged the latest `main` into the PR head and reviewed the resulting diff. No blocking issues were found.

I also verified the requested type-to-search coverage: `BrowserView::show_filter_with_optional_query` routes Columns to the focused column and routes Icons/List through `ModeViews::show_filter_with_query`; Icons and List both use the shared recursive inline-search wrapper. Clearing a recursive query restores each mode’s normal presentation, while this PR fixes the synchronous row-rebind ordering specific to Columns.

Verification: the complete local format, Clippy, and test suite passes (786 passed, 18 display-dependent tests ignored by the suite), including the recursive-search clear regression and alternate-view descendant-search restoration test. No new UI was introduced during review, so no new screenshot was needed.

l0gicgate (approved):
Reviewed against the latest main. The row-rebind ordering fix is correct and covered by a regression test; type-to-search remains wired through Columns, Icons, and List. Full local checks and required PR checks pass.

PR #396, opened 2026-09-05
feat(trash): hide Empty Trash from the sidebar menu when Trash is empty

### LINKED ISSUE #367: feat(trash): hide Empty Trash from the context menu when empty
## Problem or use case

The Trash sidebar context menu still shows **Empty Trash…** when Trash contains no items. Although the operation is unavailable/no-op, displaying a destructive action that cannot currently do anything adds clutter and suggests that Trash may not actually be empty.

This is especially noticeable because the menu then consists of **Properties**, a separator, and an unusable **Empty Trash…** row.

## Proposed solution

Hide **Empty Trash…** from the Trash sidebar context menu unless Trash is confirmed to contain at least one item.

- Track whether `trash:///` has any top-level entries even when Trash is not the currently open location.
- Keep that state synchronized when an item is trashed, restored, permanently removed, emptied, or changed by another application.
- Build/update the context menu before it is shown so **Empty Trash…** appears immediately when Trash is non-empty and disappears when it becomes empty.
- Hide the associated separator at the same time so **Properties** remains the only row in the empty state.
- If Trash state is still loading or cannot be queried, fail safely without presenting an actionable destructive option until non-empty state is confirmed.

The in-pane/header Empty Trash button can retain its existing disabled-when-empty behavior; this issue is specifically about simplifying the sidebar context menu.

## Alternatives considered

- Keep the option visible but disabled: prevents activation, but retains unnecessary visual clutter and was the behavior this request aims to improve.
- Remove the Empty Trash context action entirely: simpler, but inconvenient when Trash contains items and the user wants to empty it without navigating there first.

## Visual examples

The reported screenshot shows the Trash sidebar context menu containing **Properties**, a separator, and **Empty Trash…** while Trash is empty.

## Strata version

`main`

## Additional context

Related to #130 / #138, which disabled the in-pane Empty Trash controls and guarded the operation when Trash is empty. That change did not synchronize or hide the sidebar context-menu row.

### Acceptance criteria

- [ ] When Trash is empty, its sidebar context menu contains no **Empty Trash…** row or orphaned separator.
- [ ] When Trash contains at least one item, **Empty Trash…** is visible and retains the existing confirmation flow.
- [ ] The menu updates after trash, restore, and empty operations without restarting Strata.
- [ ] External changes to `trash:///` are reflected the next time the context menu opens.
- [ ] Unknown/error state does not expose an incorrectly actionable destructive option.
- [ ] Tests cover empty, non-empty, transition, and separator visibility states.

spandan11106:
## Description

The Trash sidebar context menu always offered **Empty Trash…**, even with an empty Trash, leaving a destructive row that could not do anything above an orphaned separator.

The sidebar now tracks whether `trash:///` has any top-level entry and hides both the **Empty Trash…** row and its separator until Trash is confirmed non-empty, so an empty Trash shows only **Properties**. Work done for the issue:

- Tracks `trash:///` top-level emptiness even when Trash is not the open location, via a cheap probe that stops after the first entry.
- Keeps that state current after trashing, restoring, permanently deleting, emptying, moving and cancelled operations (browser operation events), and when another application changes `trash:///` (a `trash:///` file monitor).
- Refreshes the state when the context menu is opened, so an external change is reflected the next time the menu is shown.
- Hides the separator together with the row, leaving **Properties** as the only entry in the empty state.
- Fails safe: an unread or failed probe leaves the state unknown, which keeps the destructive row hidden. Concurrent refreshes are coalesced so an in-flight probe cannot publish state that predates the change that triggered it.
- The in-pane/header Empty Trash button keeps its existing disabled-when-empty behavior.
- Tests cover the empty, non-empty and unknown states, the separator following the row, the probe-result mapping, which operations refresh the menu, and the probe itself against an empty directory, a populated one and an unreadable one.

## Visual evidence

N/A — the change only removes rows from a right-click popover, and this environment cannot drive a synthetic right-click at the sidebar (the compositor's focus dispatcher rejects the automation used for screenshots), so no honest before/after capture could be produced. The manual steps below show the difference in a few seconds.

## How to test

1. Empty Trash (`gio trash --empty`), start Strata, and right-click **Trash** in the sidebar. The menu contains only **Properties** — no separator, no **Empty Trash…**.
2. Without restarting, trash a file (select it in Strata and press Delete, or run `gio trash <file>` from a terminal to exercise the external-change path).
3. Right-click **Trash** again. **Empty Trash…** is back, below its separator, and still opens the usual confirmation dialog.
4. Use it to empty Trash, then right-click **Trash** once more.

**Expected result:** the **Empty Trash…** row and its separator appear only while Trash holds at least one item and disappear again as soon as it is empty, without restarting Strata.

## Related issue

Closes #367


l0gicgate (approved):
Reviewed against the latest main. The merge conflict was resolved by retaining both the new trash-state coverage and the current window tests. The implementation and tests look merge-ready, and the full local pre-push suite passes.

PR #407, opened 2026-09-06
fix(archives): stop following symbolic links during compression

### LINKED ISSUE #402: fix(archives): preserve symbolic links during compression
### Strata version

### Installation method

Built from source.

### What happened?

Compression follows symbolic links instead of preserving them. A selected folder can therefore include file contents from outside that folder in the resulting archive. This can disclose unintended contents if the archive is shared.

The behavior affects ZIP, TAR, TAR.GZ, and 7z. The compression dialog does not offer a link-handling choice or explain that targets will be included.

### Steps to reproduce

Confirmed by source review, not a GUI or disclosure reproduction:

1. Trace the ZIP and 7z directory walkers and their file-opening calls. They follow link targets without checking whether an entry is a symbolic link.
2. Check the TAR builder configuration. The pinned library defaults to following links, and Strata does not disable that behavior.
3. Check the compression dialog for a link-preservation option. None is provided.

### Expected behavior

Preserve symbolic links without copying their targets, or report unsupported links clearly. Following targets should require an explicit user choice.

### Screenshots or video

N/A. Confirmed from the implementation and pinned library sources.

### Relevant logs

Not collected. No disclosure test was performed.

### Additional context

Relevant code: [ZIP walker](https://github.com/lgse/strata/blob/a96c4a097f99d9183bd41c9976695b250858a0bb/src/adapters/local_operations.rs#L3038), [TAR builder](https://github.com/lgse/strata/blob/a96c4a097f99d9183bd41c9976695b250858a0bb/src/adapters/local_operations.rs#L3093), and [7z walker](https://github.com/lgse/strata/blob/a96c4a097f99d9183bd41c9976695b250858a0bb/src/adapters/local_operations.rs#L3473). The confidentiality impact depends on the input folder and who can access the resulting archive.

kallupx:
## Description

Prevent compression from including symbolic link targets. ZIP, TAR, and TAR.GZ preserve the links. 7z stops with an error suggesting ZIP or TAR, without committing a partial archive.

File counting and archive writing use the same traversal, which does not follow symbolic links.

Strata's extractor currently does not recreate links. Archive verification uses `bsdtar`.

## Visual evidence

7z unsupported-link error

## How to test

1. Create a folder containing ordinary files, a file link, a folder link, and a broken link.
2. Compress it as ZIP, TAR, and TAR.GZ. Inspect and extract each archive with `bsdtar`, then compare file contents and link targets with the originals.
3. Repeat with a directly selected symbolic link.
4. Try 7z compression with a link, then with a folder containing only ordinary files.

**Expected result:** ZIP and TAR preserve links without copying their targets. 7z reports unsupported links and leaves no partial archive. Ordinary files still compress correctly.

## Related issue

Closes #402


l0gicgate (approved):
Reviewed against the latest main. Symlink-safe traversal, archive behavior, error handling, and regression coverage look merge-ready; the full local suite and current PR checks pass.

PR #408, opened 2026-09-06
fix(archives): keep compression staging files private

### LINKED ISSUE #404: fix(archives): keep compression staging files private
### Strata version

### Installation method

Built from source.

### What happened?

When replacing an archive with restrictive permissions, Strata writes a temporary archive beside it using `0666` subject to the process umask. It applies the original archive's permissions only after compression finishes.

This can make the temporary archive more widely readable than the archive being replaced. Exposure requires a destination directory accessible to other users and permissive effective staging permissions.

### Steps to reproduce

Confirmed by source review and an isolated check of the pinned `tempfile` dependency:

1. Inspect `write_staged_archive`: staging is created with `0666`, before the encoder runs.
2. Check where the existing archive's permissions are restored: after encoding completes.
3. Verify the configured tempfile mode inside a private test directory under umasks `022` and `077`.

With umask `022`, staging is `0644`; with `077`, it is `0600`. No cross-user access test was performed.

### Expected behavior

Create staging files with `0600`. Apply the intended final permissions only after compression succeeds, immediately before publishing the archive. Preserve existing permissions when replacing an archive.

### Screenshots or video

N/A. Verified through file modes and the order of operations.

### Relevant logs

```text
umask 022: default tempfile=600, explicit 0666 tempfile=644
umask 077: default tempfile=600, explicit 0666 tempfile=600
```

### Additional context

[Staging and permission restoration](https://github.com/lgse/strata/blob/a96c4a097f99d9183bd41c9976695b250858a0bb/src/adapters/local_operations.rs#L1782). A private parent directory prevents access by other users. The existing replacement test checks final permissions but not permissions during compression.

wmfeht:
## Description

Compression used to create the sibling staging file with mode `0666` (subject to umask) and only restore the existing archive's permissions after encoding finished. In a shared directory that made the temp archive more widely readable than the file being replaced.

Staging files are now created with mode `0600`. Intended permissions are applied only after compression succeeds, immediately before persist. Replacing an archive still copies the original mode onto the published result; new archives still receive the usual umask-adjusted `0666` creation mode.

## Visual evidence

N/A — this is a file-mode change with no user-visible UI.

## How to test

1. Create a destination directory that other users can traverse, and an archive in it with mode `0640`.
2. Compress a replacement into that archive (Replace) and pause or inspect `.strata-compression-*` while encoding is in progress.
3. Confirm the staging file is `0600` during encoding, then that the published archive is `0640`.
4. Compress a new archive into the same directory and confirm the published mode matches a normal `File::create` (typically `0644` with umask `022`).

**Expected result:** Staging stays owner-only while encoding. Replacement preserves the original archive mode. New archives keep umask-based defaults.

## Related issue

Closes #404

### DIFF
--- a/src/adapters/local_operations.rs
+++ b/src/adapters/local_operations.rs
@@ -1769,7 +1769,7 @@ async fn write_staged_archive<F>(
 where
     F: FnOnce(std::fs::File) -> Result<(), String> + Send + 'static,
 {
-    let existing_permissions = if conflict == TransferConflict::ReplaceExisting {
+    let published_permissions = if conflict == TransferConflict::ReplaceExisting {
         match std::fs::symlink_metadata(archive_path) {
             Ok(metadata) if metadata.file_type().is_file() => Some(metadata.permissions()),
             Ok(_) => None,
@@ -1778,11 +1778,12 @@ where
         }
     } else {
         None
-    };
+    }
+    .unwrap_or_else(umask_adjusted_file_permissions);
     let mut builder = tempfile::Builder::new();
     builder
         .prefix(".strata-compression-")
-        .permissions(std::fs::Permissions::from_mode(0o666));
+        .permissions(std::fs::Permissions::from_mode(0o600));
     let staged = builder
         .tempfile_in(destination)
         .map_err(|error| error.to_string())?;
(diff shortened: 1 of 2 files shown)


wmfeht:
Superseded by the combined archive-fix PR: https://github.com/lgse/strata/pull/410

Closing this in favor of that unified PR (private staging + cancelable extraction).

PR #409, opened 2026-09-06
fix(archives): stop extraction work when cancelled

wmfeht:
## Description

Cancelling an in-progress extraction dismissed the progress UI while the `gio::spawn_blocking` worker kept creating files. The provider set an atomic flag and aborted the supervising GLib task, but only the progress timer read that flag, and dropping the blocking handle detached the worker.

Cancellation is now passed into the extraction and copy loops. The supervising task stays alive until the worker observes the flag and returns. Incomplete output is reported as a cancelled operation, and the progress UI is dismissed only after that terminal event.

## Visual evidence

N/A. Worker lifecycle and incomplete-output handling; the archive progress subtitle now matches other file operations (`Cancelling will not undo completed changes`).

## How to test

1. Extract a multi-file archive large enough that the progress dialog appears.
2. Click **Cancel** after extraction has started writing files.
3. Confirm the progress dialog stays up until extraction actually stops.
4. Confirm remaining archive members are not created, and the cancellation dialog reports completed/failed/not-attempted counts. Already-written files are left in place.

**Expected result:**

Further extraction stops promptly. The progress UI dismisses only after the worker has stopped. Incomplete output is described, and completed files are not reverted.

## Related issue

Closes #403

PR #410, opened 2026-09-06
fix(archives): private staging and cancelable extraction

### LINKED ISSUE #403: fix(archives): stop extraction work when cancelled
### Strata version

### Installation method

Built from source.

### What happened?

Cancelling an already-started extraction dismisses its progress UI but does not stop the blocking worker. That worker can continue creating files and writing data after cancellation.

### Steps to reproduce

Confirmed by source review and an isolated check of the pinned GIO dependency:

1. Trace the archive progress dialog's Cancel action to `cancel_file_operation` and the provider's cancellation handler.
2. Check the extraction worker and copy loops. They never receive or check the cancellation flag.
3. Start a small blocking GIO task, abort its supervising GLib task, then allow the worker to continue. The worker still completes.

The lifecycle check used an in-memory task. A full GUI extraction-cancellation test has not been run.

### Expected behavior

Cancellation should stop further extraction work promptly. Dismiss the progress UI only once the worker has stopped, and make the handling of incomplete output clear.

### Screenshots or video

N/A. Confirmed through the cancellation call path and worker lifecycle check.

### Relevant logs

```text
GIO lifecycle check: started blocking worker completed after supervising GLib task was aborted
```

### Additional context

The [provider](https://github.com/lgse/strata/blob/a96c4a097f99d9183bd41c9976695b250858a0bb/src/adapters/local_operations.rs#L2902) sets an atomic flag and aborts the supervising task, but only the progress timer reads the flag. Dropping a `gio::spawn_blocking` handle detaches its worker. Pass cancellation into the extraction and copy loops and check it during work.

### LINKED ISSUE #404: fix(archives): keep compression staging files private

wmfeht:
## Description

Combines the two archive-operation fixes previously opened as separate PRs (#408 and #409).

**Private compression staging (#404):** Compression created sibling staging files with mode `0666` (subject to umask) and restored the existing archive's permissions only after encoding. In a shared directory that made the temp archive more widely readable than the file being replaced. Staging files are now created with mode `0600`. Intended permissions are applied only after compression succeeds, immediately before persist. Replacing an archive still copies the original mode onto the published result; new archives still receive the usual umask-adjusted `0666` creation mode.

**Cancelable extraction (#403):** Cancelling an in-progress extraction dismissed the progress UI while the `gio::spawn_blocking` worker kept creating files. The provider set an atomic flag and aborted the supervising GLib task, but only the progress timer read that flag, and dropping the blocking handle detached the worker. Cancellation is now passed into the extraction and copy loops. The supervising task stays alive until the worker observes the flag and returns. Incomplete output is reported as a cancelled operation, and the progress UI is dismissed only after that terminal event.

This supersedes https://github.com/lgse/strata/pull/408 and https://github.com/lgse/strata/pull/409.

## Visual evidence

N/A — staging is a file-mode change. Extraction cancellation is worker lifecycle and incomplete-output handling; the archive progress subtitle now matches other file operations (`Cancelling will not undo completed changes`).

### Private compression staging

1. Create a destination directory that other users can traverse, and an archive in it with mode `0640`.
2. Compress a replacement into that archive (Replace) and pause or inspect `.strata-compression-*` while encoding is in progress.
3. Confirm the staging file is `0600` during encoding, then that the published archive is `0640`.
4. Compress a new archive into the same directory and confirm the published mode matches a normal `File::create` (typically `0644` with umask `022`).

**Expected result:** Staging stays owner-only while encoding. Replacement preserves the original archive mode. New archives keep umask-based defaults.

### Cancelable extraction

1. Extract a multi-file archive large enough that the progress dialog appears.
2. Click **Cancel** after extraction has started writing files.
3. Confirm the progress dialog stays up until extraction actually stops.
4. Confirm remaining archive members are not created, and the cancellation dialog reports completed/failed/not-attempted counts. Already-written files are left in place.

**Expected result:** Further extraction stops promptly. The progress UI dismisses only after the worker has stopped. Incomplete output is described, and completed files are not reverted.

## Related issue

Closes #403
Closes #404


l0gicgate:
Merged the latest `main` and completed the merge review. I fixed two blockers found during review:

- TAR cancellation now returns as soon as the current entry stops instead of scanning/decompressing the rest of the stream just to enumerate unattempted names.
- Compression permission tests no longer change the process-wide umask, avoiding interference with concurrently running tests.

The merge also preserves `main`'s solid-7z extraction coverage, centralized GIO location conversion, browser UI module split, and thumbnail test fixtures.

Manual verification: inspected the final archive diff and cancellation event flow. No new visual behavior was introduced by the review fixes, so no additional screenshot was captured.

l0gicgate (dismissed):
Reviewed the updated diff after merging the latest main. Archive staging, cancellation, and the merged symlink/solid-7z behavior are covered and the required local suite passes.

l0gicgate:
Fixed the folder-peek highlight regression observed during manual testing in `fb62b70` (fixes #426). The hovered directory now remains the sole highlighted row while its peek is open; the transient peek list also no longer participates in focus navigation.

Manual reproduction verified on Omarchy/Hyprland 0.56.2 using: open `compression-destination`, then hover `payload`. The full local format, Clippy, and test suite passes.

l0gicgate:
Fixed the cross-column context-menu targeting issue in `78a12f9` (fixes #427). Actions now prefer the explicitly right-clicked item when an unrelated selection remains in another column, and compression derives its output directory from the selected entries rather than transient focus. Added regression coverage.

Manual verification: `payload` compressed to `compression-destination/new-archive.zip`; staging was mode `600`, final archive is mode `644`, and no staging files remain. Full local format, Clippy, and test suite passes.

l0gicgate:
Added agent test-display guidance in `e71f098` (fixes #440). `AGENTS.md` now requires GTK tests to run under private Xvfb with accessibility bridging disabled and explicitly forbids falling back to the active desktop or accepting GTK initialization skips.

Verified the full suite under portable Xvfb: 827 passed, 18 intentionally ignored, 0 failed, with no GTK initialization skips.

PR #420, opened 2026-09-06
fix(browser): do not insert slash when starting the pane filter

### LINKED ISSUE #418: fix(browser): do not insert slash when starting the pane filter
### Strata version

v0.11.1, source commit a96c4a097f99d9183bd41c9976695b250858a0bb (`main`, synced with this upstream).

### Installation method

Built from source.

Ubuntu 24.04.4 LTS, X11, GTK 4.14.5. Isolated temp `HOME` / `XDG_*` with synthetic filenames only. Type to search left at the default (enabled).

### What happened?

Pressing `/` with a file view focused opens the pane filter already seeded with `/`. Ordinary names in the current folder do not match that query. Typing after `/` keeps the leading slash (`/n`), so `notes.txt` still does not match. Ctrl+F opens the same filter empty, and then `n` matches as expected.

This happens in every browser view Strata exposes: **Columns**, **Icons**, and **List** (Appearance menu labels; some docs also say Grid / Explorer).

### Steps to reproduce

1. Open a folder of ordinary names (repro used `LICENSE`, `notes.txt`, `readme.md`, `beta-report.txt`, `alpha-folder/child.txt`). Confirm Settings → General → Type to search is on.
2. Click a file so the file view has keyboard focus (not the location bar).
3. Press `/`.
4. Observe the filter field and the listing.
5. Type `n`.
6. Press Escape, then Ctrl+F, then `n` for contrast.
7. Repeat steps 2–6 in Appearance → Icons and Appearance → List.

### Expected behavior

`/` should activate type-to-search / the pane filter without inserting `/` as the first query character (same as Ctrl+F: empty field, subsequent keys filter by name). `n` should match `notes.txt`.

### Screenshots or video

https://cursor.com/artifacts/v/art-2c87df1a-de69-48b9-92bc-69a621e93a43

Isolated recording covering Columns, Icons, and List:

- `/` opens the filter with `/` already in it.
- Current-folder names disappear. Nested `alpha-folder/child.txt` can appear because the recursive query `/` matches `/` in relative paths.
- After `n`, the field is `/n` and the view shows **No matching files** (while `notes.txt` is still in the folder).
- Ctrl+F opens an empty “Filter … items…” field; `n` matches `notes.txt` and `LICENSE`.

### Relevant logs

No crash. Directory load of the five-entry fixture completed normally.

### Additional context

`/` is handled as a printable type-to-search seed, not as a filter activator:

- `type_to_search_query` in `src/ui/window.rs` accepts any non-control unicode, including `/`.
- `TypeToSearch::show` then calls `show_filter_with_query` with that character, and `focus_filter_entry` writes it into the Gtk Entry.
- Ctrl+F uses `show_filter()` with no seed, so the field stays empty.
- A non-empty filter starts recursive search; `/` scores against `search_path`, which is why a nested `dir/file` can match while top-level names do not.

Related: type-to-search shipping in #307 / #304; default-on in #387; Space vs type-to-search in #414. Keep a slash-activator fix independent of the open browser refactoring.

### Repro update

**Verdict: REPRODUCED** on v0.11.1 / a96c4a0.

| View (Appearance label) | `/` seeds filter with `/` | Ordinary names fail to match | Ctrl+F empty filter works |
| --- | --- | --- | --- |
| Columns | yes | yes (`/` can still hit nested paths) | yes |
| Icons | yes | yes | yes |
| List | yes | yes | yes |

No other file-view modes are in the Appearance menu.

wmfeht:
## Description

Pressing `/` with a file view focused opened the pane filter already seeded with `/`, so ordinary names in the current folder did not match. Typing after kept the leading slash (`/n`). Ctrl+F opened the same filter empty and worked.

`/` still activates type-to-search when that preference is on, but it now opens the empty filter field used by Ctrl+F instead of inserting `/` as the first query character. Later keys still seed or type into the filter as before.

## Visual evidence

N/A for this revision: the change is keyboard seeding of an existing filter field, not layout or chrome. Screenshots from Columns / Icons / List can be attached after manual verification.

## How to test

1. Open a folder of ordinary names (for example `LICENSE`, `notes.txt`, `readme.md`) with Settings → General → Type to search enabled.
2. Click a file so the file view has keyboard focus (not the location bar).
3. Press `/`.
4. Confirm the filter field is empty (same as Ctrl+F), and that current-folder names are still listed.
5. Type `n` and confirm `notes.txt` matches. The field should be `n`, not `/n`.
6. Press Escape, then Ctrl+F, then `n` and confirm the same match.
7. Repeat steps 2–6 in Appearance → Icons and Appearance → List.

**Expected result:** `/` opens an empty pane filter; subsequent keys filter by name. Nested paths are no longer matched only because `/` appears in `search_path`.

## Related issue

Closes #418

### DIFF
--- a/src/ui/window.rs
+++ b/src/ui/window.rs
@@ -53,7 +53,13 @@ struct TypeToSearch {
 
 impl TypeToSearch {
     fn show(&self, query: char) -> bool {
-        self.preferences.type_to_search() && self.view.show_filter_with_query(&query.to_string())
+        if !self.preferences.type_to_search() {
+            return false;
+        }
+        match type_to_search_seed(query) {
+            Some(seed) => self.view.show_filter_with_query(&seed.to_string()),
+            None => self.view.show_filter(),
+        }
     }
 }
 
@@ -1183,6 +1189,11 @@ fn type_to_search_query(key: gtk::gdk::Key, modifiers: gtk::gdk::ModifierType) -
         return None;
     }
     key.to_unicode().filter(|character| !character.is_control())
+}
+
+fn type_to_search_seed(query: char) -> Option<char> {
+    // `/` activates the pane filter; it is not a query character.
+    (query != '/').then_some(query)
 }
 
 fn is_open_terminal_shortcut(key: gtk::gdk::Key, modifiers: gtk::gdk::ModifierType) -> bool {
--- a/src/ui/window/tests.rs
+++ b/src/ui/window/tests.rs
@@ -16,7 +16,7 @@ use super::{
     parse_pinned_places, pin_status, remove_pinned_place, reorder_pinned_places, reorder_places,
     resolve_place_order, serialize_pinned_places, should_show_standard_place,
     sidebar_accepts_file_drop, sidebar_update_label, standard_place, type_to_search_query,
-    vim_focus_direction, volume_release_action,
+    type_to_search_seed, vim_focus_direction, volume_release_action,
 };
 
 fn release(version: &str, kind: BuildKind) -> ReleaseMetadata {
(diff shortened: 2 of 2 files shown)


l0gicgate:
I will fix as part of #416 is for a related type to search issue.

wmfeht:
Closing — leaving this for the project author / another PR already covering #418. Thanks.

PR #421, opened 2026-09-06
feat(ui): add application chooser to file context menu

### LINKED ISSUE #174: feat(ui): add application chooser to file context menu
### Problem or use case

Strata opens files with their default application, but users cannot choose a different installed application from the file context menu. They must change the system default or leave Strata to open the file with another application.

### Proposed solution

- Add **Open With...** directly below **Open** in the file context menu.
- Use GIO application metadata and Strata's existing GTK dialog patterns to show compatible installed applications for the selected file's content type. Do not parse `.desktop` files directly.
- Launch the selected `gio::AppInfo` with the selected `gio::File` values without changing the system default.
- Support multiple selected files when they share a content type. Hide or disable the action for directories and incompatible mixed selections.
- Preserve local and URI-native locations and show the existing error dialog if the selected application cannot open them.
- Keep the chooser keyboard accessible and consistent with Strata's semantic theming.

### Alternatives considered

- Add a submenu listing every compatible application: this duplicates chooser behavior and makes the context menu longer.
- Rely only on the default application: this does not support workflows such as opening an image in either an editor or a viewer.
- Cover this through custom actions in #36: custom commands are broader and should not replace standard desktop MIME integration.

### Visual examples

N/A. Follow the conventional file-manager placement immediately below **Open**.

### Strata version

Current `main`.

### Additional context

`docs/todo.md` already tracks both an **Open With** chooser and its context-menu action. Strata currently uses GIO to launch the default application and resolve the default handler for a content type, so this should extend the existing desktop-integration path rather than introduce custom application discovery.

sandmor:
## Description

Adds an **Open With...** action below **Open** in file context menus. The chooser uses GIO application metadata, supports compatible multi-file selections, preserves native and URI locations, and launches the selected application without changing the system default.

## Visual evidence

image

## How to test

1. Right-click a supported file and choose **Open With…**.
2. Select an installed application and choose **Open**.
3. Select multiple files sharing a content type and repeat.
4. Check directories and mixed incompatible selections.

**Expected result:**

The chooser lists compatible applications, supports keyboard navigation, opens the selected files with the chosen application, and remains unavailable for directories or incompatible selections.

## Related issue

Closes #174


l0gicgate:
hey thanks for the contribution @sandmor just a heads up we are freezing new features for a couple days until we can stabilize things. There are quite a few bugs/annoyances I want to take care of before we can start adding new stuff.

This will be a nice addition though!

JoeJoeflyn:
Beautiful!!

l0gicgate:
Merged current `main`, preserving its context-selection and capability handling. Fixed initial chooser focus so Enter opens the selected app, restored browser focus after backdrop dismissal, added semantic focus/active styling, and used the display’s launch context. Added regression coverage and refreshed the reviewed context-menu baseline. Audited all added comments; only license headers remain.

Verified on a private display: Open With → Enter launches the selected file in all three views without changing MIME defaults; launch failures show the existing error dialog; Escape closes the chooser; directories omit the action.

Current chooser screenshot (committed with the review):
Open With chooser

wmfeht:
## Merge Blocker

The two GTK unit tests added in `38d34be` never actually run. They call `gtk_test` with short aliases (`open_with_empty_and_dismissal`, `open_with_preparation`) instead of the full rustc paths every other caller uses, so the child process matches **zero** tests, prints `running 0 tests`, and the parent still reports `ok` in ~0.03s.

That leaves empty-state Open / backdrop dismissal focus restore / URI preservation / stale-generation coverage unexecuted while CI stays green — including the dismissal-focus regression this commit claimed to lock in.

Fix: pass the full paths (`ui::open_with::tests::empty_chooser_disables_open_and_restores_focus_after_backdrop_dismissal` and `ui::browser::context_menu::tests::open_with::preparation_preserves_file_uris_and_rejects_incompatible_or_stale_queries`), then confirm those closures actually execute.

l0gicgate:
Thanks @wmfeht—confirmed. My short aliases caused zero-test child processes, so the earlier GTK coverage claim was incorrect. Fixed both calls to use their full rustc paths in 7780c30.

Verified on a private Xvfb display: each child now reports `running 1 test`. I temporarily inserted a failing assertion inside each closure and confirmed both failures propagated to the parent, then removed those probes and reran successfully. The complete local gates also pass, including the 334-scenario canonical container suite. This follow-up changes only test dispatch names and rustfmt indentation.

Issue #354, opened 2026-09-05
fix(file-operations): make rename and copy/paste workflows consistent


l0gicgate:
## Strata version

Not provided in the original Discord report; needs confirmation.

## Installation method

Not provided; needs confirmation.

## Linux distribution and desktop environment

Not provided. The screenshots appear to use an Omarchy-style environment, but the exact distribution, compositor, and versions need confirmation.

## What happened?

Several rough edges were reported around inline rename and copy/paste workflows:

1. When renaming a long filename, the entry does not horizontally scroll to keep the caret/end of the name visible.
2. After deleting files, copying files, and pressing Undo, Strata restores the previously deleted files instead of undoing the newer copy operation.
3. After copying into the same folder, the newly created copy is not selected.
4. The **X** close button on the “File already exists” conflict dialog appears to do nothing.
5. Copy/paste behavior appears inconsistent when the destination is the current folder versus a different folder.

These behaviors make it difficult to see what is being renamed and to predict the result of copy, conflict-resolution, selection, and undo actions.

### Long inline rename

1. Select a file whose name is wider than the available column.
2. Start inline rename.
3. Move the caret to, or begin editing at, the end of the filename.
4. Observe that the entry remains scrolled toward the beginning and the caret/edit is not visible.

### Copy followed by Undo

1. Delete one or more files using an undoable operation.
2. Copy and paste other files.
3. Press <kbd>Ctrl</kbd>+<kbd>Z</kbd>.
4. Observe that the deleted files are restored instead of the latest copy being removed.

### Same-folder copy and conflict handling

1. Copy a file and paste it into its current folder.
2. Observe whether the generated copy becomes selected.
3. Repeat until the “File already exists” conflict dialog appears.
4. Click the dialog's **X** close button.
5. Observe that it does not dismiss/cancel the conflict flow.

The exact sequence demonstrating the current-folder/different-folder inconsistency needs to be recovered from the original recording or reproduced during triage.

## Expected behavior

- Inline rename keeps the caret visible, including at the end of long filenames, without changing the intended stem-only initial selection.
- Undo targets the most recent completed file operation. A completed copy should be undoable by removing only the copies it created, rather than exposing an older trash undo as though it were the latest operation.
- A successful copy selects and reveals the newly created destination item when its destination is visible.
- Closing the conflict dialog with **X** behaves like Cancel: dismiss the dialog safely and do not overwrite anything.
[...]

PR #443, opened 2026-09-06
fix(rename): hide the columns size badge while renaming

JoeJoeflyn:
## Description

In Columns view the per-row size badge (`0 B`, `12 KB`, …) is an overlay positioned at the row's trailing edge. During an inline rename of a long file name, the badge sat on top of the rename entry's tail, hiding the end of the name and the file extension exactly where the user needs to see what they are editing (#354, point 1). GTK already scrolls the caret into view; the badge covering it made the field look broken.

`ActiveRename` now also holds the row's size label: `begin_rename` hides it for the duration of the edit, `cancel_rename` restores it, and the row rebind already restores it after a successful rename (columns/rows.rs binds `size.set_visible(!size_text.is_empty())` on every bind).

Grid and List views have no such overlay, so they needed no change.

## Visual evidence

Before:

https://github.com/user-attachments/assets/5eafe013-5e77-47f1-90da-182515a6efd2

After:

https://github.com/user-attachments/assets/e3815ff4-b290-42e4-afbd-cc3392a5fba0

## How to test

1. In Columns view, open a folder containing a file whose name is much longer than the column width (e.g. `this-is-a-deliberately-very-long-file-name-to-test-inline-rename-scrolling-behavior-in-strata.txt`).
2. Select the file and press F2 to start the inline rename.
3. Move the caret to the end of the name (End key).

**Expected result:** the entry scrolls to keep the caret visible and the size badge is hidden while renaming, so the end of the name and the extension are fully readable. Escape or submitting the rename brings the badge back.

## Related issue

Partially addresses #354 (point 1: long-name rename visibility).


l0gicgate:
Review follow-up:

- Prevented late size-metadata updates from revealing the badge again while a Columns rename is active.
- Restored the badge on cancel only when it has content, preserving hidden badges for folders or unavailable metadata.
- Added GTK coverage for the long-name rename lifecycle.

Manual verification: exercised a constrained Columns pane under isolated Xvfb, opened a long filename with F2, moved to the end, confirmed the extension/caret remained unobscured, and confirmed Escape restored the size badge. The existing PR after-image remains representative. I also captured a current image, but the available GitHub CLI cannot upload issue-comment attachments; it is available locally at `/tmp/strata-pr-443-review/long-columns-rename.png`.

PR #444, opened 2026-09-06
fix(operations): make copy participate in latest-operation undo

### LINKED ISSUE #390: fix(file-operations): make copy participate in latest-operation undo
Split from the grouped rename and copy/paste report ([#354](https://github.com/lgse/strata/issues/354), item 2 of 5).

## Strata version

Reproduced on source build at commit `183d744` (v0.10.0, debug build from `wmfeht/strata`).

## What happened?

After deleting files with an undoable trash operation, then copying other files, `Ctrl+Z` restores the previously deleted files instead of undoing the copy.

Copy does not replace the latest reversible operation, so an older trash undo is still treated as current. Users cannot reverse a completed copy, and they can accidentally restore unrelated deleted items.

## Steps to reproduce

1. In a writable test directory, delete one or more files with the undoable trash operation (Delete, not Shift+Delete).
2. Select different files, copy them (`Ctrl+C`), and paste (`Ctrl+V`) so the copy completes (same-folder paste creating `name (1).ext` is fine).
3. Press `Ctrl+Z`.
4. Observe that the deleted files are restored from Trash instead of the copies created in step 2 being removed.

## Expected behavior

Undo targets the most recent completed file operation.

A completed copy must participate in that global latest-operation ordering. Undoing it removes only the copies that operation created — including same-folder generated names and conflict-resolved destination names — and does not restore, overwrite, or otherwise touch unrelated items.

Trash undo remains available afterward if it is still the previous reversible operation.

## Screenshots or video

Repro recording (trash → copy/paste → Ctrl+Z restores trash, duplicate remains):

https://cursor.com/artifacts/v/art-bb08bf82-f002-4edd-805f-7b18c9fbb42b

## Additional context

Undo is a single latest entry (`PENDING_UNDO`). There is no `Copy` variant; `replace_pending_undo` is called for trash and for moves that left their directory. Copies never replace it.

On a copy paste, `moved_locations` stays empty when `moving != Some(true)`, so `replace_pending_undo` returns early and the prior `UndoEntry::Trash` remains the Ctrl+Z target.

### Acceptance criteria

- [ ] A completed copy becomes the latest undoable operation and displaces an older trash or move undo.
- [ ] `Ctrl+Z` after that copy removes only the copies it created, including same-folder generated names and conflict-resolved destination names.
- [ ] Original source items are left in place. Unrelated trash/move undo entries are not applied as if they were latest.
- [ ] After the copy undo completes, the previous reversible operation (if any) is available again.
- [ ] Automated tests cover copy undo ordering, same-folder generated names, and conflict-resolved destinations.

spandan11106:
## Description

Undo kept a single latest entry, and a completed copy never replaced it. `moved_locations`
stayed empty whenever `moving != Some(true)`, so `replace_pending_undo` returned early and an
older `UndoEntry::Trash` stayed the Ctrl+Z target: a copy could not be reversed, and undoing
after one restored unrelated deleted items.

Paste now reports the destination it writes for each item through `TransferProgress`, the same
way deletion and restoration already report per-item locations. The browser accumulates those
for the running operation, so a completed, failed, or cancelled copy records an
`UndoEntry::Copy` holding exactly what it created — the adapter's own generated name for a
same-folder duplicate, and the resolved destination when a conflict was answered with Replace.
Undoing it trashes only those destinations; nothing derives a destination path a second time,
so a generated or conflict-resolved name can never be missed or guessed wrong.

Undoing a copy sends the copies to Trash rather than erasing them, so pressing Ctrl+Z can never
destroy data outright, and that internal deletion deliberately records no undo entry of its own.

The single pending-undo slot becomes a bounded history (32 entries), because a copy undo has to
leave the operation before it reachable. Undoing the latest entry now pops it and exposes the
previous one instead of clearing the slot; an entry left with nothing to retry is dropped for the
same reason. Deletion is extracted into one pass shared by delete and copy undo, since copy undo
starts from locations rather than `FileEntry` values.

Addressing the issue's acceptance criteria:

- A completed copy becomes the latest undoable operation and displaces an older trash or move undo.
- Ctrl+Z after that copy removes only the copies it created, including same-folder generated names and conflict-resolved destination names.
- Original source items are left in place, and unrelated trash or move entries are never applied as if they were latest — a copy that created nothing (a no-op paste) leaves the previous entry current rather than presenting it as the copy's undo.
- After the copy undo completes, the previous reversible operation is available again.
- Automated tests cover copy undo ordering, same-folder generated names, conflict-resolved destinations, partial and cancelled copies, and the history bound.

## Visual evidence

N/A — nothing in the interface changes. The change is only in which items an undo removes or
restores, which the reproduction steps below show directly in the file view.

## How to test

1. In a writable directory containing at least two files, select one and press <kbd>Delete</kbd> to trash it.
2. Select a different file, press <kbd>Ctrl</kbd>+<kbd>C</kbd>, then <kbd>Ctrl</kbd>+<kbd>V</kbd> in the same folder to create a `name (1).ext` duplicate.
3. Press <kbd>Ctrl</kbd>+<kbd>Z</kbd>.
4. Press <kbd>Ctrl</kbd>+<kbd>Z</kbd> again.
5. Repeat steps 1–3 with the paste targeted at a different folder, and again after answering a "File already exists" conflict with Replace.

**Expected result:** step 3 removes only the duplicate the copy created — the copy source stays
put and the trashed file stays in Trash. Step 4 then restores the trashed file, because the
trash operation is the previous reversible one again. The cross-folder and conflict-resolved
pastes behave the same way: undo removes just the destination that paste wrote.

## Related issue

Closes #390


spandan11106:
https://github.com/user-attachments/assets/5666df22-1579-48af-9341-b45589f2e6a9

l0gicgate (dismissed):
Reviewed the complete diff after merging the latest main. Copy undo records the adapter-resolved destinations, safely trashes only completed copies, preserves earlier undo history, and is ready to merge.

l0gicgate (approved):
Re-reviewed the cleanup at 8cea53c. The redundant comments are removed, the safety invariant remains documented, and all required checks pass.

l0gicgate:
Review follow-up: removed 28 lines of redundant implementation/doc comments that paraphrased nearby fields, methods, control flow, and tests. Kept only the concise safety invariant that copy undo uses Trash rather than permanent deletion.

Manual verification remains unchanged: same-folder copy undo removed only the duplicate, and a second undo restored the preceding trashed file.

PR #445, opened 2026-09-06
fix(rename): flag invalid names instead of aborting

### LINKED ISSUE #428: fix(rename): invalid names crash Strata via RefCell panic
### Strata version

v0.11.1 (repro on `a96c4a0`; upstream `main` since moved)

### Installation method

Built from source

Ubuntu 24.04.4 LTS, X11, GTK 4.14.5 (isolated temp HOME/XDG; synthetic names only)

### What happened?

Renaming a file to an invalid basename (empty name, a name containing `/`, or `.`) and confirming with Enter aborts the whole process.

`thread 'main' panicked at src/ui/browser.rs: … RefCell already mutably borrowed` → process abort. Files are left untouched. Collision renames (async path) still show the red field instead of crashing.

### Steps to reproduce

1. Open any view (Columns confirmed; List/Icons share the path).
2. Select a file → F2.
3. Clear the name (or type `bad/name`, or `.`) → Enter.
4. Observe the process panic/abort.

### Expected behavior

Inline validation (“Enter a name” / “Names cannot contain /”) without crashing, as basename validation already intends.

### Screenshots or video

https://cursor.com/artifacts/v/art-0d144783-77b4-4906-8eb5-89101e59c1c1

### Relevant logs

Panic: `RefCell already mutably borrowed` in `ViewState::submit_rename` while holding `active_rename.borrow_mut()` and handling synchronous `RenameFailed` from `validate_basename`.

### Additional context

Found in Fable 5.1 exploratory QA. Same `submit_rename` is shared by all views. Collisions (async provider) are fine.

spandan11106:
## Description

Renaming a file to an invalid basename in the Columns view aborted the whole process instead of
rejecting the name.

`ViewState::submit_rename` held `active_rename.borrow_mut()` while calling `Browser::rename`.
`Browser::rename` validates the basename itself and reports a bad one by emitting `RenameFailed`
**before it returns**, and that handler borrows `active_rename` again to flag the field — so every
synchronous rejection hit `RefCell already mutably borrowed` and aborted (a panic across a GTK
callback cannot unwind).

The fix reads the renamed entry in a borrow scoped to end before the call. The failure then lands
where it was always meant to: the inline field turns red, keeps focus, stays editable, and its
tooltip carries the reason. No other behaviour changes.

Covered by all three names the issue reports:

| Typed name | Result |
| --- | --- |
| *(empty)* | "Enter a name" |
| `bad/name` | "Names cannot contain /" |
| `.` | "That name is reserved" |

One note on the report: only Columns was affected. List and Icons drive the rename from
`ModeViews`, which never holds the borrow across the call — verified by running the new test
against them with the fix reverted. The regression test covers all three modes anyway, since the
crash was one shared handler away from them.

## Visual evidence

Screenshots cannot be attached from the CLI, so they are described here; the new test renders the
exact states below.

**Before:** nothing to capture — the process aborts on Enter, the window disappears, and the
terminal prints `thread 'main' panicked at src/ui/browser/events.rs:381: RefCell already mutably
borrowed` / `thread caused non-unwinding panic. aborting.`

**After:** the row stays in place with the inline rename field still open, the entry drawn in the
theme's error state (red border and text) around the rejected text, focus back in the field, and a
tooltip giving the reason. Same in Columns, List, and Icons.

## How to test

1. Build and run: `cargo run`.
2. Select any file in the Columns view and press <kbd>F2</kbd>.
3. Clear the name and press <kbd>Enter</kbd>.
4. Repeat with `bad/name` and with `.`, and in the List and Icons views.

**Expected result:** Strata stays running. Each invalid name leaves the rename field open and
flagged red with a tooltip explaining the problem ("Enter a name", "Names cannot contain /", "That
name is reserved"), the field stays editable and focused so the name can be corrected, and the file
on disk is untouched. A valid name still renames, and re-submitting the unchanged name still just
closes the editor.

## Related issue

Closes #428


l0gicgate (dismissed):
Reviewed against the updated base. The scoped  borrow fixes the synchronous rename-failure re-entry, and the regression test covers invalid names across Columns, List, and Icons. Full local formatting, lint, and isolated-Xvfb test suites pass.

l0gicgate:
Follow-up review fix: opening the inline rename editor in Icons view no longer changes the grid cell's requested size and reflows neighboring cards. The label and editor now share a fixed overlay slot, and the editor expands within the existing card width.

The mapped GTK regression test now records every icon card's bounds before rename and verifies they remain identical after the editor opens, while retaining the invalid-name coverage across all three modes. I also reviewed the supplied manual reproduction recording at `/home/l0gicgate/Videos/screenrecording-2026-09-06_09-04-31.mp4`. GitHub CLI does not provide attachment upload support here, so the recording could not be uploaded to this comment.

l0gicgate (dismissed):
Re-reviewed after the Icons layout follow-up. The rename field now stays within the card name slot, and the mapped GTK test verifies that opening it preserves all card bounds. Full local formatting, lint, and isolated-Xvfb test suites pass.

l0gicgate:
Fixed the startup-only preview regression found during manual review. The main browser previously kept its default single-click-preview state until the lazily built Settings page applied the saved preference, so the first interaction could preview despite the option being disabled. Window construction now applies the saved preference before navigation or clicks are possible, with an isolated GTK regression test covering the disabled startup state.

l0gicgate (dismissed):
Re-reviewed after the startup preference fix. The saved single-click preview preference is now applied during main browser construction, before any click can be handled, and the isolated GTK regression test covers disabled startup behavior. Full local checks pass.

l0gicgate (dismissed):
Re-reviewed after merging the latest main and resolving the accessibility and click-activation conflicts while preserving both sides. The complete local formatting, Clippy, and isolated-Xvfb test suites pass.

l0gicgate:
Resolved the E2E failure from run 34041782795. It was a strict XPASS: the startup preference fix also corrected the known pointer-selection focus behavior, so Space now opens quick preview on the first press. Removed the obsolete expected-failure marker; the full local E2E suite passes (183/183).

l0gicgate (approved):
Re-reviewed after removing the obsolete strict xfail. The previously expected pointer-preview failure now passes because the startup preference fix corrects the behavior. Full local and GitHub E2E suites pass.

PR #453, opened 2026-09-06
fix(browser): center file row labels with their icons

### LINKED ISSUE #452: bug(browser): file row labels are not vertically centered with their icons
### Strata version

0.10.0 (built from source, commit c19f8b7)

### Installation method

Built from source

Arch Linux (Omarchy), Hyprland 0.56.2

### What happened?

In the file browser pane (list and Explorer/miller-column views), the file/folder name label sits noticeably above the vertical center of its row instead of lining up with the icon and the disclosure chevron next to it. It's most visible when the pane is showing a tall listing (e.g. the Home directory) with many single-line rows in a row.

### Steps to reproduce

1. Open Strata to any folder with several files/subfolders (e.g. the Home directory).
2. Look at a row in the file list: the folder/file icon, the name label, and the chevron on the right.
3. Observe that the name label appears shifted upward relative to the icon and chevron, rather than vertically centered with them.

### Expected behavior

The name label should be vertically centered in the row, aligned with the icon and chevron.

### Screenshots or video

Reporter supplied before/after screenshots during triage (not attached here); the misalignment is visible in any populated folder listing at default zoom.

### Additional context

This appears to be a regression from the search feature work that wrapped the row's label in an extra `Overlay` + vertical `Box` (to host the search-result path line), which changed the row's internal size negotiation and left the label's containing widgets defaulting to top alignment instead of centering.

errpoulos:
## Description

File row labels (in the list and Explorer/miller-column views) were sitting above the vertical center of the row instead of aligning with the icon and chevron next to them. The label lives inside an `Overlay` + vertical `Box` that was added to host the search-result path line; that wrapper defaulted to top alignment instead of centering within the row's height. Centered the icon, the overlay, and the chevron explicitly.

## Visual evidence

Before: the folder/file name sat visibly above the icon's vertical center in every populated listing (e.g. the Home directory).
After: the name lines up with the icon and chevron in the same row.

## How to test

1. Build and run Strata from this branch.
2. Open a folder with several files/subfolders (e.g. your Home directory) in list or Explorer view.
3. Look at any row: the icon, name label, and chevron.

**Expected result:** the name label is vertically centered with the icon and chevron, not shifted upward.

## Related issue

Closes #452


l0gicgate:
@errpoulos please attach screenshots/videos when making UI changes.

Thank you for this contribution

l0gicgate:
I am on `main` and I can't repro this. The labels are vertically aligned:
image
image

l0gicgate:
Thanks for working on this. I am having trouble reproducing the reported misalignment. By “row labels,” I assume we mean filename text such as **Videos** and **Documents** beside the icon.

I built and captured both the exact reported base revision (`c19f8b7`) and this PR's head (`c2b114c`) in isolated Xvfb sessions, using the same fixture, window size, and default settings. The resulting 1280×900 captures were pixel-identical:

```
ImageMagick compare -metric AE: 0 (0)
base SHA-256: b9d9f05a4a181ad5a0557ce5cc596b3cf2a313b1ff5d09123b3cf66187eaa631
PR   SHA-256: b9d9f05a4a181ad5a0557ce5cc596b3cf2a313b1ff5d09123b3cf66187eaa631
```

I also cannot see an offset on current `main`; the filename text appears vertically centered with the icon and chevron. In the documented default Explorer-view scenario, the three explicit `set_valign(gtk::Align::Center)` calls therefore appear to be a layout no-op on this system.

Could you please attach the requested before/after screenshots and provide the exact conditions that expose the difference—view mode, theme, text size, density, display scale, and GTK version? Marking the offset in the screenshots would also help. The branch currently conflicts with `main` because the row implementation moved to `src/ui/browser/columns/rows.rs`, so we should first establish a reproducible case before asking you to resolve that conflict.

l0gicgate:
I am closing this for now. Feel free to reopen if you think there's still something to fix.

PR #454, opened 2026-09-06
fix(browser): paste into a single selected folder

### LINKED ISSUE #449: feat(browser): paste into a single selected folder
## Summary

Selecting a folder and pressing Ctrl+V pastes into the visible column instead of the selected folder. The selection highlights a destination the paste ignores.

## Current behavior

1. Select a file, Ctrl+C.
2. Single-click a folder (selection highlight, no descent).
3. Ctrl+V.
4. The copy lands in the folder's parent column, next to the folder.

Related wrinkle: clicking a folder that opens a child column also keeps the paste destination on the parent column, because the mouse stays parked there while the pointer owns navigation.

## Expected behavior

- Exactly one selected directory receives the paste itself.
- Files selected, multiple items selected, or no selection: paste into the destination column, as today.
- A click that opens a child column moves the paste destination to that column until the pointer navigates elsewhere.

This deliberately diverges from Nautilus/Thunar/Finder, where Ctrl+V always pastes into the current view and "Paste Into" is a separate context action. Here the single-folder selection is the destination.

JoeJoeflyn:
## Description

Selecting a folder and pressing Ctrl+V pasted into the visible column instead of the selected folder — the selection highlighted a destination the paste ignored (#449).

When exactly one directory is selected, paste now targets it. Files selected, multiple items, or no selection keep pasting into the destination column. A pointer click that opens a child column also moves the paste destination to that column (the mouse stays parked on the parent while the pointer owns navigation, so the hover used to keep the target stuck one column up).

This deliberately diverges from Nautilus/Thunar/Finder, where Ctrl+V always pastes into the current view; here the single-folder selection is the destination.

## Visual evidence

https://github.com/user-attachments/assets/00f9e4b2-516f-4dfe-99de-b3ba4d6a5059

## How to test

1. Create `source.txt` and an empty folder `inside/`.
2. Select `source.txt`, Ctrl+C.
3. Single-click `inside/` so it is highlighted (do not descend into it).
4. Ctrl+V.

**Expected result:** `source.txt` appears inside `inside/`. Repeat with a file selected, multiple items, or nothing selected — the paste lands in the current column. Click a folder to open its child column, then Ctrl+V without moving the mouse — the paste lands in the opened column, not its parent.

## Related issue

Closes #449


l0gicgate:
Review fixes pushed:

- Re-resolve pointer ownership from each click and only advance a parked-pointer paste target when the opened column is the hovered column’s child. This prevents the synthetic child target from remaining stale after a later click and avoids redirecting unrelated column additions.
- Extracted and unit-tested selected-folder destination resolution, including file, multi-selection, empty-selection, and missing-column fallbacks.
- Added headless end-to-end coverage for pasting into a selected folder in all three views and for Ctrl+V immediately after a pointer opens a child column.

Manual verification: the focused E2E scenarios passed (4 tests) on a private Xvfb display. Screenshots are N/A because this changes paste routing without a visual delta.

PR #464, opened 2026-09-06
fix(dialogs): submit modal forms with Enter

### LINKED ISSUE #439: Enter key does not submit modal forms consistently
### Strata version

Source build at commit `fb62b70f9c3c47c902f35b0f78d1a670121517f4` (PR #410 head).

### Installation method

Built from source

Omarchy, Hyprland 0.56.2

### What happened?

Pressing Enter while editing a single-line field does not consistently activate the form’s primary action. This was confirmed in the compression dialog and may affect SMB and other modal forms. Users must move to or click the submit button even when the form is valid.

### Steps to reproduce

1. Open the Compress dialog for an item.
2. Focus the Archive name field and enter a valid name.
3. Press Enter.
4. Observe that compression is not submitted.
5. Audit the SMB connection and other modal forms for the same behavior.

### Expected behavior

Enter in a single-line form field should invoke the enabled primary action, matching a click on that button. Invalid forms should remain open and show their existing validation state. Multiline inputs should retain newline behavior where applicable.

### Screenshots or video

N/A; keyboard interaction issue.

### Relevant logs

No relevant errors.

### Additional context

The fix should use a consistent form-level/default-action pattern and include coverage for each affected dialog rather than patching only the compression field.

spandan11106:
## Description

Enter in a single-line modal field did not reliably invoke the form's primary action. The archive dialogs built by `build_archive_modal` connected the confirm button with `connect_clicked` only and never touched their entries, so Enter fell through to `focus_navigation`, which skips activation while an editable has focus. `Copy to…` and the mount authentication dialog each worked around this separately with their own per-field `connect_activate` handlers.

Rather than patch the compression field, this adds one default-action pattern in `src/ui/modal.rs` and routes every modal form through it:

- `submit_on_enter(fields, confirm)` walks a form's widget tree and connects `activate` on each single-line field (`gtk::Entry`, `gtk::PasswordEntry`) to the primary action, the way a `GtkWindow`'s default widget behaves. The overlay modals are not `GtkWindow`s, so `set_default_widget` is not available to them.
- Multi-line inputs are not descended into as fields, so a `GtkTextView` keeps its newline.
- The action fires only while it is sensitive, so a form mid-operation (for example "Creating folder…") stays inert.
- Validation is untouched: an invalid form still keeps the dialog open and shows its existing error state.

Dialogs covered by the shared pattern:

| Dialog | Before | After |
| --- | --- | --- |
| Compress (name, password, confirm password) | Enter did nothing | Enter runs **Compress** |
| Extract to (destination) | Enter did nothing | Enter runs **Extract here** |
| Extract password (password) | Enter did nothing | Enter runs **Extract** |
| Copy to / Move to (destination) | worked via an ad-hoc handler | same behavior, now the shared pattern, and inert while the folder is being created |
| Mount authentication (username, domain, password) | worked via three ad-hoc handlers | same behavior, now the shared pattern |

Coverage added rather than just the one field:

- Unit test for the pattern itself: a nested password field submits, a `GtkTextView` is not treated as a field, and a disabled primary action stays inert.
- End-to-end scenarios driving real keys: Enter submits Compress, Enter submits Extract to, Enter submits Copy to (regression guard for the dialog that already worked), and an invalid archive name keeps the Compress dialog open without producing an archive.

## Visual evidence

N/A — keyboard interaction only; no pixels change. The behavior is asserted by the end-to-end scenarios listed above.

## How to test

1. Right-click a file → **Compress…**, type a valid archive name, press Enter. The archive is created and the dialog closes.
2. Repeat with an invalid name such as `../escape` and press Enter. The dialog stays open and the field shows its validation state; no archive is written.
3. Right-click that archive → **Extract to…**, type a destination folder, press Enter. The archive is extracted there.
4. Right-click a file → **Copy to…**, type a destination folder, press Enter. The file is copied.
5. Open a password-protected archive so the password prompt appears, type the password, press Enter.

**Expected result:** Enter in each single-line field does exactly what clicking that dialog's primary button does; invalid forms stay open with their validation state intact.

## Related issue

Closes #439

### DIFF
--- a/src/ui/browser/archive.rs
+++ b/src/ui/browser/archive.rs
@@ -14,7 +14,9 @@ use crate::ui::controls::{
     ModalTone, form_entry, form_label, form_password_entry, message_dialog_description,
     message_dialog_layout, modal_layout, segmented_control,
 };
-use crate::ui::modal::{ModalHost, dismiss_modal_layer, modal_layer, show_error_dialog};
+use crate::ui::modal::{
+    ModalHost, dismiss_modal_layer, modal_layer, show_error_dialog, submit_on_enter,
+};
 use gtk::prelude::*;
 use gtk::{gio, glib};
 use std::cell::Cell;
@@ -344,6 +346,7 @@ impl ViewState {
                 );
             }
         });
+        submit_on_enter(&body, &confirm);
         name_entry.grab_focus();
     }
 
@@ -460,6 +463,7 @@ impl ViewState {
             dismiss_for_confirm();
         });
 
+        submit_on_enter(&body, &confirm);
         field.grab_focus();
     }
 
@@ -491,6 +495,7 @@ impl ViewState {
             dismiss_for_confirm();
             browser.extract(entry.clone(), destination.clone(), password);
         });
+        submit_on_enter(&body, &confirm);
         password_entry.grab_focus();
     }
 }
--- a/src/ui/browser/location.rs
+++ b/src/ui/browser/location.rs
@@ -11,7 +11,9 @@ use crate::ui::browser::{BrowserView, ViewState};
 use crate::ui::controls::{
     form_entry, form_label, form_password_entry, modal_layout, segmented_control, wrap_dialog_text,
 };
-use crate::ui::modal::{ModalHost, dismiss_modal_layer, modal_layer, show_error_dialog};
+use crate::ui::modal::{
+    ModalHost, dismiss_modal_layer, modal_layer, show_error_dialog, submit_on_enter,
+};
 use gtk::prelude::*;
 use gtk::{gio, glib};
 use std::cell::{Cell, RefCell};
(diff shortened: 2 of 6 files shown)


l0gicgate:
Review follow-up: removed the verbose `submit_on_enter` doc comment because it restated the helper’s implementation and call-site behavior; the code and focused tests communicate the contract directly.

Manual verification: ran the four affected headless GUI scenarios with real Return key input (Compress, invalid Compress validation, Extract to, and Copy to); all passed. This keyboard-only cleanup has no visual change, so screenshots remain N/A.

l0gicgate (approved):
Reviewed the updated diff and found no merge-blocking issues. The shared Enter submission behavior, disabled-action guard, validation paths, and coverage are sound; required local checks and the affected headless GUI scenarios pass.

PR #473, opened 2026-09-06
perf(search): reduce query latency and index memory

### LINKED ISSUE #471: perf(search): reduce recursive search latency and memory
### Problem or use case

Recursive search can retain hundreds of thousands of path records, rebuild indexes for each search surface, and rescore broad queries slowly. Search should remain responsive without raising its RAM ceiling.

### Proposed solution

Keep the 200,000-entry safety cap, prune only high-volume generated/cache content, avoid repeated normalization and path-depth work while scoring, and retain only the best 100 matches with minimal cloning. Add regression coverage and benchmark indexing and query latency.

### Alternatives considered

A 400,000-entry cap improves coverage after pruning but approximately doubles worst-case memory and query work. A persistent shared index may be valuable later but is a larger lifecycle and invalidation change.

### Visual examples

N/A — non-visual performance work.

### Strata version

main at 1a35288

### Additional context

The implementation should preserve hidden-file search semantics and report partial results accurately.

JoeJoeflyn:
## Description

Keep recursive search at its 200,000-entry RAM ceiling while pruning high-volume generated trees without hiding tool configuration. Cache normalized query data and path depth, remove the duplicate normalized-name allocation, and select the best 100 completed-index matches with a bounded heap.

A 200,000-entry synthetic benchmark reduced a broad query from about 94 ms to 10 ms. Each retained index item also drops from 104 to 88 inline bytes plus one fewer allocated string.

## Visual evidence

N/A — search behavior is unchanged; this is non-visual performance work.

## How to test

1. Open recursive search in a large home or project tree and enter both broad and exact queries.
2. Enable hidden files and verify tool configuration such as `.cargo/config.toml` remains searchable while generated registry/build content is omitted.

**Expected result:** Search remains responsive, returns the best 100 matches, preserves tool configuration results, and stays within the existing index-entry limit.

## Related issue

Closes #471


JoeJoeflyn:
### Before/after search benchmark

Release-mode A/B benchmark using the exact `src/services/search.rs` from `origin/main` and this PR. The fixture contained 199,000 empty files across 199 directories; each query was run 14 times across alternating before/after rounds, with the median reported.

| Operation/query | Before | After | Change |
|---|---:|---:|---:|
| Build index | 473.69 ms | 474.03 ms | unchanged |
| Broad substring (`ordinary`) | 86.61 ms | 11.77 ms | **7.4× faster** |
| Selective substring (`199998`) | 26.92 ms | 26.85 ms | unchanged |
| Fuzzy (`of199`) | 30.98 ms | 32.02 ms | unchanged/noise |
| Miss (`zzzzzz`) | 25.30 ms | 25.53 ms | unchanged |

This optimization primarily removes repeated depth work and cloning for broad result sets. Generated-tree pruning improves real home/project indexing when those trees would otherwise consume the entry/time budget; the controlled fixture intentionally contained no generated directories.

JoeJoeflyn:
### Final shared-index benchmark

Release-mode, 199,000-file fixture.

- Query latency: broad **93.64 ms → 7.80 ms**, selective **31.92 ms → 9.15 ms**, fuzzy **34.54 ms → 9.53 ms**, miss **26.27 ms → 8.84 ms**.
- Two simultaneous same-root searches: RSS **113,252 KB → 58,760 KB** (about **54 MB / 48% less**).
- Second same-root index: **889.61 ms → 0.13 ms**.
- Real Home query `kaze`: first result, ranked first, in **~51 ms**.
- Index cap remains **200,000 entries**.

The full local suite passes under private Xvfb: 830 passed, 18 ignored.

l0gicgate:
Reviewed and updated this branch against current `main`.

- Resolved the search conflicts while preserving multi-drive fairness, nested-mount boundaries, and detailed partial-coverage reporting.
- Fixed a completion/result-snapshot race and inconsistent ordering of tied results between incremental and completed queries.
- Restricted generated-tree exclusions to directories so ordinary files named `target`, `node_modules`, or `.cache` are not hidden.
- Added regressions for shared-index retirement, final batches, priority traversal, parallel ranking, and Unicode paths. Audited all added/modified comments.

Manual verification in a private Xvfb session: searching `target` finds the ordinary file; searching `needle` with hidden files enabled finds `.cargo/config-needle.toml` but not the registry fixture; Escape restores the listing. No layout changes.

<details>
<summary>Current verification screenshots</summary>

Ordinary target file remains searchable

Cargo configuration is retained while generated registry contents are pruned

</details>

l0gicgate:
### Post-review performance verification

**The query-speed and shared-index memory improvements are retained.** I also measured a first-index slowdown versus OP's final version; details below.

Release-mode comparison of the **unmodified search service** at four commits, using identical dependencies and build settings:

| Metric | Original base `1a35288` | Current main `cd11c8e` | OP final `14c2350` | Reviewed `ee81eab` |
|---|---:|---:|---:|---:|
| First index (ms) | 99.65 | 116.25 | 117.61 | **130.70** |
| Broad `ordinary` (ms) | 25.590 | 9.393 | 0.914 | **0.872** |
| Selective `199998` (ms) | 6.929 | 7.574 | 1.659 | **1.595** |
| Fuzzy `of199` (ms) | 8.032 | 7.636 | 1.649 | **1.594** |
| Miss `zzzzzz` (ms) | 7.249 | 7.471 | 1.613 | **1.711** |
| Second same-root index, first still alive (ms) | 100.496 | 116.217 | 0.034 | **0.036** |
| Two live sessions: process RSS (MiB) | 129.87 | 132.96 | 58.59 | **58.08** |
| Inline bytes per index item | 104 | 112 | 88 | **88** |

- Versus current main: **10.8× faster broad queries**, **4.4–4.8× faster other queries**, and **56.3% lower two-session RSS**. The second session still reuses the index.
- Versus OP final: broad/selective/fuzzy medians are similar or slightly faster; memory savings remain intact. **First indexing is 11.1% slower (+13.09 ms)** and **misses are 6.1% slower (+0.098 ms)** in this run. This is not a claim that every operation improved.

**Method:** Ryzen 9 9950X3D, Rust 1.98.0, release/LTO, warm tmpfs; 199,000 empty files across 199 directories, no generated trees. Four variants run in interleaved order for 14 fresh-process rounds each. Each process discards one query warm-up cycle and measures five cycles; table reports the median of the 14 run medians (70 samples/query/variant). Indexing and RSS have 14 observations each. Current main's search module is unchanged from the base merged during review.

Every completion was checked for full coverage; result counts and independent session queries were validated. `199998` returns one exact-substring match ranked first plus fuzzy matches. The **200,000-entry cap remains unchanged**.

These are service/API timings and standalone-process RSS, not GUI latency or whole-application RAM. This reconstructs OP's fixture scale/query categories, not their unpublished exact harness; different hardware/filesystem means absolute timings should not be compared directly with OP's posted numbers. Cold-disk and real-Home performance were not measured.

[Reproducible harness, pinned commits/dependencies, all raw measurements, and dispersion statistics](https://gist.github.com/l0gicgate/d5c1efafc83f215162d2e61546eddf8e)

PR #475, opened 2026-09-06
fix(updates): refresh opted-in portal after upgrade

### LINKED ISSUE #474: fix(updates): refresh the opted-in portal after in-place updates
### Strata version

0.11.0 upgrading to 0.11.2

### Installation method

GitHub release archive

Omarchy, Hyprland 0.56.2

### What happened?

An in-place update replaced the installed Strata executable while the opted-in FileChooser portal backend remained running from the deleted old executable. Image previews subsequently failed because the stale v0.11.0 backend attempted to start its sandbox helper from the deleted executable path. More generally, portal behavior and fixes remain stale until the portal frontend/backend is restarted.

### Steps to reproduce

1. Opt into Strata as the system FileChooser portal.
2. Start a portal file chooser so the Strata backend remains resident.
3. Install a Strata update in place.
4. Continue using or reopen a portal file chooser.
5. Observe that the old portal process remains resident; on affected older builds, sandboxed previews report “The sandboxed preview renderer failed.”

### Expected behavior

After a verified in-place update replaces Strata, detect whether the user has opted into the Strata FileChooser portal. If configured, refresh the portal frontend so the next chooser is served by the newly installed executable. Users who did not opt in must not have their portal services restarted.

### Screenshots or video

N/A — process/executable lifecycle issue.

### Relevant logs

```shell
$ readlink /proc/10366/exe
/home/user/.local/bin/strata (deleted)

$ strings /proc/10366/exe | grep -m1 0.11.0
0.11.0

$ strings ~/.local/bin/strata | grep -m1 0.11.2
0.11.2
```

### Additional context

The installed executable was replaced successfully. Linux correctly kept the old process image alive. v0.11.2 preserves a deleted running executable for preview helpers, but the long-lived portal still runs old application code until refreshed. The update path should reuse the existing opt-in status and portal refresh behavior without altering configuration.

l0gicgate:
## Description

Refresh an opted-in Strata FileChooser backend after an in-place update so it cannot remain attached to a deleted, outdated executable. A newly launched build also detects an older portal by executable identity and refreshes it, covering the first upgrade from a version that lacks the post-install hook. Portal services remain untouched when Strata is not the configured chooser or belongs to another installation.

## Visual evidence

N/A — process lifecycle fix with no visual changes.

## How to test

1. Opt into Strata as the system file chooser and open a portal dialog to activate the backend.
2. Replace the installed Strata executable with a newer build, then launch that newer build.
3. Open another portal dialog and preview an image.

**Expected result:** The stale backend is stopped, the next dialog runs from the installed executable inode, and its sandboxed image preview succeeds. Users who have not opted into Strata retain their existing portal processes.

## Related issue

Closes #474

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ Download the matching `strata-<version>-<target>.debug` asset from the same rele
 
 #### 3. Update or uninstall
 
-For a manual installation, use **Settings → Updates** for verified in-app updates, or repeat the download, verification, and `install` steps for a newer release. An in-app update also refreshes an already installed desktop entry and application icon from the new archive; it never creates desktop metadata that was not installed before. Package-managed installations are updated only by their system package manager. To remove a per-user installation:
+For a manual installation, use **Settings → Updates** for verified in-app updates, or repeat the download, verification, and `install` steps for a newer release. An in-app update also refreshes an already installed desktop entry and application icon from the new archive; it never creates desktop metadata that was not installed before. If the user opted into Strata's system file chooser, the update restarts the portal frontend so subsequent dialogs use the newly installed build. Package-managed installations are updated only by their system package manager. To remove a per-user installation:
 
 ```bash
 rm -f ~/.local/bin/strata \
--- a/src/main.rs
+++ b/src/main.rs
@@ -74,6 +74,9 @@ fn main() -> gtk::glib::ExitCode {
     if let Err(error) = tracing_subscriber::fmt::try_init() {
         eprintln!("Unable to initialize logging: {error}");
     }
+    if let Err(error) = portal_setup::refresh_stale_portal() {
+        tracing::warn!(%error, "could not refresh the stale Strata portal");
+    }
 
     // Timed from here so `window presented` covers the whole launch, the way
     // the field harness measures mapped from process start.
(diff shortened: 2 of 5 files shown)


l0gicgate:
I also staged the update lifecycle locally without changing the desktop’s installed portal configuration.

The fixture used a private D-Bus session, private XDG data/config/cache directories, an Xvfb display, and a disposable Strata installation under `~/.cache`. I replaced `systemctl` and `gdbus` with fixture commands so the test could record the requested service operations without restarting the live desktop portal.

Steps exercised:

1. Copy the built binary to a stable fixture path and install its portal metadata into the private XDG directories.
2. Start `strata --portal` from that path and record `/proc/<pid>/exe`’s inode.
3. Atomically replace the fixture executable with a different inode at the same path.
4. Launch the replacement Strata build.
5. Verify that the original portal process is stopped and that the refresh commands were requested.

Observed result:

```text
PASS: stale portal inode 10249889 was detected after replacement inode 10250190
PASS: opted-in backend stop and portal frontend restart were requested

Systemctl --user stop dbus-:*-org.freedesktop.impl.portal.desktop.strata@*.service
gdbus call --session --dest org.freedesktop.DBus --object-path /org/freedesktop/DBus --method org.freedesktop.DBus.ReloadConfig
systemctl --user restart xdg-desktop-portal.service
```

This validates the bootstrap case where the update is installed by a version that does not yet contain the post-install refresh hook: launching the newly installed build still detects and retires the stale portal.

PR #485, opened 2026-09-06
fix(browser): paste into the current directory after leaving a folder

ogarza:
## Description

After `#454`, Ctrl+V pastes into a single selected folder. Icons and List auto-select the first item on load (usually a folder), and List often selects the folder under the pointer because rows are full width. Copy or cut in a child, go to the parent, and paste went into that folder instead of the current directory.

Treat the post-load selection as a cursor until the user actually clicks or arrows to a folder. List folder activation now navigates in place, same as Icons.

## Visual evidence

N/A — paste destination only; nothing in the layout or styling changed.

## How to test

1. Open a directory that contains at least one folder and a child folder with a file in it.
2. Switch to Icons, copy or cut a file in the child, go to the parent (Alt+Up, Backspace, or the Up button), and paste without clicking another item.
3. Repeat in List.
4. In either mode, explicitly select a folder (click or arrow keys) and paste.

**Expected result:** Steps 2–3 land the file in the parent. Step 4 still pastes into the selected folder.

## Related issue

Closes #483

PR #499, opened 2026-09-06
fix(trash): gate actions by supported Trash operations

### LINKED ISSUE #433: fix(trash): hide unsupported context actions in Trash
### Strata version

v0.11.1 (repro on `a96c4a0`)

### Installation method

Built from source

Ubuntu 24.04.4 LTS, X11, GTK 4.14.5 (isolated temp HOME/XDG)

### What happened?

Trash item and blank-area context menus offer actions the backend cannot perform (Rename, Cut, Move to…, Compress…, Print; New Folder / New File on blank area). Each fails with “Operation not supported”. The Trash column also shows a “Keyboard · Paste here” destination footer.

### Steps to reproduce

1. Put a file in Trash and open Trash in Strata.
2. Right-click the item and try Rename / Cut / Compress….
3. Right-click blank area and try New Folder.
4. Observe unsupported errors.

### Expected behavior

Hide or disable actions that Trash cannot support (same spirit as capability-gating elsewhere).

### Screenshots or video

https://cursor.com/artifacts/v/art-97d6cbc7-cfc6-484c-a063-ac4559790eb7

### Relevant logs

No crash.

### Additional context

Adjacent to #367 (Empty Trash visibility) and #66 (capability-gate). Found in Fable exploratory QA.

kallupx:
## Description

Hide unsupported Trash actions in Columns, Icons, and List across single-item, multiple-item, and blank-area menus. Guard related keyboard entry points, remove misleading paste destination hints, and describe clipboard contents without promising paste availability.

Preserve Restore, permanent deletion, and Cut/Move to for top-level Trash items, as verified with isolated GVfs fixtures. Keep reading and copying, including text preview/Print; hide native-path-only preview/Print actions and mutations of children inside trashed folders. Normal-directory actions remain available.

## Visual evidence

**Columns, single item:** Rename and Compress are removed, along with the “Paste here” destination indicator. Supported actions remain.

| Before | After |
| --- | --- |
| Before: unsupported actions and paste destination in Trash | After: supported Trash actions only |

**List, blank area:** New Folder, New File, and unsupported destination actions are removed.

| Before | After |
| --- | --- |
| Before: creation actions in Trash | After: supported blank-area actions |

**Icons, multiple selection:** Restore, moving out, copying, and permanent deletion remain available.

After: multiple-item Trash menu in Icons

## How to test

1. Create some disposable files—a text file, image, archive, and folder containing files—and move them to Trash.
2. In Columns, Icons, and List, inspect single-item, multiple-item, and blank-area menus at the Trash root and inside the trashed folder.
3. Try F2, Ctrl+D, Ctrl+V, and Ctrl+Shift+N in Trash. Verify that Ctrl+X works for top-level items but is blocked for nested children.
4. Copy a file out, cut/paste a top-level item into a normal directory, restore another item, and permanently delete another. Preview a trashed text file.
5. Repeat menu and keyboard checks in a normal directory.

**Expected result:** Unsupported actions and paste destination hints are absent. Keyboard entry points follow the same restrictions. Supported reading, copying, top-level move/restore/delete, and normal-directory actions remain available.

## Related issue

Closes #433


spandan11106:
I pushed 7e25df8 to this branch with two fixes from my review, plus a regression test.

**Separator above Properties.** `archive_separator.set_visible(rename.is_visible() || can_compress)` never saw Rename as visible. gtk4-rs `WidgetExt::is_visible()` binds `gtk_widget_is_visible`, which requires the whole ancestor chain to be visible, and at that point the popover has not been popped up and `single` has not been shown yet — so the expression collapsed to `can_compress`. Ordinary local files masked it, since Rename and Compress are visible together. A location with no native path that is not Trash — a remote mount — offers Rename without Compress and lost the separator. Now taken from a `rename_visible` binding.

**Drag and drop in Icons and List.** The Columns column-background drop target is guarded in `install_directory_drop_target`, but `install_mode_directory_drop_target` (the Icons/List pane background) and the per-row folder drop targets in `columns/rows.rs` and `browser_modes.rs` still highlighted a trashed folder and accepted the drop. With the new `start_transfer` guard those drops became silent no-ops, so the UI invited an action and then did nothing. They now refuse Trash destinations and no drop-zone highlight appears. `is_trash_location` is `pub(in crate::ui)` so `browser_modes.rs` can reach it.

**Doc comment on `move_to_trash_is_visible`.** It says the fallback exists rather than "ever hiding the only delete option this menu has". That no longer holds for a child inside a trashed folder, which your test asserts — correctly, I think — offers neither delete nor restore. I noted the exception on the function instead of changing the behavior.

The new check rides along inside the existing menu test rather than a new one: two popover-driven GTK tests run concurrently and fight over the pointer grab, so one never maps its menu. I renamed `tests/trash.rs` to `tests/menus.rs` and the test to match, since it now covers a remote location as well. It fails on the old separator line and passes on the new one.

Local suite is green after the change — fmt, clippy, and the full test run under Xvfb.

One thing I left alone: `removable` and `restorable` infer capability from the URI shape, while `trash_visible` and `permanent_delete_visible` right below them fold in the resolved `can_trash_at` / `can_delete_at`. That reads fine for local GVfs Trash and I did not want to rework it here, but it is worth a thought.

l0gicgate:
Merged current `main`, preserving both load-cursor-aware paste behavior (#485) and Trash destination restrictions. Reviewed the complete diff and tightened the added comments.

Fixed two test races found during verification: GTK subprocesses now serialize access to their shared display (clipboard/pointer grabs), and optional menu capture waits for a render node instead of failing before the first frame.

The isolated GVfs fixture confirmed text preview/read, recursive copy, top-level move/restore/permanent deletion, and rejection of nested mutations. The menu/keyboard fixture passed across Columns, Icons, and List; I also inspected a captured Trash menu. No additional production UI changes in this review, so new visual evidence is N/A. Optional captures are local at `/tmp/pr499-menu-visuals/`; upload tooling is unavailable, and these were not posted.

l0gicgate:
Final status: all checks are green at `b487003`, including the end-to-end GUI suite on retry. The first GUI run passed 201 tests but failed four baselines solely because runner-specific numbered devices and a sidebar scrollbar appeared (0.561% pixel difference versus the 0.500% budget). I inspected the downloaded expected/actual/diff images; a failed-job rerun passed without code or baseline changes. No thresholds were weakened. This runner-dependent mount discovery remains a potential source of baseline flakiness.

Local fmt, Clippy, the full isolated-Xvfb suite (918 passed, 19 ignored), and the opt-in isolated GVfs fixture passed. The branch is clean and pushed; the PR has not been merged.

PR #502, opened 2026-09-06
fix(trash): validate and confirm restore destinations

### LINKED ISSUE #248: feat(transfer): default cross-volume drops to copy, with modifiers for move

### LINKED ISSUE #478: security(trash): Restore trusts trash::orig-path without validating destination
### Strata version

`main` / skills-branch review (source review on tracking fork; code path present in current tree)

### Installation method

Built from source

N/A (source / design review; affects any Linux install with Trash restore)

### What happened?

**Severity: High** (confirmed in security review)

Trash **Restore** takes the destination verbatim from the GVfs `trash::orig-path` attribute (the `Path=` line of a `.trashinfo` file) with **no validation**, then moves the trash entry there with **no overwrite check beyond “must not exist”** and **no user confirmation of the destination**.

An attacker who can write a trash directory the victim mounts (shared USB, writable network share, etc.) can plant:

- `.Trash-<uid>/info/<name>.trashinfo` with `Path=/home/victim/.config/autostart/payload.desktop` (or similar)
- matching payload under `files/`

When the victim clicks **Restore** in Strata’s trash view, the payload is moved to that attacker-chosen path. Targets that typically do not already exist (`~/.config/autostart/*.desktop`, `~/.bash_profile`, `~/.ssh/authorized_keys` when `~/.ssh` is absent) enable persistent code execution as the user.

### Steps to reproduce

1. Prepare a filesystem the victim will mount with write access for the attacker.
2. Under that mount, create `.Trash-<victim-uid>/files/` and `info/` with a crafted `.trashinfo` whose `Path=` points at a sensitive absolute path under the victim’s home (path must not already exist).
3. In Strata, open Trash, select the planted entry, choose **Restore** (no destination confirmation).
4. Observe the file appear at the attacker-chosen path.

### Expected behavior

- Validate restore targets: reject absolute `orig-path` values that violate the freedesktop trash rules for the trash topdir / home trash, and reject any path whose canonical form escapes the mount root hosting that trash directory.
- Defense in depth: confirm the **full restore destination** with the user before moving.

### Screenshots or video

Source review evidence (security-review skill, agent `bc-295dd478-2da2-477b-9f81-d852102b7138`):

- `src/adapters/local_operations.rs` (~2828): restore uses `trash::orig-path` → `gio::File::for_path` → `move_local` with no path validation.
- Reachable from `src/ui/browser/context_menu.rs` (~555) Restore with no destination confirmation.

### Additional context

- Finding ID from review: **VULN-001**
- Confidence: High (independently re-verified against source)
- Related hardening already present for copy/move/delete (`openat2` / symlink checks) does **not** cover this restore path.

wmfeht:
## Description

Validate untrusted Trash restore destinations and confirm their full paths before moving. Resolve physical GVfs items through `standard::target-uri`, handle both volume-trash layouts, reject traversal/symlink escapes and nested mount boundaries, and revalidate the confirmed destination at execution. Undo uses the same validator.

Restore retains kernel-enforced confinement and **atomic no-replace renames**. Unsupported filesystems now fail safely with the payload and metadata intact instead of falling back to an overwrite-prone rename. Mountinfo parsing preserves arbitrary path bytes. Destination lookup is bounded and cancellable; the unsafe live-Trash fixture utility has been removed.

This PR also retains the cross-device drop work from #338: **Always Ask** by default, configurable Copy/Move/Ask, Ctrl to copy, Shift to move, and unchanged Cut+Paste behavior. Unresolved volume lookups are described honestly. Mixed selections skip individual no-ops, and Enter activates the focused Copy, Move, or Cancel button.

## Visual evidence

Isolated GTK 4.14 widget captures of restore lookup feedback (synthetic files, not user Trash):

| Before: no lookup feedback | After: cancellable lookup |
| --- | --- |
| Before lookup feedback | Cancellable restore lookup |

Cross-device drop screenshots are also in the PR discussion; the original recording is on #338.

## How to test

Use a disposable session and expendable volume. See [restore safety and testing](https://github.com/lgse/strata/blob/d45602e89c7cf6fc5819b1eba64c969bf1aee01d/docs/trash-restore-testing.md) for details; do not use the removed fixture/cleanup utility against real Trash.

1. Trash a file or directory on the test volume, open Trash, choose Restore, and inspect the full destination. Exercise both `.Trash-$UID` and shared `.Trash/$UID`, nested paths, spaces, and duplicate display names. Cancel during lookup/confirmation, then retry and confirm.
2. Change a test item's `Path=` to another filesystem, change it after confirmation is presented, or create a destination collision. Also test a filesystem without atomic no-replace rename support.
3. Drag between folders on the same filesystem and across devices. Exercise Copy/Move/Ask and modifier overrides. In Ask, focus Move or Cancel with the keyboard and press Enter. Drop a mixed-parent search selection into one of its source folders.

**Expected result:** Valid restores return to exactly the confirmed location and remove their metadata. Cancelled or rejected restores preserve their payloads and metadata; no destination is overwritten. Unsupported atomic renames produce an explicit safety error. Same-volume plain drops move; cross-device drops follow the chosen strategy. Keyboard actions match the focused button, and valid members of mixed drops still transfer.

## Related issue

Closes #478
Closes #248
Supersedes #338
Includes the follow-ups from #507 and #514.


wmfeht:
Local automated testing on `03e96a2` (`cursor/fix-478-trash-restore-1d3c`) passed.

- `./scripts/check.sh`: `cargo fmt --all --check`, `cargo check --all-targets --all-features`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --all-targets --all-features` (display variables cleared), `cargo deny check`, and `typos` all succeeded. Cargo tests: **1007 passed**, 0 failed, **19 ignored** by design (GTK/display/D-Bus cases that `check.sh` skips).
- `./scripts/e2e.sh`: **210 passed** (harness unit tests and GUI scenarios, including visual baselines) in 4m41s.

Host: Omarchy (Linux x86_64), rustc 1.98.0, GTK 4.22.4. No failures.

wmfeht:
image
image
image

spandan11106:
Read through both commits. The trash validation logic is the right shape overall — physical-item resolution via `standard::target-uri`, `..`/NUL rejection, canonical parent resolution against symlink escapes, filesystem-id equality, `openat2(BENEATH | NO_MAGICLINKS | NO_XDEV)`, `RENAME_NOREPLACE`, and the re-plan-vs-confirmed guard. The unit tests in `trash_restore/tests.rs` cover the attack cases from #478 directly.

Below is what I found in the code. The first one I'd want fixed before this lands.

### 1. Relative `Path=` resolves against the wrong topdir for shared `.Trash/$uid`

`src/adapters/trash_restore.rs`, `resolve_restore_destination`:

```rust
trash_root.parent()
    .ok_or_else(|| RestoreTargetError::new("The original location is invalid"))?
    .join(orig_path)
```

A relative `Path=` is relative to `$topdir`. That holds for `$topdir/.Trash-$uid`, where `parent()` is `$topdir`. It does not hold for the shared layout, which `valid_shared_trash_dir` explicitly supports: there `trash_root` is `$topdir/.Trash/$uid` and `parent()` is `$topdir/.Trash`.

Probed it: source `$top/.Trash/1000/files/report.txt` with `Path=Documents/report.txt` resolves to `$top/.Trash/Documents/report.txt` instead of `$top/Documents/report.txt`. The plan is then *accepted* — it is on the same volume and under `allowed_root`, and the `path_is_within(&destination, trash_root)` guard misses it because `trash_root` is `$top/.Trash/1000`, one level deeper than the directory the file lands in. So Restore files the item into a hidden directory inside the trash tree and then removes its `.trashinfo`, which leaves it unrecoverable from the Trash view.

`find_trash_item_by_orig_path` makes the same `trash_root.parent()` assumption, so for a shared root the entry never matches and discovery falls through to the ambiguous name scan or fails outright.

Worth a test for the shared layout — nothing currently covers it.

### 2. `EINVAL` reported as a cross-volume failure, with no fallback

`src/adapters/local_operations.rs`, `move_restore_path`:

```rust
rustix::io::Errno::XDEV | rustix::io::Errno::INVAL => {
    io_error(format!("Could not restore {display_name} across volumes"))
}
```

`move_local_path` maps that same errno pair to `WouldRecurse` so `move_local_with_progress` falls back to copy-then-delete. `move_restore_path` makes it terminal. `renameat2(RENAME_NOREPLACE)` returns `EINVAL`/`ENOSYS` on filesystems that do not implement renameat2 flags — several FUSE backends, older overlayfs, NFS — which is exactly what a `.Trash-$UID` on removable media tends to be. Restore fails permanently there, with a message blaming volumes rather than the missing flag. The destination is already validated same-volume at that point, so the copy fallback would be safe to reuse.

### 3. `NO_XDEV` rejects destinations the volume check deliberately allowed

`open_local_parent_beneath`. `restore_volume_relation` compares GIO `id::filesystem`, which is identical across a bind mount of the same filesystem. A restore into, say, a bind-mounted `~/Downloads` passes validation and then fails at `openat2` with `Could not safely open restore destination …: Invalid cross-device link`. The policy says yes and the mechanism says no, and the user sees the mechanism's error.

### 4. `VolumeRelation::Unknown` is treated as a confirmed cross-device drop

`src/services/transfer_action.rs`, `drop_commit` routes `Unknown` through `cross_volume_commit`. `lookup_drop_volumes` returns `Pending` (so `Unknown`) whenever a queried directory has a symlink component, is a URI, or sits on a remote/FUSE mount, and it stays `Unknown` if the 2s timeout wins. Under Always Copy, a plain same-disk drop landing in that window silently copies instead of moving. Either default `Unknown` to same-volume behaviour or hold the commit until the lookup lands.

Downstream of that, `src/ui/browser/transfer.rs` states "The destination is on a different device." as fact in the confirmation dialog, which is reached whenever the relation is merely unresolved.

### 5. `drop_is_noop` uses `any`, so one no-op source forbids the whole drop

`src/services/transfer_action.rs`. Dropping a multi-selection that spans directories — recursive search results, for instance — into one of those directories yields an empty drag action for every item, rather than transferring the ones that would actually move.

### 6. Restore silently does nothing when there is no modal host

`src/ui/browser/trash.rs`, `show_restore_confirmation` returns early with no dialog, no restore and no error, after the user has clicked Restore and after the destination lookups have completed. `confirm_cross_volume_drop` handles the same case by falling back to performing the action; this path should fall back or surface an error.

### 7. Destination lookups are serial, each with a 2s timeout

`request_restore` loops `restore_destination_for_location(...).await` per entry, and each call builds a fresh `RestoreContext::current()` that re-reads and re-parses `/proc/self/mountinfo`. Fifty selected items is up to 100s before the confirmation dialog appears, with no progress and no way to cancel. Build the context once and resolve concurrently.

### 8. The end-to-end regression tests for #478 pass vacuously

`src/adapters/local_operations/tests.rs`:

```rust
let Some((home, stick)) = distinct_restore_devices() else {
    return Ok(());
};
```

`distinct_restore_devices()` returns `None` unless `/dev/shm` has a different `st_dev` from `TMPDIR`. `restore_rejects_a_volume_orig_path_on_another_device` is the only full-provider test for the exploit in #478, and it green-passes doing nothing when that does not hold, with nothing in the output saying so. Same shape in `trash_restore/tests.rs::distinct_device_dirs`. The unit-level coverage is genuinely there — the problem is that these four give a false signal about it. Skip loudly, or assert the environment.

### 9. `unescape` loses non-UTF-8 mount points

`src/adapters/volume/mounts.rs` ends in `String::from_utf8_lossy`, so a mount point with non-UTF-8 bytes comes back with replacement characters and `mount_point_for` never matches a path on it. The user-visible result is "Unable to determine the trash volume because the mount table is unavailable" and a refused restore. Everything else in this code is deliberately byte-safe (`OsStr::from_bytes`, `OsString::from_vec`), so this should be too.

### Smaller things

- `trash_scan_mounts` excludes `fuse*` (via `is_remote_fs_type`), so the discovery fallback never scans FUSE-mounted removable media — precisely where volume trash lives. It works in practice through `standard::target-uri`, but the safety net has a hole where it is most needed.
- `FileEntry::thumbnail_path` now carries the physical trash path at three call sites. Same value, and the doc comment is updated, but the name works against the reader. Worth renaming.
- `build_list_pane` installs a drop target on `view` while `stack`, its ancestor, already has one with the same destination. Both get `file-drop-zone`, so both can match `:drop(active)` in `style.css` during one hover.
- Dead branches: the `dest_actions & source_actions` fallback in `offered_file_actions` (the target is always constructed with `COPY`); `relative.as_os_str().is_empty()` in `open_local_parent_beneath` (`Path` equality is component-wise, so the earlier `==` already caught it); the `!allowed_root.exists()` arm in `resolve_restore_destination` (it came from mountinfo).
- `PendingVolumeLookup::start` re-evaluates `resolves_synchronously()`, re-stat'ing each component, and if every directory now resolves synchronously it defers `on_ready` to the full 2s timeout.
- `find_trash_item_by_orig_path` compares paths with raw `PathBuf` equality and no normalization, so `/home/u/./x` against `/home/u/x` misses and falls through to basename matching.
- `escaped_restore_error` appends "Bind mounts and subvolumes are treated as separate volumes." to every escape rejection, including the crafted-`orig-path` case, where it points at the wrong cause.
- `decode_trashinfo_path`: `.trim()` already strips `\r`, so the following `trim_end_matches('\r')` is redundant.

For what it is worth, `cargo clippy --all-targets --all-features -- -D warnings` is clean on the head, and the suite passes apart from `ui::browser::tests::paste::paste_reveals_and_selects_completed_items_in_every_mode`, which fails on the merge base too — pre-existing, not yours.

spandan11106:
Pushed 667291d onto this branch with the mechanical parts of the review above:

- **Shared `.Trash/$uid` topdir** — added `topdir_for_trash_root()` and used it in both `resolve_restore_destination` and `find_trash_item_by_orig_path`, plus regression tests for destination resolution and discovery on that layout.
- `EINVAL`/`ENOSYS` from `renameat2` split out from `EXDEV` and reported as a missing no-clobber rename rather than a cross-volume failure.
- Destination lookups for a selection now share one mount-table snapshot and run in flight together (`restore_destinations_for_locations`), so a slow mount costs one timeout for the batch rather than one per item.
- `show_restore_confirmation` surfaces an error when there is no modal host instead of returning silently.
- `unescape` keeps mount points as bytes, so a non-UTF-8 mount point still matches paths on it.
- Both sides normalized before matching a discovered original path.
- The two duplicated `distinct_device_dirs` helpers consolidated into `test_support.rs`; it now announces the skip and fails under `STRATA_REQUIRE_DEVICE_TESTS`, so those tests cannot pass vacuously.
- Removed three unreachable branches and a redundant `trim_end_matches('\r')`.

Left alone deliberately, since they are behaviour calls rather than defects — happy to go either way on them:

- `NO_XDEV` rejecting bind-mounted destinations that the volume check allows
- `VolumeRelation::Unknown` routing through `cross_volume_commit`, and the dialog copy that states the device difference as fact
- `drop_is_noop` using `any`, so one no-op source forbids the whole drop
- `trash_scan_mounts` excluding `fuse*`
- renaming `FileEntry::thumbnail_path` — worth doing, but it would swamp this diff

`cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings` are clean. One suite failure, `ui::browser::tests::paste::paste_reveals_and_selects_completed_items_in_every_mode` — I checked it at 03e96a2 with my changes stashed and it fails there too, so it is pre-existing and not from this branch. I will open a separate issue for it.

ogarza:
Branch: cursor/fix-478-trash-restore-1d3c
Head: 03e96a2 (fix(trash): validate and confirm restore destinations) on top of 8dfaa21 (feat(transfer): default cross-volume drops to copy)
Base: main
Size: +5724 / −193 across 35 files, two commits
Related: Closes #478 (P0 security), #248; supersedes #338

This review treats the existing Strata architecture as the spec for how to land the fix, and the freedesktop.org Trash spec plus #478 as the spec for what must be true. It does not assume the caller's preferred API names are required.





Executive verdict

Request changes.

The security diagnosis in #478 is correct, and the bones of the trash commit are the right shape: there is already a domain-level RestorePlan produced by plan_restore_from_known_paths / plan_restore_for_location; the Trash UI asks that resolver for a destination and shows the full path; the restore operation re-runs the same planner and refuses to move if the confirmed path no longer matches; the move itself is openat2(RESOLVE_BENEATH) plus renameat2(RENAME_NOREPLACE), not a blind gio::File::move of trash::orig-path. That is substantially the architecture this review would have asked for.

It is not mergeable as it stands. The PR lands a second, conceptually separate product feature (cross-volume drop defaults, settings, clipboard, cursor, Ask dialog) in the same change as a P0 restore-jailbreak fix, and then uses that transfer volume-identity stack as the restore security oracle. Restore is local-only and already has LocalFileIdentity { st_dev, st_ino } plus openat2 helpers in local_operations.rs; it does not need GIO id::filesystem, /proc/self/mountinfo scanning, or a 2s remote-query timeout to decide whether a USB trash item may land in ~/.config/autostart. Three overlapping sameness checks (GIO id, mount-prefix starts_with, RESOLVE_NO_XDEV) disagree on bind mounts and on renameat2-flag EINVAL, which is exactly the class of bug a single st_dev check on opened fds would have made unrepresentable.

There is also a real restore-correctness bug in the shared .Trash/$uid layout (relative Path= joined to .Trash rather than the volume topdir), vacuous skip of the only full-provider exploit test, and ~1,350 lines of Python GVfs-fixture infrastructure whose CI tests never execute Strata. Fix the restore invariant, split the transfer commit, and shrink discovery/fixture complexity. Do not rewrite trash_restore.rs from scratch.





What the PR gets right





The threat model is the right one. Volume trash on a victim-writable mount, crafted Path= / trash::orig-path, restore into ~/.config/autostart (or similar) with no overwrite and no confirmation is a real, high-severity issue. Treating .trashinfo as untrusted metadata is the correct starting point.



RestorePlan is the right domain object. source_path, destination, allowed_root, trash_info is exactly the information the UI and the operation should share. plan_restore_from_known_paths is the small, testable function the security invariant belongs in. Keep it.



Physical item resolution via standard::target-uri is necessary, not ornamental. GVfs names volume-trash entries after the escaped physical path (trash:///%5Crun%5Cmedia%5C…) and writes relative Path= values. Guessing from the trash:/// basename is wrong. Preferring the listing's target URI (already stored today as FileEntry.thumbnail_path) is the correct GVfs accommodation.



Relative Path= is joined to the trash topdir, not interpreted as CWD. That matches the spec for $topdir/.Trash-$uid and is independently tested.



Lexical .. is rejected before any filesystem walk, including percent-encoded %2e%2e. Path::starts_with is component-wise, and there is an explicit similar-prefix test (/media/usb vs /media/usb-evil). Those are the easy ways to get path containment wrong; they did not.



Canonical parent resolution before the last component is the right way to handle missing destination parents without following a dangling dest symlink. The symlink-to-home-from-volume test is the actual #478 attack, and it lives at the domain layer.



Re-plan vs confirmed destination is the right TOCTOU pattern. The UI does not hand a path to restore for the operation to trust. Restore recomputes RestorePlan and aborts if plan.destination != confirmed. The user cannot be shown /media/usb/doc.txt and have the move go to ~/.bashrc because the metadata changed in between.



The move is fail-closed at the kernel. openat2(BENEATH | NO_MAGICLINKS | NO_XDEV) rooted at the allowed area, renameat2(NOREPLACE), no GIO copy onto another filesystem, native-only. Even if planning were wrong, a mount-point swap that turned allowed_root into a symlink to $HOME still dies on EXDEV when renaming off the USB. That is real defense in depth, in the same family as the existing delete openat2 walk.



Undo of a just-trashed item skips the extra prompt but not the validator. Correct UX, same gate.



VolumeRelation::Unknown is treated as not-same on the restore path (!= Same → reject). Fail closed for the security check, even though the transfer commit treats Unknown as cross-volume copy (a different, and worse, policy — see Critical issues).



Domain tests in src/adapters/trash_restore/tests.rs actually encode the exploit. Autostart path from volume trash, .., percent-encoded .., symlink parent off-volume, dest inside the trash directory, similar prefixes, empty mount table. That is the right test layer for the invariant. Do not replace these with GVfs-only tests.





Critical issues



1. Two product changes in one PR





Severity: high (process / reviewability; blocks a P0 security merge)



Where: commits 8dfaa21 (20 files, +2451/−128) and 03e96a2 (21 files, +3284/−78); PR body explicitly wants both so #338 can be closed



What is wrong: Cross-volume drop defaults, Always Copy/Move/Ask, clipboard, compositor MOVE-vs-COPY, list-view drop targets, and settings are a user-visible transfer-policy feature (#248). Trash restore validation is a local filesystem jailbreak fix (#478). They share a tempting helper (VolumeIdentity) and almost nothing else. Reviewing them together hides transfer bugs (Unknown-as-copy, drop_is_noop using any, confirmation copy claiming "different device" while lookup is still pending) behind a security headline, and it makes the security fix wait on DnD UX.



Why it matters: A P0 restore bug should land as soon as the invariant is isolated and tested. The transfer work is large enough to be its own PR (and already was, as #338). Coupling them also pushed restore to use GIO filesystem ids "so it matches drops" instead of st_dev on opened fds, which is the wrong oracle for a local move.



Recommendation: Split. Land fix(trash): … against main with only the restore identity it actually needs. Reopen or retarget the transfer commit as its own PR that may depend on a small shared st_dev / mount helper if one remains, not the other way around. Do not close #248 from this PR.

wmfeht:
Local automated testing on `db0f02e` (`cursor/fix-478-trash-restore-1d3c`) passed.

- `./scripts/check.sh`: `cargo fmt --all --check`, `cargo check --all-targets --all-features`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --all-targets --all-features` (display variables cleared), `cargo deny check`, and `typos` all succeeded. Cargo tests: **1014 passed**, 0 failed, **19 ignored** by design (GTK/display/D-Bus cases that `check.sh` skips).
- `./scripts/e2e.sh`: **210 passed** (harness unit tests and GUI scenarios, including visual baselines) in 4m36s.

Host: Omarchy (Linux x86_64), rustc 1.98.0, GTK 4.22.4. No failures.

wmfeht:
Manual testing for trash restore/reject on `db0f02e` (`cursor/fix-478-trash-restore-1d3c`) succeeded.

Planted `scripts/trash-restore-fixture` on `/run/media/$USER/External-2TB` (writable extra volume; 14 labeled GVfs items). Same-volume restores returned to the destinations shown in the confirm dialog, including nested parents, the directory, and both `RESTORE-collision.txt` copies. Cross-volume rejects were refused; nothing appeared in the home/volume reject sinks, the symlink sink, or `/tmp/478-REJECT-*`. The missing-parent item looked up and the move failed as expected.

Fixture cleaned up afterward; unrelated volume trash left alone.

l0gicgate:
## Validation verdict: request changes

Reviewed **`db0f02eaa0144c8e37be03f688b885fd8dcdb244`**, all PR comments and follow-ups, related issues/PRs, screenshots, and a test merge with current `main` (`9713eb1`).

**This is not ready for approval pending only manual testing.** The core security approach is sound, but I confirmed five remaining defects.

### Merge blockers

| Priority | Finding | Evidence |
|---|---|---|
| **P1** | **The rename fallback can overwrite existing data.** `src/adapters/local_operations.rs:1154–1162` | The existence check followed by an unflagged rename is not atomic. I injected a destination creation between those steps: restore succeeded, replaced the new file, and removed the source. Pinning the parent does not prevent this race. |
| **P1** | **Fixture cleanup deletes unrelated Trash items.** `scripts/trash-restore-fixture/trash_restore_fixture.py:221–224,833–852` | Cleanup treats `478-` anywhere in `.trashinfo` as fixture ownership. An unrelated invoice originally named `invoice-478-final.txt` was permanently deleted in my isolated reproduction. `plant --force` invokes this cleanup too. |
| **P2** | **Non-UTF-8 mount handling remains broken.** `src/adapters/volume/mounts.rs:20–36` | Although `unescape` now preserves bytes, mountinfo still uses `read_to_string`. A real mount containing byte `0xff` caused the **entire mount table** to disappear, disabling ordinary restores too. |
| **P2** | **Enter on focused Move or Cancel performs Copy.** `src/ui/browser/transfer.rs:168–180` | The capture-phase handler always activates Copy. Both cases reproduced through actual keyboard input in the pinned GTK container. |
| **P2** | **One no-op source forbids an otherwise valid multi-item drop.** `src/services/transfer_action.rs:179–188` | Dropping items from directories A and B into A rejects everything, rather than transferring B’s item. The original review identified this correctly; it remains unfixed. |

#### Reproductions and required corrections

1. **No-clobber race:** exercise `rename_without_replacing` through its existing injected rename closure. Return `EINVAL` for `NOREPLACE`; immediately before the unflagged retry, create the destination with different contents; then perform the actual rename. The helper returns success, the original source disappears, and the newly created destination is overwritten. **Fail closed or implement a genuinely no-clobber, descriptor-confined fallback with appropriate failure cleanup.** Add the concurrent-destination regression, not just a test where the destination exists before `statat`.
2. **Fixture cleanup:** in a temporary fake Trash, create `files/invoice.txt` and `info/invoice.txt.trashinfo` with `Path=/home/user/Documents/invoice-478-final.txt`. Call `clean_trash_root` with `fixture_markers("478", "strata-478")`. It reports one removal and deletes the unrelated payload. Require exact records of items created by a unique fixture instance, rather than substring matching, or remove the tool from this PR. Cleanup also scans other discovered volumes even when `--volume` identifies one. **I would not use its current cleanup against real Trash.**
3. **Mount bytes:** inside a private user/mount namespace, mount tmpfs at a pathname containing `0xff`. Linux emits that raw byte in `/proc/self/mountinfo`; the UTF-8 read fails. Both that mount and ordinary `/tmp` then return `None` from `mount_point_for`. Read and parse mountinfo as bytes end-to-end, with a raw non-UTF-8 regression test.
4. **Keyboard dialog:** with Always Ask, drag a file to a temporary second filesystem. From the initially focused Copy button, Shift+Tab focuses Move; another Shift+Tab focuses Cancel. Verify focus through accessibility, then press Enter. Both paths copy. Let the focused button handle activation; do not unconditionally capture Enter for Copy. Add actual GTK/E2E tests for both choices.
5. **Mixed drop:** `drop_is_noop(/fixture/a, [/fixture/a/already-here, /fixture/b/to-transfer])` returns true and forbids the entire drop. Filter no-op sources while retaining per-item self/recursive-drop protections, and transfer the valid remainder. This matters for recursive search selections.

### Were the comments valid, and fixes correct?

| Original comment | Assessment |
|---|---|
| Shared `.Trash/$uid` uses the wrong topdir | **Valid and correctly fixed**, including discovery and rejection of destinations elsewhere within shared `.Trash`. |
| Unsupported rename flags fail with misleading errors | **Valid diagnosis; unsafe final fix.** See the overwrite race above. |
| Bind mounts pass planning but fail execution | **Valid observation, but the proposed simplification is insufficient.** I verified that even plain Linux `rename` returns `EXDEV` across the tested bind boundary. Removing `NO_XDEV` or comparing only `st_dev` does not solve it. Choose a consistent mount policy and report rejection during planning; do not weaken confinement as a quick fix. |
| Unknown volume treated as cross-volume | **Accurate observation, but not inherently wrong.** Copy/Ask is a defensible conservative policy. Assuming “same volume” and moving would be less safe. The misleading dialog wording **was fixed correctly**. |
| `drop_is_noop` uses `any` | **Valid and unfixed.** |
| Missing modal host silently abandons Restore | **The claimed fix is ineffective.** `show_error_dialog` calls `show_error_dialog_after_close`, which requires the same missing modal host and returns too (`src/ui/modal.rs:220–226`). Lower priority because normal rooted windows have a host. Do not fix this by restoring without the required confirmation. |
| Serial destination lookups | **Substantially fixed.** All tasks start before awaiting them, sharing one mount snapshot. The await loop does not serialize their start. Cancellation/progress and bounded concurrency remain follow-up concerns. |
| Cross-device tests can pass without exercising anything | **Partially fixed.** Enforcement exists, but CI does not enable `STRATA_REQUIRE_DEVICE_TESTS`; `volume/tests.rs` also retains a silent-skipping helper. Ordinary test capture can hide the new `eprintln` skips. I enabled enforcement locally. |
| Non-UTF-8 mountpoints | **Valid, incompletely fixed.** The failure occurs before `unescape`. |

Qualifications to the smaller comments:

- The claimed `/home/u/./x` versus `/home/u/x` equality failure is **incorrect**: Rust `Path` equality already handles that component-wise. I verified it.
- FUSE is excluded from the all-mount discovery scan, but physical `target-uri` resolution avoids that path, and absolute orig-path discovery selects a specific mount. This is not evidence that FUSE restores universally fail; avoiding scans of potentially dead mounts is defensible.
- Reusing `thumbnail_path` for the already available physical GVfs target is correct despite the confusing name. A broad rename is not a merge prerequisite.
- Duplicate list/ancestor drop targets deserve focused hover/drop testing, but their presence alone does not prove a double transfer or visible defect. Existing DnD scenarios pass.
- The reclassification edge case where `PendingVolumeLookup::start` reaches zero remaining queries but waits for the timeout remains a low-priority issue. Cache classification or schedule completion immediately in that case.
- The redundant CR trim and relevant dead-branch cleanups were mostly applied; the offered-actions fallback remains. These are not merge blockers. The escape error also still mentions bind mounts/subvolumes for unrelated rejection causes.
- The Python fixture tests do not execute Strata. They are fixture-helper tests, not end-to-end security coverage. Useful manual tooling is reasonable, but its cleanup must be safe.
- The actual default strategy is **Ask**, not Copy as some PR wording implies; #248 permits either, so this is a description correction rather than a product blocker.

The fixes described in **#514 are already incorporated into #502**; merging that follow-up separately would not resolve these findings.

### Is this the right approach?

**Yes for the central restore design:**

- Resolve the physical GVfs trash item rather than guessing from escaped `trash:///` names.
- Treat its metadata as untrusted.
- Produce a shared `RestorePlan`.
- Validate containment and show the full destination.
- Re-plan before execution and reject a changed confirmed destination.
- Execute through confined descriptors with atomic no-clobber behavior.
- Apply the same validator to Undo.

I would retain that design, not rewrite it around a single filesystem-ID comparison. Device equality alone proves neither containment nor rename compatibility. Mount-root policy and kernel-enforced confinement are complementary, not automatically redundant checks.

I **agree with splitting the DnD/settings feature from the security fix**. It adds substantial unrelated behavior and currently introduces two of the blockers above. Keep the planner and physical-source discovery, make planning/execution policy consistent, and preserve fail-closed behavior.

### Independent validation results

The **existing suites pass**, but the targeted regressions expose gaps:

| Check | PR head | With `main` at `9713eb1` |
|---|---:|---:|
| Format / Clippy, all targets/features and warnings denied | Pass | Pass |
| Rust tests, all targets/features, isolated GTK | 1,014 passed | 1,017 passed |
| E2E in the pinned Ubuntu/GTK 4.14 container | 210 passed | 210 passed via canonical `scripts/e2e.sh` |
| Python script tests | 96 passed | 101 passed |

Both Rust runs had **19 explicitly ignored tests**. GTK ran only on private displays, with accessibility bridging disabled for Rust tests and disposable HOME/XDG/D-Bus environments. Cross-device enforcement was enabled. Cargo deny passed with duplicate-dependency warnings; spelling policy was not independently rerun.

The existing suite results above exclude the review-only adversarial probes: the overwrite, mount-byte, mixed-drop, fixture-cleanup, and both keyboard-choice regression assertions **fail** on this head.

The previously reported paste-focus race is tracked separately in **#511** and predates this PR; it did not reproduce in my clean runs. That does not invalidate the documented timing-sensitive failure.

At review time GitHub shows **draft / review required**, with the main CI jobs **skipped**, not independently green.

**Bottom line:** fix the confirmed defects and add their regression tests before treating manual testing as the only remaining gate. Then repeat the actual removable-media/shared-trash, unsupported-rename filesystem, Undo, mixed restore selection, and modifier/cursor/drop-target acceptance cases. The author's manual results are useful but do not exercise the races and keyboard paths reproduced here.

All destructive reproductions used temporary fake Trash contents or private mount namespaces; no real user Trash was modified.

l0gicgate:
Implemented the review fixes and pushed **d45602e** (after merging main), in three focused fix commits.

### Confirmed blockers addressed

1. **Overwrite race:** removed the check-then-rename fallback. Restore requires atomic `RENAME_NOREPLACE`; unsupported operations fail safely, preserving payload and metadata. Added collision, unsupported-operation, cancellation, and confinement regressions.
2. **Unsafe fixture cleanup:** removed the live-Trash fixture utility and its substring-based cleanup entirely. Destructive fixtures now stay in disposable test directories. Replacement guidance: [trash-restore-testing.md](https://github.com/lgse/strata/blob/d45602e89c7cf6fc5819b1eba64c969bf1aee01d/docs/trash-restore-testing.md).
3. **Non-UTF-8 mounts:** read mountinfo as bytes, preserving arbitrary pathname bytes throughout parsing. Added regression coverage; cloned mount snapshots share their entries.
4. **Enter chooses the wrong action:** removed the capture-phase Enter-to-Copy handler. Container E2E tests verify focused **Copy, Move, and Cancel** and the resulting filesystem state in all three view modes.
5. **Mixed no-op drops:** filter individual no-op sources consistently during classification and execution rather than rejecting the selection. Behavioral tests verify that eligible files transfer while no-ops stay untouched.

### Other actionable findings

- Bound restore lookup concurrency to eight; preserve ordering and cancel active lookup futures with the batch. Added visible, cancellable lookup feedback and automatically scrolling destination confirmation.
- Missing modal hosts now surface an error through the requesting overlay instead of silently failing. Cancel, Close, and Escape stop pending lookup; backdrop clicks cannot accidentally dismiss it while work continues.
- Reject nested bind-mount/subvolume boundaries during planning; keep descriptor confinement at execution. Unsupported atomic operations remain an explicit, documented limitation—not a reason to downgrade safety.
- Cache directory lookup classification so filesystem changes cannot strand a pending query until timeout.
- Enforce real distinct-device fixtures through `STRATA_REQUIRE_DEVICE_TESTS=1` in CI and local check/headless runners; eliminate the remaining silent-skipping volume helper.
- Add GTK coverage for full destinations, mixed valid/invalid restores, cancellation, metadata changes after confirmation, and absent modal hosts. Removed the misleading pending-lookup comment and the dead hover-action fallback.

### Validation

On the exact pushed commit: formatting and Clippy pass; isolated Rust suite **1,035 passed, 19 existing ignored**; canonical pinned-container E2E **222 passed**. Also passed: the seven new restore/drop widget tests on GTK 4.14, **88** Python script tests, spelling, and dependency policy.

The PR description now reflects the actual **Always Ask** default, documents manual acceptance, and includes isolated before/after lookup screenshots. No user Trash or unrelated checkout work was modified.

**Implementation and automated validation are complete.** The PR remains draft for manual acceptance on real GVfs/removable/network filesystem setups; I have not represented those manual checks as completed.

l0gicgate:
Reviewed and pushed `aacda2b` after merging verified `lgse/strata` main `6a3e25ed0efbf9d096d92250ea0bfb44fa4c8017` (merge commit `88859c1`; no rebase).

- Integrated the restore/drop changes with main’s refactored settings, preferences, clipboard, and CI infrastructure.
- Restored the Columns drop-completion delay: immediate transfer removed the source label during the animation. Existing real-input E2E regressions exposed both move and failed-move cases.
- Added actual cross-device dispatch coverage for saved preferences before Settings opens, live changes across two browser views, and a view rebuild.
- Audited every added/modified comment, removed redundant narration, and aligned 18 first-party SPDX headers with main’s MIT migration.

Final pinned-container verification: formatting and Clippy passed; Rust **1,325 passed, 20 explicitly ignored**; full canonical E2E **736 passed**. GTK/display, D-Bus, and cross-device fixtures were isolated. No real user Trash was touched.

Manual check: with animations enabled in Columns, drag a disposable file into a folder, then repeat against a non-writable destination. The source label should stay in place during completion; successful moves then remove the source, while failures preserve it. Real GVfs/removable/NFS/FUSE manual acceptance remains outstanding, so I preserved draft status.

Current passing-run screenshot (failed-move result): `/home/l0gicgate/Documents/github/strata/target/e2e-artifacts/gw0/test_drag_completion_keeps_the_source_label_in_place-failed-/screenshot.png`. This is an end-state capture, not proof of animation timing; the real-input E2E checks cover timing. No GitHub attachment-upload/browser tool is available in this session, so this media was **not uploaded**.

Exact-head CI **passed all 52 jobs**: https://github.com/lgse/strata/actions/runs/34418943883. This was explicitly dispatched to exercise the gates while preserving draft status. The PR-triggered checks remain skipped because it is draft; those skips are not claimed as passes. GitHub reports the branch mergeable, but merge is blocked by draft status and required review; real-filesystem manual acceptance is still outstanding. The pull request has not been merged.

ogarza:
Between us: yes, this is worth merging—but I’d make a small volume-lookup correction first, not demand another
 redesign.

 I reviewed final head aacda2b, including the discussion. The earlier serious blockers—overwrite race,
 shared-trash destination, mount-byte parsing, Enter choosing Copy, and mixed-drop rejection—are fixed.

 ### Remaining issues that matter

 1. Automounted network paths can still freeze drag-hover.
 src/adapters/volume.rs:147–185

 Autofs paths remain eligible for synchronous filesystem probing on the GTK thread. An unavailable automounted
 share can therefore block before the two-second timeout applies.

 - Who notices: users with automounted NAS/network directories.
 - When: potentially immediately, not two years later.
 - Recommendation: classify autofs as potentially blocking before probing it. I confirmed the incorrect
   classification; I did not deliberately hang a real server.

 2. A symlink to a network mount can get the wrong Copy/Move policy.
 src/services/transfer_action.rs:17–21

 For example, /home/me/nas → /mnt/nas. The code can recognize identical filesystem IDs but still declare them
 different because one lexical path was marked “remote” and the other wasn’t.

 - Who notices: NAS/FUSE users accessing directories through symlinks.
 - Effect: unnecessary Ask dialogs, or copying instead of the expected same-volume move under Always Copy. No
   demonstrated data loss.
 - Recommendation: separate filesystem identity from “might this lookup block?” I reproduced the mismatch with a
   synthetic mount-table fixture.

 3. Restore’s eight-lookup limit doesn’t bound the underlying work.
 src/adapters/trash_restore.rs:165–275

 A timeout stops awaiting a blocking job; it doesn’t stop that job. The batch then starts another. My probe
 demonstrated nine outstanding blocking jobs despite the limit of eight.

 - Who notices: large restore selections involving stalled filesystems; abandoned jobs can compete with other
   GIO operations.
 - Priority: lower. A focused follow-up is reasonable rather than delaying the security fix indefinitely.
 - Recommendation: retain worker capacity until the actual job finishes, not merely until its awaiting future
   times out.

 ### Architecture and maintenance

 The central restore architecture is good. Keep the shared RestorePlan, physical-source resolution,
 confirmation, execution-time revalidation, descriptor confinement, and atomic no-replace rename.

 I would soften the earlier objection that its safety checks are redundant. Filesystem identity, mount
 boundaries, and kernel confinement answer different questions. Replacing everything with one st_dev comparison
 would not automatically improve correctness.

 The maintenance problem worth addressing is smaller: is_remote currently mixes identity with scheduling policy.
  Separating those responsibilities fixes a real bug and makes future backend support easier.

 Splitting DnD and restore would have made this PR easier to review. At this stage, I wouldn’t require history
 surgery solely for tidiness—unless DnD corrections hold up the security fix.

 ### Comments and practical limitations

 - The callback comment at volume.rs:110 promises an invocation, but synchronous Ready deliberately never
   invokes it. Small documentation fix.
 - “Bounded lookup” needs the worker-versus-future qualification above; the documentation already correctly
   acknowledges that cancellation cannot kill a kernel call.
 - Unsupported atomic renames deliberately prevent some NFS/FUSE restores. That restriction affects real users,
   but is preferable to an unsafe fallback. Actual filesystem acceptance remains worthwhile.

 My recommendation: fix the two volume-classification issues, smoke-test the intended real filesystems, then
 merge. This is useful security work with bounded remaining problems—not an architectural dead end.

 Validation: 135 targeted tests passed; three private probes exposed the gaps above. Exact-head CI has 52
 successful jobs. I didn’t rerun full E2E or real-network acceptance. Private notes and exact commands
 (/tmp/strata-502-review.md) include the unrelated /tmp portal-fixture failure from an initial broader
 selection.

 Nothing was posted to GitHub, and both checkouts are clean.

wmfeht:

Agent: https://cursor.com/agents/bc-743925f2-c0d1-4ff6-9f59-8aee96fb1ade  
Head: `aacda2b5c8cdf8e526bd237fde0ff39d394cc506` (pinned, clean)

### Candidate issues

#### 1. Home-trash restore to another filesystem is rejected (medium — false reject)
- **Expected:** Item in `$XDG_DATA_HOME/Trash` with `Path=` on another mount restores there, or is refused at lookup with a clear “other filesystem” reason.
- **Actual:** Planner/undo fail with “original location crosses a bind mount or subvolume boundary…”. Payload + metadata stay in Trash.
- **Repro:** Home-trash files with `Path=` under `/dev/shm` (distinct `st_dev`). Same-mount home-trash undo still works.
- **Area:** `trash_restore.rs` (`mount_point_for` / `restore_volume_relation`), undo in `local_operations.rs`.

#### 2. Confirmation accepts destinations whose parents do not exist (medium)
- **Expected:** Recreate missing parents under the allowed root, or refuse at lookup/confirmation so Restore isn’t offered.
- **Actual:** Dialog shows full missing-parent path; after confirm, execution fails “Could not safely open restore destination … No such file or directory”. Nothing moved.
- **Repro:** Volume-trash item with `Path=` under missing `gone/nested/`.
- **Area:** `canonical_restore_destination`, Trash UI confirm, `open_local_parent_beneath`.

No high-confidence jailbreak, overwrite, or validation bypass found.

### Checked and consistent
Traversal/`..`/%2e%2e rejects; off-volume symlink parents; similar-prefix non-containment; both `.Trash-$UID` and shared `.Trash/$uid` layouts; re-plan after `.trashinfo` change; `RENAME_NOREPLACE` conflicts; unsupported `renameat2` fail-closed; mixed selection; cancel; same-mount undo. `./scripts/test-headless.py restore` → 80 passed, 2 ignored.

### Limits
Live `trash:///` / `standard::target-uri` and real removable/NFS mounts not available in this environment.

PR #466, opened 2026-09-06
fix(portal): accept bulky file filter lists instead of rejecting them

JoeJoeflyn:
## Description

Clicking "attach files" on github.com opened nothing: GitHub sends its accepted-types list as one file filter with 100+ MIME rules, and the portal backend's request hardening capped a single filter at 64 rules and rejected the whole request (`Backend call failed: a file filter has too many rules` in the journal), so the browser silently failed (#465).

Filter-list size is not an abuse signal worth refusing a dialog over. Large but well-formed lists are now accepted — the rule budget only logs when exceeded. Genuinely malformed input (too many filters, backtracking-heavy globs, overlong strings and labels) is still rejected as before.

## Visual evidence

https://github.com/user-attachments/assets/f4ac4ae8-30eb-44e3-bc8b-69033d8ec9e9

## How to test

1. Make Strata the FileChooser portal backend (`~/.config/xdg-desktop-portal/*.conf` with `org.freedesktop.impl.portal.FileChooser=strata;...`).
2. In Brave, open a GitHub PR or comment box and click attach.

**Expected result:** the Strata file chooser opens and files can be attached. Before the fix the request was rejected with "a file filter has too many rules" and no dialog appeared.

## Related issue

Closes #465

PR #504, opened 2026-09-06
fix(portal): accept file chooser requests with more than 32 filters

### LINKED ISSUE #498: fix(portal): accept file chooser requests with more than 32 filters
### Strata version

0.11.2 (source build, commit `5caa851`). The affected code path is unchanged on current `main` (`46c86c2`).

### Installation method

Built from source

Omarchy (Arch), Hyprland 0.56.2

### What happened?

Uploading a file to NotebookLM in a portal-aware Chromium browser does nothing. No file chooser appears at all — not Strata's, and not a fallback dialog. The click is silently swallowed.

The portal frontend logs the reason on every attempt:

```
Backend call failed: too many file filters
```

That message is Strata's own, from `validate_filters` in `src/portal.rs:385`, gated on `MAX_FILTERS: usize = 32` (`src/portal.rs:38`):

```rust
if filters.len() > MAX_FILTERS
    || current.is_some_and(|current| !filters.contains(current) && filters.len() == MAX_FILTERS)
{
    return invalid_argument("too many file filters");
}
```

NotebookLM's upload input accepts a long list of types (PDF, plain text, Markdown, several audio and document formats). Chromium expands that `accept` list into more than 32 separate portal filters, so Strata rejects the request with `InvalidArgument` before building a window, and the browser gets an error instead of a chooser.

This is a recurrence of #466 one axis over. That fix removed the per-filter and total *rule* caps because "filter-list size is not an abuse signal worth refusing a dialog over", but deliberately kept "too many filters" as a rejection. NotebookLM sends many separate filters rather than one filter with many rules, so it hits the cap that was left in place. The same reasoning applies: the FileChooser spec sets no limit on the number of filters, and a well-formed request from an ordinary web upload form is not abuse.

The inconsistency is visible within the same function — an over-budget rule count warns and accepts (`FILTER_RULE_WARNING_THRESHOLD`), while an over-budget filter count fails the whole request.

Impact: file upload is broken for any site whose `accept` list exceeds 32 types, with no user-visible error and no fallback. From the user's side the button simply does not work.

### Steps to reproduce

1. Enable Strata as the preferred FileChooser backend (`strata --install-portal`).
2. Open NotebookLM in a portal-aware Chromium browser, open a notebook, and click to add a source.
3. Observe that no file chooser opens.
4. Check the frontend log: `journalctl --user -u xdg-desktop-portal.service --since "10 minutes ago"` shows `Backend call failed: too many file filters`.

Any upload form offering more than 32 accepted types reproduces this; NotebookLM is just a convenient case.

### Expected behavior

The chooser opens and shows the requested filters. A filter list that is large but well formed should be accepted — trimmed or coalesced for display if the list is too long to present usefully — and never cause the request to be refused. Following the precedent set by #466, an excessive count should log a warning rather than fail.

### Screenshots or video

N/A — the failure is that no window appears.

### Relevant logs

```shell
$ journalctl --user -u xdg-desktop-portal.service --since "1 hour ago"
xdg-desktop-portal[…]: Backend call failed: too many file filters
xdg-desktop-portal[…]: Backend call failed: too many file filters
xdg-desktop-portal[…]: Backend call failed: too many file filters
```

(Repeats once per upload attempt. Hostname and PID removed.)

### Additional context

Reproducible, not intermittent.

`src/portal/tests.rs:242` builds a `(0..=MAX_FILTERS)` case asserting the current rejecting behavior, so that test needs updating alongside the fix.

Analysis of the failing path is from the frontend log plus the source; I have not stepped through the request itself. `python3 scripts/portal-test.py filters --binary target/debug/strata` extended past 32 filters would confirm it directly.

spandan11106:
## Description

Uploading a file to NotebookLM in a portal-aware Chromium browser opened no chooser at all — not Strata's, and no fallback. The portal frontend logged `Backend call failed: too many file filters` on every attempt.

NotebookLM's upload input accepts more than 32 types, and Chromium expands that `accept` list into one portal filter per type. `validate_filters` capped the list at 32 (`MAX_FILTERS`) and rejected the entire request with `InvalidArgument` before building a window, so the browser got an error instead of a dialog and the user got silence.

This is #466 one axis over. That fix removed the per-filter and total *rule* caps on the grounds that a large but well-formed filter list is not an abuse signal worth refusing a dialog over, but deliberately kept the filter *count* as a rejection. A site that sends many separate filters rather than one filter with many rules lands on the cap that was left standing, with the same failure and the same absence of any user-visible error. The inconsistency was visible inside one function: an over-budget rule count warned and accepted, an over-budget filter count failed the request.

The count now only warns, matching the rule budget's precedent. Genuinely malformed input — backtracking-heavy globs, overlong strings, oversized choice lists — is still rejected.

Accepting long lists exposed a second defect in the chooser itself, reported in review: the filter dropdown was not scrollable. Its options went into a plain `GtkBox` inside a popover, so a long list grew the popover past the screen and its ends could not be reached. The list now sits in a `GtkScrolledWindow` bounded to the room beside the button, and opens toward the roomier side so a control near the window bottom is not squeezed into the last few pixels. Choice dropdowns share the widget and get the same treatment.

## Visual evidence

Before and after, with an 80-filter request, are in `docs/screenshots/498/`.

Before, clicking **Filter** produced no visible list: the popover was sized to all 80 rows and placed off-screen. On a desktop session the same list appeared clipped at the top and bottom screen edges.

After, the list is bounded, opens upward because the control sits near the window bottom, and scrolls through the rest.

The portal rejection itself has no visual evidence to show — the bug was that no window appeared at all.

## How to test

1. Enable Strata as the preferred FileChooser backend (`strata --install-portal`).
2. Build this branch and let it serve the portal.
3. Open NotebookLM in a portal-aware Chromium browser and click to add a source.

Alternatively, without touching desktop portal config, drive the backend directly with more than 32 filters:

```bash
cargo build
python3 scripts/portal-test.py filters --binary target/debug/strata --filter-count 80
```

`--filter-count` is new here; the fixture previously sent three filters, which could not reach either bug. Open the **Filter** dropdown and scroll it.

**Expected result:** the Strata chooser opens with the requested filters listed in its dropdown. Before this change the request failed with `too many file filters` and nothing appeared.

## Related issue

Closes #498


wmfeht:
I believe the list is not scrollable:

image

spandan11106:
Confirmed, and my description was wrong to wave it off — the list was not scrollable at all.

`ChooserDropdown` put every option straight into a `GtkBox` inside the popover, with nothing bounding it. A popover is its own surface and sizes to its content, so 80 rows grew it past the screen and neither end could be reached. Accepting long filter lists is what made that reachable in ordinary use, so it belongs in this PR rather than a follow-up.

Fixed in 10eeec2:

- the options now sit in a `GtkScrolledWindow` with `propagate-natural-height`, so short lists are unchanged and long ones scroll
- the height is bounded to the room available beside the button, and the popover opens toward the roomier side — the filter row sits near the window bottom, so it now opens upward instead of being squeezed into the last few pixels
- it reuses the existing `.context-menu-scroll` styling, so the scrollbar follows `@theme_*` like the context menus

Choice dropdowns share the widget and get the same fix.

Before/after with an 80-filter request are in `docs/screenshots/498/`. `scripts/portal-test.py` grew a `--filter-count` flag to make the case reproducible — the fixture sent three filters, which could not reach either bug:

```bash
python3 scripts/portal-test.py filters --binary target/debug/strata --filter-count 80
```

Two tests cover it: the placement arithmetic including the cramped-window floor, and a GTK test asserting the list is bounded below its natural height so it actually scrolls.

l0gicgate:
Reviewed the complete diff against current `main` and merged the base without conflicts. No functional or high-confidence security blockers found. Audited every added comment and shortened the dropdown comment to the non-obvious GTK surface constraint; no UI behavior changed during review.

Verification: an isolated Xvfb/private-bus request with 80 filters opened the chooser and returned a clean cancellation. The long-dropdown GTK regression test and complete local pre-push gates, including canonical rootless-Podman E2E, passed. NotebookLM/browser routing and interactive scrolling were not manually verified.

PR #526, opened 2026-09-07
fix(browser): share one range anchor between clicks and the keyboard

### LINKED ISSUE #521: fix(browser): Shift-click multi-select fails in List after navigating from an empty folder
## Strata version

0.11.2 / source commit `a0173607e8c9acf841410a9c3aa247ee800162f2` (tracking fork of `lgse/strata` `main`)

## Installation method

Built from source

Debian-based cloud VM, XFCE, X11 (`GDK_BACKEND=x11`). Isolated temp `HOME` with synthetic pinned places (`EmptyDir` empty, `FilesDir` with `file-01.txt`…`file-06.txt`). List view, single-click previews on, type-to-search off. Default List click activation (files 2-click, folders 2-click).

## What happened?

In List view, Shift-click range multi-select does not create a range after navigating from an empty folder (sidebar) into a folder with files. The second click selects only that file (previous selection is replaced). Keyboard `Shift+Down` can still extend a multi-selection in the same listing.

Impact: the usual click-anchor then Shift-click-end range gesture is unreliable right after that navigation, so bulk select/copy/trash from List is easy to get wrong.

## Steps to reproduce

1. Open Strata in List view with the sidebar visible.
2. Pin or otherwise reach an empty directory and a directory that contains several files.
3. Click the empty directory in the sidebar. Confirm the pane shows “This directory is empty.”
4. Click the populated directory in the sidebar. Confirm the files appear. The first row may show a keyboard outline / auto-focus from navigate.
5. Click the first file, then Shift-click the last file.

## Expected behavior

Shift-click selects the contiguous range between the anchor file and the clicked file (Finder/Nautilus-style).

## Actual behavior

Only the Shift-clicked file remains selected. The range is not created.

## Screenshots or video

[Reproduced on XFCE/X11 with synthetic fixtures. Recordings captured during this investigation:

https://cursor.com/artifacts/v/art-8000cce4-350f-4888-8218-4ea3eb19baf5

## Relevant logs

No crash. No relevant log lines; the listing loads normally.

## Additional context

Not a duplicate of open/closed issues searched for List/Explorer Shift-click, range select, multi-select after navigation, empty folder, or sidebar. Nearby but different: #241 (keyboard selection not scrolled into view), #374 (Grid sidebar Enter does not focus files), #462 (first arrow after dismissing a filter only selects), #483 (paste follows auto-selected folder after navigation).

### Narrowing (same VM, isolated HOME)

- **Empty folder specifically?** The reported EmptyDir → FilesDir path reproduces. The failure is not limited to “empty” as a filesystem fact: navigating between two populated pinned places (OtherFiles → FilesDir) also left Shift-click unable to form a range. Any navigation that rebuilds the List pane with no pointer click yet is enough. Clicking blank space under the rows did not give a cleaner signal than sidebar navigation.
- **Sidebar vs other navigation:** Sidebar is sufficient. Opening `FilesDir` by double-clicking it from Home in the listing also failed to range-select afterward. Address-bar typing was not used.
- **List / Icons / Columns:** List is the reported mode and was reproduced. Icons uses the same pointer Shift-click helper as List and also failed to range-select in a live pass. A Columns pass was attempted; the UI still appeared to be List in the screenshot, so Columns is **not confirmed** here. Columns has a separate row-click path.
- **Anchor / preview:** After EmptyDir → FilesDir, the first file is often keyboard-outlined (`select_first_on_load`) but there is no filled multi-selection and no preview until an unmodified click. The original report said selecting one file first (preview appears) then Shift-clicking another works. That contrast was **not reliable** on this VM: click `file-02.txt` (preview “preview content 02”) then Shift-click `file-05.txt` still often selected only `file-05.txt`. When keyboard range *did* work, it did not depend on preview.
- **Keyboard vs mouse:** Window-level `Shift+Up`/`Shift+Down` calls `Browser::extend_selection`, which uses the column `selection_anchor` (set by `select_first_on_load`). List/Icons mouse Shift-click does **not** use that anchor. It uses a per-view `Cell<Option<u32>>` created when the List/Icons pane is built.
- **Click activation:** Left at List defaults (files 2-click, folders 2-click) with Settings “Single-click file previews” on. Did not need to change single- vs double-click activation to reproduce.

### Code observation (no patch)

List and Icons install `install_modified_selection_click` on each row. Unmodified click writes the mouse `Cell` and returns (GTK performs the select). Shift-click reads `anchor.get().unwrap_or(position)`, calls `select_range`, **never writes the Cell**, and **claims** the sequence so GTK ListView native Shift-range does not run. `rebuild_list()` / `rebuild_icons()` on `ColumnAdded` allocate a new `Cell` at `None`. `select_first_on_load` fills the *browser* `selection_anchor` and GTK selection, but that value is not copied into the mouse Cell. After navigation, the first Shift-click therefore ranges from the clicked row to itself (one item).

No fix is included with this report.

spandan11106:
## Description

Every mode kept the pointer range anchor in its own `Cell<Option<u32>>` — List and Icons in
`install_modified_selection_click`, Columns in `column_rows`. Navigating rebuilds the pane
and allocates a fresh cell at `None`, while the anchor `select_first_on_load` fills lives on
the navigation column — so the first Shift-click in a freshly loaded listing ranged from the
clicked row to itself and collapsed the selection to one file. `Shift+Down` kept working
because the keyboard path already reads the column anchor.

The pointer path now reads and writes that same anchor: a plain or Ctrl-click moves it, a
Shift-click ranges from it, and because it is stored as a location it survives a re-sort.
That also makes a click re-anchor the keyboard extension, which previously kept using
whatever the load had selected.

## Visual evidence

Before: navigating into `documents` and Shift-clicking `spreadsheet.csv` leaves only that
row highlighted.

After: the same gesture highlights `notes.txt`, `report.md` and `spreadsheet.csv`.

## How to test

1. Open a directory containing several files from the sidebar or by opening it in the listing. Repeat in all three view modes.
2. Without clicking a file first, Shift-click the last file.
3. Then click the first file and Shift-click the last one.

**Expected result:** step 2 selects every row from the one the listing auto-selected down to
the clicked file; step 3 selects the range between the two clicked files.

## Related issue

Closes #521


spandan11106:
Amended and force-pushed after a closer look at the other view modes.

The first version only rewired List and Icons. Columns keeps its own anchor cell in
`column_rows`, and since it is the default mode the reported bug was still live there:
opening `documents` auto-selects `notes.txt`, and Shift-clicking `spreadsheet.csv`
selected only `spreadsheet.csv`. Columns now reads and writes the same column anchor
through `ViewMap`, which already exposed both position directions, and the new
regression tests run against all three modes instead of just the single-pane ones.

One related behavior I left alone: after a rubber-band selection, a following Shift-click
ranges from the load or last-click anchor rather than collapsing to the clicked row,
because the marquee writes the GTK selection directly and never moves the anchor. Neither
the old nor the new behavior anchors at the marquee origin, so fixing it means teaching
the marquee to set the anchor — a separate change.

spandan11106:
image

l0gicgate:
Fixed the remaining Columns cases: initial/sidebar navigation now initializes the first-entry anchor, and clicking back into an inactive column selects its first visible entry without closing descendants. Clicking empty space in the already-active column still clears selection. Added regression coverage and removed redundant comments.

Verification: reproduced both failures with isolated real-pointer scenarios, then verified initial/sidebar navigation in all three modes and parent-column re-entry followed by Shift-click. The expected contiguous range is selected.

Reviewed Columns baseline (first-entry selection):

Before:
Before

After:
After

wmfeht:
## Merge Blocker

Landed head `a640858` still has these miller-column regressions (shared range-anchor itself is fine):

1. Columns calls `select_first_on_load` on `ColumnAdded`, which selects the first visible row instead of only seeding a range anchor. Navigate/sidebar and preview-open therefore commit a child selection, so F2/Copy/Delete/Right can target that first entry instead of the open-folder / empty-selection contracts. Existing Columns rename tests stay green because they never take the ModeViews path.

2. Returning to an inactive column via content background click `browser.select()`s the first visible entry after clearing. That clears `load_cursor` and can make paste land inside the first directory rather than the column the user clicked. Header clicks still preserve selection; content return does not.

For Columns, write the location `selection_anchor` without treating it as a user pick. Do not keep miller auto-select unless the paste/F2/Delete contracts are intentionally changed and covered on the real UI path.

l0gicgate:
Follow-up: the filename-specific failure was a focus bug. Tracing confirmed that Shift-click selected the full range, but claiming the content click suppressed GTK’s focus transfer. The parent therefore kept inactive selection styling, nearly invisible in the reported theme.

Modified content clicks now explicitly focus the clicked row in all three modes. The regression reproduces file → populated folder → Shift-click the original filename and checks both keyboard focus and selected entries; clicking beside the filename remains covered. Temporary tracing has been removed.

Manual reproduction: click a file, click a folder to open its child pane, then Shift-click the original file name. Expected: the parent range has active highlighting and the clicked row owns keyboard focus, without closing the child.

[Capture details](https://github.com/spandan11106/strata/blob/a640858e17b9b62b1975dda78322c16910e6d7c0/docs/screenshots/521/README.md)

Before:
Before

After:
After

PR #529, opened 2026-09-07
feat(selection): clear the active selection with Escape

### LINKED ISSUE #528: feat(selection): clear the active selection with Escape
### Problem or use case

There is no simple keyboard-only way to dismiss the current file selection. After selecting one or more entries, pressing Escape leaves them selected, even though Escape conventionally clears selection in file managers and similar list interfaces.

### Proposed solution

When the active file pane has a selection, pressing Escape should clear it.

- Clear both single-item and multi-item selections.
- Affect only the active pane in a multi-pane workspace.
- Work consistently in Columns, Icons, List, and other browsing views.
- Preserve focus in the pane so keyboard navigation can continue.
- Do not navigate, activate an item, or alter the current directory.
- Respect existing Escape precedence: dismiss an open menu, dialog, location editor, filter, or other transient surface first. If a selection remains afterward, a subsequent Escape should clear it.
- If quick preview is open, retain the established preview-dismiss behavior before clearing selection.

### Alternatives considered

- Require clicking empty space. This is less efficient for keyboard workflows and may be impossible when a pane is full.
- Require a dedicated shortcut such as Ctrl+Shift+A. That can be added separately, but Escape is the conventional and discoverable action for cancelling the current selection state.

### Visual examples

N/A — keyboard interaction only.

### Strata version

Current development build (`v0.11.2`).

#### Acceptance criteria

- [ ] Escape clears a single selected entry when no higher-priority transient UI is active.
- [ ] Escape clears a multi-selection.
- [ ] Only the active pane's selection is affected.
- [ ] Pane focus and the current location are preserved.
- [ ] Existing Escape behavior for filters, editors, menus, dialogs, and quick preview continues to take precedence.
- [ ] Tests cover every browsing view and Escape precedence.

l0gicgate:
## Description

Clear the active pane’s selection with Escape in Columns, Icons, and List, retaining its keyboard cursor and current directory. Menus, dialogs, inline editors, filters, and previews still dismiss first; a subsequent Escape clears the selection even when dismissal returns focus to pane controls.

## Visual evidence

N/A — keyboard-only interaction; no layout or styling changes.

## How to test

1. In each view, select one file, then multiple files, and press Escape. Press an arrow key: selection and focus should move to the adjacent item on the first press.
2. In Columns, open a child folder and select entries there. Press Escape and verify the parent selection and both columns remain unchanged.
3. With a file selected, open a context menu, Properties, rename/new-folder editor, location editor, filter, or quick preview. Press Escape twice.
4. Focus a folder with the keyboard, press Escape, then Enter. The focused folder should open without jumping to the first entry.

**Expected result:** Escape clears only the active selection without moving focus or navigating. Transient UI dismisses on the first Escape, leaving selection intact; the next Escape clears it.

## Related issue

Closes #528

### DIFF
--- a/src/app/browser.rs
+++ b/src/app/browser.rs
@@ -891,8 +891,22 @@ impl Browser {
         closed
     }
 
+    pub fn clear_active_selection(&self) -> bool {
+        let cleared = self.state.borrow_mut().clear_active_selection();
+        if let Some((depth, focused)) = cleared {
+            self.emit(BrowserEvent::SelectionSetChanged {
+                depth,
+                positions: Vec::new(),
+                focused,
+                take_focus: false,
+            });
+            return true;
+        }
+        false
+    }
+
     pub fn escape(self: &Rc<Self>) {
-        if self.close_peek() {
+        if self.close_peek() || self.clear_active_selection() {
             return;
         }
 
--- a/src/app/browser/tests.rs
+++ b/src/app/browser/tests.rs
@@ -2513,7 +2513,7 @@ fn keyboard_selection_and_activation_descend_without_the_ui() {
 }
 
 #[test]
-fn escape_closes_a_peek_before_the_deepest_column() {
+fn escape_closes_a_peek_before_clearing_selection_and_closing_the_deepest_column() {
     let browser = Browser::new(Rc::new(FakeFileSource));
     let events = Rc::new(RefCell::new(Vec::new()));
     let observed = events.clone();
@@ -2539,6 +2539,16 @@ fn escape_closes_a_peek_before_the_deepest_column() {
     );
 
     events.borrow_mut().clear();
+    let location = browser.active_location();
+    browser.escape();
+    assert!(browser.selected_positions(1).is_empty());
+    assert_eq!(browser.active_location(), location);
+    assert!(events.borrow().iter().any(|event| matches!(
+        event,
+        BrowserEvent::SelectionSetChanged { depth: 1, positions, .. } if positions.is_empty()
+    )));
+
+    events.borrow_mut().clear();
     browser.escape();
     assert!(
         events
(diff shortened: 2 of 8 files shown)


wmfeht:
## Merge Blocker

1. **Icons/List: Escape does not resume keyboard navigation from the preserved cursor.** After Escape clears the GTK selection, the next arrow key re-selects the same item instead of moving. `clear_active_selection` empties `selected_locations` / GTK `MultiSelection` while leaving `column.selected` as the cursor; the next key hits `sync_mode_selection()`, which reads the empty GTK selection and writes it back with `set_selection(depth, &[], None)`, wiping the cursor. Icons/List then hand off to native GTK, which re-selects the focused widget without moving. Columns is fine. The PR’s e2e only waits for a non-empty selection, so it passes even when `readme.md` is still selected after Right/Down.

2. **Icons/List: Enter after Escape does not open the focused item.** `activate_focused_in_place` treats a missing cursor as `move_selection(1)`, which selects the first listing entry and does not activate. After focusing `documents` → Escape → Return, Icons/List jump to `archive` and stay in the root listing.

Fix direction: keep the keyboard cursor when GTK selection is empty (do not let `sync_mode_selection` / `set_selection` clear `column.selected`), and keep GTK item focus independent of selection so native arrows move from that item. Strengthen the e2e assertions to require the *next* name after Escape, and that Enter still opens the focused folder.

l0gicgate:
Addressed both findings in 589a6fe.

- Empty GTK selection sync now retains the keyboard cursor and uses actual item focus rather than the last selected index.
- Native arrows resume selection at that cursor without scheduling a focus restore that would undo the move.
- Regression tests assert the exact adjacent selection **and focus** after the first arrow, in both directions with single/multi-selection, and verify Escape → Enter opens the focused folder in every view.

The strengthened tests reproduced both regressions before the fix (10 failures in Icons/List). All local gates now pass, including 323 canonical container E2E tests.

Issue #533, opened 2026-09-07
Decide global search's local-mount inclusion policy (follow-up to #532)


spandan11106:
### Context

Follow-up to #532. That issue removes Strata's raw `/proc/self/mountinfo` fallback from the Devices sidebar so sidebar visibility fully defers to the GIO/GVfs volume-monitor backend's own policy.

`global_search_roots()` (`src/ui/window/devices.rs`) currently reuses the exact same raw mount enumeration for a different purpose: deciding which local mount points Ctrl+K global search scans, in addition to Home. That behavior is intentionally left unchanged by #532 — this issue is to decide, separately, whether search should keep including mounts the volume-monitor backend does not surface as a Volume/Mount (e.g. a hidden subvolume like the Pacman cache), or whether search scope should also defer to the backend the way sidebar display now will.

### Problem to decide

Sidebar visibility and global search scope are two different questions that happened to share one implementation (`system_mounted_devices()`). Now that sidebar display defers entirely to the backend's visibility policy (#532), search's own policy needs its own explicit decision rather than silently inheriting whatever #532 leaves behind:

- Should search continue to scan backend-hidden mounts (e.g. a cache subvolume) since a user might legitimately want to search "everything under this machine's local mounts," even things hidden from the Devices list?
- Or should search scope also be limited to what the backend exposes, for consistency with what the sidebar shows and to avoid surprising results from directories the user never sees as "locations" in the app?
- Is there a case for a middle ground (e.g. still search hidden-but-local mounts, but exclude well-known non-user system paths regardless of backend visibility)?

### Additional context

Raised during design/review of #532: "keep search behavior unchanged in this fix... the global search searches in volumes too... it's two PRs... one to show the correct devices, and another to track the search policy."

PR #536, opened 2026-09-07
fix(devices): defer sidebar mount visibility to the volume-monitor backend

### LINKED ISSUE #532: Devices sidebar shows mounts GVfs already hides (e.g. Pacman cache subvolume)
### Strata version

main branch, built from source (nightly local build tracking `origin/main`)

### Installation method

Built from source

Arch Linux, Hyprland (Omarchy)

### What happened?

The Devices section of the sidebar lists mounts that GVfs/udisks2 already decided to hide from the user — for example a Pacman package cache directory that lives on its own mount point or btrfs subvolume (a common layout, e.g. a separate `@cache` subvolume).

Root cause: `append_devices()` (`src/ui/window.rs`) merges GIO's own volumes/mounts with a raw fallback built by `devices::system_mounted_devices()` (`src/ui/window/devices.rs`), which parses `/proc/self/mountinfo` directly and only excludes paths matched by GLib's `is_mount_path_system_internal()` (a short hardcoded list: `/`, `/boot`, `/proc`, `/sys`, `/run`, gvfs mounts, etc.). `unrepresented_devices()` only adds a fallback entry when GIO's own `mounts()`/`volumes()` did **not** already report it — which means the fallback exists specifically to reinstate mounts GIO chose not to expose. In practice this is exactly what the udisks2-backed GVfs volume monitor is supposed to decide on its own (its "what-is-shown" policy only surfaces mounts under `$HOME`, `/media`, `/run/media/$USER` unless overridden per-mount with `x-gvfs-hide`/`x-gvfs-show`, or the KDE/udisks2 equivalents). Strata's raw scan has none of that policy, so it shows things the backend deliberately hid.

Separately, `gvfs`/`udisks2` are not declared as runtime dependencies in packaging (only `gvfs-smb` is optional-listed), so on minimal setups GIO's `VolumeMonitor` can silently degrade to GLib's bare "unix" monitor, which lacks this policy and misses real removable drives — this is very likely why the raw-mountinfo fallback was added in the first place (#441).

### Steps to reproduce

1. Have a system directory such as `/var/cache/pacman/pkg` mounted on its own mount point or subvolume backed by a real block device (not tmpfs) — e.g. a separate `@cache` btrfs subvolume.
2. Confirm GVfs's own policy would hide it (not under `$HOME`, `/media`, `/run/media/$USER`, no `x-gvfs-show` override).
3. Open Strata and look at the Devices section of the sidebar.
4. Observe an entry named after the mount's last path segment (e.g. `pkg`) even though it is not a real user-facing device.

### Expected behavior

Devices should reflect exactly what the desktop's volume-monitor backend reports: GIO volumes, plus non-shadowed GIO mounts not already represented by a volume. Strata should not reimplement or second-guess the backend's own mount-visibility policy (no raw `/proc/self/mountinfo` scanning, no reading `/etc/fstab`, no reimplementing `x-gvfs-hide`/`x-udisks2.hide`).

### Additional context

**Proposed fix**

- Drop the raw-mountinfo fallback (`system_mounted_devices()` / `unrepresented_devices()`) from the sidebar path entirely. `append_devices()` becomes: GIO volumes + non-shadowed GIO mounts not already tied to a volume.
- Do not attempt to detect "backend is degraded" as a reason to keep a conditional fallback: an empty (or short) device list is a legitimate answer, not evidence of a broken backend, and `gio::VolumeMonitor::get()` can return a union monitor whose type does not reveal which child implementation (e.g. udisks2) is actually active — there is no established, reliable way to distinguish "no devices" from "backend unavailable" from Strata's side. Absent a real detection mechanism, removing the fallback outright is the correct call rather than guessing.
- Require, per distro, whichever package provides the udisks2 volume-monitor GIO extension (e.g. `gvfs` on Arch; `gvfs-udisks2`/`gvfs-daemons` on Fedora/Debian) as a runtime dependency, rather than a single hardcoded package name across all packaging.
- Global search (`global_search_roots()`) is explicitly **out of scope** for this issue — it currently reuses the same raw mount enumeration for a different purpose (search scope) and should keep doing so unchanged here. Deciding whether/how to change search's own mount-inclusion policy is tracked separately.

**Acceptance criteria**

- The Pacman-cache subvolume (or any similarly backend-hidden mount) stays hidden, whether hidden by the backend's default policy or by an explicit `x-gvfs-hide`.
- A legitimately empty device list renders correctly — no error state, no "backend unavailable" messaging inferred from it.
- An unmounted `gio::Volume` still appears in Devices and mounts correctly on click.
- Hotplug and unmount events still update the Devices list live.
- Network shares (SMB) still appear, exactly once, with no duplicate row against their backing volume.
- Behavior with the udisks2/GVfs backend missing or unavailable is verified, even though it becomes a required dependency.
- Global search scope (`global_search_roots()` and its test coverage) is provably unchanged by this fix.

spandan11106:
## Description

The Devices sidebar merged GIO's own volumes/mounts with a raw `/proc/self/mountinfo` scan (`devices::system_mounted_devices()` + `unrepresented_devices()`), which reinstated any `/dev`-backed mount the volume-monitor backend chose not to expose. That raw scan has no notion of the backend's own visibility policy (the home/media/run-media allowlist, `x-gvfs-hide`/`x-udisks2.hide`, etc.), so it surfaced mounts the backend deliberately hides — most visibly a Pacman cache directory living on its own mount point or subvolume, which then showed up in Devices named after its last path segment.

This drops that fallback from the sidebar entirely. `append_devices()` now shows exactly what `gio::VolumeMonitor` reports: its volumes, plus non-shadowed mounts not already tied to a volume. No fstab parsing or reimplementation of the backend's hide/show conventions was added — visibility policy stays owned by GIO/GVfs/udisks2, the same way Nautilus, Nemo, Thunar, and Dolphin already do it.

Global search's mount enumeration (`global_search_roots()`) is intentionally unchanged — it still uses the raw mount-table scan for its own root-inclusion policy, which is tracked separately in #533.

## Visual evidence

N/A for this PR from me — reproducing the reported layout requires a mount the backend actually hides (e.g. a real Pacman-cache subvolume), which isn't something this sandboxed environment can set up (no privileged mount access). This is the exact configuration the report came from, so a before/after screenshot from that machine is more convincing than a synthetic one; happy to attach it during review if wanted.

## How to test

1. On a system where a directory like `/var/cache/pacman/pkg` lives on its own mount point or subvolume (not tmpfs) that the volume-monitor backend doesn't surface as a Volume/Mount, open Strata and check the Devices section — confirm the entry is gone.
2. Insert a USB drive and confirm it still appears under Devices, in both its "not yet mounted" state (mounts on click) and once mounted.
3. Unplug/unmount it and confirm the sidebar updates live.
4. Mount an SMB share and confirm it appears exactly once, with no duplicate row against its backing volume.
5. On a machine with no external drives and nothing else to show, confirm the Devices heading and separator don't render at all (no error state).

**Expected result:** Devices shows only what the volume-monitor backend actually reports as a volume or mount; nothing the backend hides on purpose reappears. Global search behavior is unaffected.

## Related issue

Closes #532

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,12 +157,16 @@ On Arch Linux or Omarchy:
 
 ```bash
 sudo pacman -S --needed bubblewrap ffmpeg ffmpegthumbnailer fontconfig \
-  gst-libav gst-plugins-good gtk4 gtksourceview5 poppler-glib
+  gst-libav gst-plugins-good gtk4 gtksourceview5 gvfs poppler-glib
 # Optional SMB and broader camera RAW support:
 sudo pacman -S --needed gvfs-smb imagemagick libraw dcraw
 ```
 
 GTK **4.12 or newer** and glibc **2.39 or newer** are required. Other glibc-based distributions may work when they provide equivalent runtime libraries, but their package names and binary compatibility vary. Systems with an older glibc must [build Strata from source](#development-and-documentation).
+
+Device discovery requires the GVfs UDisks2 volume monitor (`gvfs` on Arch and
+Fedora; `gvfs-daemons` on Debian/Ubuntu). Without that backend, removable drives
+may be absent from Devices. SMB support remains optional.
 
 #### 2. Download and verify
 
@@ -385,7 +389,7 @@ Build requirements are the latest stable Rust toolchain, a C toolchain, `pkg-con
 
 ```bash
 sudo pacman -S --needed base-devel rust bubblewrap ffmpeg ffmpegthumbnailer fontconfig \
-  gst-libav gst-plugins-good gtk4 gtksourceview5 poppler-glib
+  gst-libav gst-plugins-good gtk4 gtksourceview5 gvfs poppler-glib
 make start-dev        # rebuild and restart as files change
 make run-dev          # build and launch the main app once
 make run-chooser-dev  # build and open an isolated Save chooser with choices
--- a/install.sh
+++ b/install.sh
@@ -7,7 +7,7 @@ APP_ID="io.github.lgse.Strata"
 MIN_GLIBC="2.39"
 REQUIRED_PACKAGES=(
   bubblewrap desktop-file-utils ffmpeg ffmpegthumbnailer fontconfig gst-libav
-  gst-plugins-good gtk4 gtksourceview5 poppler-glib github-cli xdg-utils
+  gst-plugins-good gtk4 gtksourceview5 gvfs poppler-glib github-cli xdg-utils
 )
 RAW_PREVIEW_PACKAGES=(imagemagick libraw dcraw)
 
@@ -422,7 +422,8 @@ main() {
     fi
   else
     printf '\nStrata needs GTK 4.12+, GtkSourceView 5, Poppler GLib, Fontconfig, Bubblewrap,\n'
-    printf 'FFmpeg, ffmpegthumbnailer, and GStreamer runtime plugins.\n'
+    printf 'FFmpeg, ffmpegthumbnailer, GStreamer plugins, and the GVfs UDisks2 volume monitor.\n'
+    printf 'For the volume monitor, install gvfs-daemons on Debian/Ubuntu or gvfs on Fedora.\n'
     if [[ $NON_INTERACTIVE == yes ]]; then
       die "Non-interactive dependency installation currently supports Arch-based systems only."
     fi
--- a/packaging/aur/PKGBUILD.in
+++ b/packaging/aur/PKGBUILD.in
@@ -15,6 +15,7 @@ depends=(
   'glibc>=2.39'
   'gtk4>=4.12'
   'gtksourceview5'
+  'gvfs'
   'poppler-glib'
 )
 optdepends=(
--- a/packaging/aur/strata-bin/.SRCINFO
+++ b/packaging/aur/strata-bin/.SRCINFO
@@ -11,6 +11,7 @@ pkgbase = strata-bin
 	depends = glibc>=2.39
 	depends = gtk4>=4.12
 	depends = gtksourceview5
+	depends = gvfs
 	depends = poppler-glib
 	optdepends = ffmpeg: video previews
 	optdepends = ffmpegthumbnailer: video thumbnails
--- a/packaging/aur/strata-bin/PKGBUILD
+++ b/packaging/aur/strata-bin/PKGBUILD
@@ -15,6 +15,7 @@ depends=(
   'glibc>=2.39'
   'gtk4>=4.12'
   'gtksourceview5'
+  'gvfs'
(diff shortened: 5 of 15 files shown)


l0gicgate:
Review is blocked locally; **no commits have been pushed**.

Merged `main` cleanly in an isolated worktree and prepared two fixes:
- Require the GVfs UDisks2 monitor in Arch packages/installer, with distro-specific guidance, completing the missing dependency requirement in #532.
- Give each RAW-thumbnail fallback its own temporary staging directory. The existing CI failure comes from concurrent tests racing on `/tmp/raw-thumb`; added concurrent regression coverage.

The complete diff/comment audit found no additional sidebar blockers and no added/modified comments. Global-search enumeration remains unchanged in behavior.

Formatting, compilation, Clippy, 103 script tests, and the canonical Podman E2E suite (323 tests) pass. The private-Xvfb native Rust suite on GTK 4.22.4 still fails `ui::browser::tests::paste::paste_reveals_and_selects_completed_items_in_every_mode` at the scroll wait in `src/ui/browser/tests/paste.rs:112` (975 passed, 1 failed, 19 ignored). Investigation did not produce a reliable fix; speculative changes were reverted. This prevents pushing under the repository's pre-push policy. Existing GitHub CI therefore remains unchanged/red.

Inspected the captured empty Devices state on the isolated display: no Devices heading or extra device separator. Real hidden mounts and USB hotplug were not exercised. Media upload tooling is unavailable; the screenshot was **not uploaded**: `/home/l0gicgate/strata-review-536/target/review/devices-empty.png`.

Retained worktree: `/home/l0gicgate/strata-review-536`, branch `fix/532-review-536-1788792741`; local fixes `33a994e`, `b39d862`. Logs are under `target/review/` there. PR remains unmerged.

l0gicgate:
The earlier blocker is resolved. Pushed through `37eefcb`; all GitHub checks passed, including packaging and E2E.

- Fixed the CI RAW-preview race with per-invocation temporary staging.
- Fixed the native paste-scroll failure: an old pre-layout scroll to row 0 could overwrite the newer scroll to pasted files. New requests now cancel pending scrolls; the regression test fails when that cancellation is removed.
- Included the reviewed GVfs dependency changes and base merge.

Local verification: formatting, compilation, Clippy, dependency policy, 103 script tests, 977 Rust tests on private Xvfb, and 323 canonical Podman E2E tests passed. Captured the E2E cut/paste result with the destination file selected. Upload tooling is unavailable; screenshots were not posted and are preserved at `/tmp/pr536-review-artifacts/paste-columns.png` and `/tmp/pr536-review-artifacts/devices-empty.png`.

The PR was merged by @l0gicgate while I was finishing verification; I did not invoke the merge.

PR #538, opened 2026-09-07
fix(chooser): size and center portal dialogs on Hyprland

### LINKED ISSUE #535: fix(chooser): respect available bounds in split-window layouts
### Strata version
Not confirmed. Screenshot shows an installation-complete message, but the exact version/commit is not reliably legible.

### Installation method
Other — not confirmed; installation output is visible in the screenshot.

Linux with split/tiled windows; distribution, compositor and versions not confirmed.

### What happened?
The floating portal file chooser is too large when invoked from a split window. Its initial size appears identical to opening it from a fullscreen application rather than adapting to the available area.

### Steps to reproduce
1. Enable Strata as the system portal file chooser.
2. Tile/split a browser alongside another window.
3. Trigger a file upload/open dialog.
4. Compare the chooser dimensions with those obtained from a fullscreen browser.

### Expected behavior
The floating chooser should respect available compositor bounds and remain usable in smaller layouts, rather than always requesting monitor-sized dimensions.

### Screenshots or video
User supplied a Discord screenshot from reporter `fadilasif`, with the browser, oversized floating chooser, and a terminal in a split layout.

Reporter at 8:10 AM: “file chooser in split windows doesnt fit it”

Reporter at 8:14 AM: “it is a floating windows but the size isnt being adjusted according to the size of the window trying to open it” / “its the same size as opening it on full screen”.

Maintainer `l0gicgate` acknowledged the report and promised a ticket at 8:13 AM.

Reporter screenshot: oversized chooser in a split-window layout

The above is reporter evidence, not an independently reproduced result. [Controlled layout comparisons and implementation scope](https://github.com/lgse/strata/blob/4ad1ef9b279be3391135b6e2049787a7837794de/docs/screenshots/535/README.md).

### Relevant logs
No reliably readable diagnostic logs supplied; no crash reported.

### Additional context
Current chooser implementation computes initial dimensions from monitor geometry and reapplies them on monitor entry (`src/ui/chooser.rs`). Wayland exported parent handles establish transiency but do not expose the foreign parent's dimensions. GTK already respects compositor-provided toplevel bounds. Exact parent-relative sizing cannot be promised when the compositor does not provide parent geometry. The proposed improvement uses a bounded read-only Hyprland query only for an unambiguous app-ID match; other cases retain the monitor-based fallback.

l0gicgate:
## Description

Improve Strata's portal chooser sizing and placement on Hyprland:

- Use a size hint when exactly one application window matches the requesting app ID.
- Center floating choosers on their monitor by default, not over the calling window.
- Preserve the parent/modal relationship and manual resizes across monitor entry.

Strata installs a bounded, chooser-only named runtime rule before showing the dialog. It supports Lua and legacy named-rule configurations, refreshes after config reloads, and never edits user config files or targets normal file-manager windows. Other compositors or unavailable IPC retain compositor placement; missing/ambiguous app identities retain monitor-based sizing.

## Visual evidence

Actual before/after captures from an isolated Hyprland 0.56.2 session, with the same 720×460 chooser and off-center parent:

| Before: centered over the calling window | After: centered on the monitor |
| --- | --- |
| Parent-centered chooser | Monitor-centered chooser |

[Original reporter screenshot](https://raw.githubusercontent.com/lgse/strata/dc12eeaf37792a4a9bf7921e4a38c4ad31bd15b7/docs/screenshots/535/reporter.png) · [Sizing comparisons, geometry results, and capture details](https://github.com/lgse/strata/blob/dc12eeaf37792a4a9bf7921e4a38c4ad31bd15b7/docs/screenshots/535/README.md).

## How to test

1. On native Wayland under Hyprland with this build's portal enabled, invoke Open or Save from an off-center or split application window.
2. Resize the chooser and move it between monitors. Close it, reload Hyprland's configuration, and invoke another chooser.
3. Check that regular Strata windows retain their usual placement. For compact sizing, use one application window whose current or initial class matches its portal app ID; additional matching windows should use the sizing fallback.

**Expected result:** Floating choosers open centered on their monitor, including after config reload. Unambiguous split applications get smaller initial dialogs. Parent/modal behavior, manual resizes, and normal file-manager placement remain intact.

## Related issue

Closes #535. The reporter confirmed the sizing improvement; monitor centering implements the requested follow-up default.


l0gicgate:
Merged `main` (`0fb4e06`) into this PR as `ee62fc12403bbf98ea32e2268fc8391ce6091380` after passing the required local checks.

Started the [RC release workflow](https://github.com/lgse/strata/actions/runs/34151236809) with `mode=rc`, `bump=patch`, and the merge commit pinned as its source. Source verification has passed and both architecture builds are in progress.

Issue #557, opened 2026-09-07
refactor(archives): consolidate backends and share extraction state


ogarza:
### Problem or use case

Archive operations currently live in `src/adapters/local_operations/archive.rs`, using separate `zip`, `tar`/`flate2`, and `sevenz-rust2` backends. Although destination handling and name resolution are shared, each extraction path repeats member processing, progress, cancellation, cleanup, and result tracking. This increases maintenance effort and makes behavior across formats harder to keep consistent.

### Proposed solution

Deliver this in three separately reviewed PRs:

1. **Shared extraction session — #562:** centralize destination safety, conflict naming, progress, cancellation, partial-file cleanup, and result reporting. Organize archive modules by responsibility without changing decoder libraries.
2. **Decoder evaluation:** compare libarchive, maintained Rust alternatives, and retaining the current libraries. Evaluate format/encryption compatibility, streaming, cancellation, maintenance, licensing, and packaging before choosing an approach.
3. **Sandboxed decoding:** isolate decoders in a helper with read-only archive access, no network or destination access, and bounded IPC to the parent-owned extraction session. Build on Strata’s bubblewrap approach, with archive-specific resource/output limits, cancellation/timeout cleanup, and fail-closed behavior.

Preserve traversal/symlink protections, browser integration, and supported capabilities throughout. Review compression separately where backend capabilities differ. The goal is a smaller, consistent archive layer—not a UI redesign; implementation details belong in the respective PRs.

### Alternatives considered

Use libarchive for broad format support, adopt a higher-level Rust archive library, or retain the existing libraries while consolidating extraction bookkeeping. A separate extraction backend and compression writers may be appropriate if needed for compatibility. Sandboxing does not depend on replacing the current decoders.

### Visual examples

N/A — internal refactor.

### Additional context

Related: #434 (invalid-archive errors), #479 (extraction limits). A backend change alone does not resolve these policy concerns. Keep this issue open through the follow-up work.

wmfeht:
Combining this with #479

l0gicgate:
[...]

PR #562, opened 2026-09-07
refactor(archives): share extraction state across decoders

ogarza:
## Description

Introduce a per-operation `ExtractionSession` shared by the existing ZIP, TAR/gzip, and 7z decoders. It centralizes member processing, conflict resolution, progress, cancellation, cleanup, and outcomes without changing decoder libraries.

Separate archive coordination, compression, destination safety, and decoding into focused modules with adjacent tests. Compression and confined filesystem-write helpers are mechanical moves; name mapping is shared with pending-result reporting. Add regression coverage for the shared session and format compatibility.

This is PR 1 of 3: shared extraction session, decoder evaluation, then sandboxed decoding. The latter two remain separately reviewed follow-ups.

### Review follow-up

Fix two inherited cancellation-reporting bugs: completed files now report their actual conflict-renamed destinations, and 7z tracks individual header entries rather than filename suffixes or assumed callback order. Decoders supply pending names; the session validates and maps them without filesystem probes, name reservations, or scanning unread TAR contents.

Legacy TAR-name conversion and link-flattening limitations are documented, not changed. Buffer optimization remains deferred pending measurement; no decoder trait hierarchy was added.

### Why the diff grows

Including the review fixes, the net increase of **911 lines** comes from **922 additional test/fixture lines**, **24 documentation lines**, and **35 fewer production-file lines**. Excluding blank lines and line comments, production grows by 36 lines: module imports add 44 lines, while the remaining non-comment lines shrink by 8.

The gain is centralized extraction policy and focused tests, not library consolidation yet. Existing format libraries remain deliberately in place. A future backend such as libarchive may reduce that glue, but the shared session remains necessary for Strata-specific destination safety, conflicts, cancellation, and cleanup.

## Visual evidence

N/A — internal refactor with no intended visual changes.

## How to test

1. In a temporary folder, create a source directory containing nested files, an empty file, and an empty directory. Use Strata to compress it as ZIP, 7z, TAR, and TAR.GZ; also try password-protected ZIP and 7z.
2. Extract each archive twice into its own destination, supplying the password when prompted.
3. With a sufficiently large archive, cancel extraction while it is running and inspect the destination after the operation stops.

**Expected result:** Contents and empty entries are preserved, repeat extraction uses a consistently renamed top-level directory without overwriting earlier contents, and cancellation retains completed files while removing the interrupted partial file. Existing password prompts and progress behavior remain intact.

## Related issue

Refs #557. Leave open for the decoder evaluation and sandboxing PRs.


ogarza:
Owner-reported verification: the manual archive smoke test passed for bdc8a09, before the latest upstream-main synchronization.

l0gicgate:
## Isolated code review — `f0dac39`

**Overall: good architectural direction.** Centralizing extraction policy while keeping decoder replacement separate makes this easier to verify. I found no newly introduced merge-blocking regression.

### Reproduced findings

Both issues below are **inherited from the previous implementation**, not regressions introduced by this PR. Two temporary review tests reproduced them; they are suitable focused fixes or follow-ups.

1. **7z cancellation can omit pending entries** — [`decoders.rs:53–69`](https://github.com/lgse/strata/blob/f0dac390be66b6f1b3ef00a7ebef069af8a7ddfa/src/adapters/local_operations/archive/decoders.rs#L53-L69)

   `sevenz_locations_from` assumes callback order matches header order. `sevenz-rust2` processes streamed entries before a separate pass over streamless entries. With a directory followed by a populated file in the archive, cancelling before extraction reports only the file as pending, omitting the directory.

   **Recommendation:** track processed members using stable identities rather than a filename-based suffix of the header list. Filename matching also becomes ambiguous for duplicate names. Add cancellation coverage mixing directories, empty files, and populated files.

2. **Completed outcomes can report the wrong destination** — [`extraction.rs:104–121`](https://github.com/lgse/strata/blob/f0dac390be66b6f1b3ef00a7ebef069af8a7ddfa/src/adapters/local_operations/archive/extraction.rs#L104-L121)

   `create_file` returns the actual `created` path, but successful completion records `outpath`. Processing two members named `same.txt` creates `same.txt` and `same (2).txt`; cancelling afterward reports `same.txt` twice in `completed`.

   **Recommendation:** record `created` for files and add duplicate-member outcome coverage.

### Abstraction and library usage

- The main split is sound: decoders enumerate/lend streams; the session owns extraction policy; destination helpers own confined filesystem operations. Replacing the 7z `RefCell` bookkeeping with a mutable session fits the library's `FnMut` API well. No additional decoder trait hierarchy seems necessary yet.
- **Tighten cancellation ownership:** decoders still construct destination `Location`s. Prefer supplying pending member identities/names and letting the session consistently handle validation, rename mapping, and destination reporting. The current helpers duplicate that policy differently across formats.
- Before freezing the backend contract, reconsider `&str` plus `Directory/File`: TAR names undergo lossy conversion, and non-directory entry types, including links, are flattened into files. These are existing limitations, but the abstraction should make them explicit rather than conceal them.
- Streaming through `Read`, including solid 7z streams, is appropriate. Keeping writes under Strata's control instead of using convenience `extract`/`unpack` APIs preserves its destination policy. `rustix` descriptor-relative operations and `tempfile` staging remain justified. Library replacement should stay in the separately planned compatibility/packaging evaluation.

### Optimization

The clearest candidate is **reusing the 1 MiB copy buffer per operation**: [`archive.rs:408`](https://github.com/lgse/strata/blob/f0dac390be66b6f1b3ef00a7ebef069af8a7ddfa/src/adapters/local_operations/archive.rs#L408) allocates and initializes it for every file, including empty files. Benchmark many-small-file archives before claiming a performance gain. Optimizing away `&mut dyn Read` is unlikely to be worthwhile beside decoding and filesystem I/O.

### Review validation

Formatting, Clippy, and all **42 archive tests passed** under private Xvfb. The two additional probes reproduced the findings above and were removed afterward. Full-suite/E2E validation was not run for this review. Review was performed in a detached worktree; no production changes were made.

ogarza:
Addressed the review in two focused follow-up commits:

- 8ca48c5: completed file outcomes record the actual created path. Regression coverage includes duplicate members at the root and inside a conflict-renamed directory.
- 656b59e: 7z cancellation tracks header-entry identities independently of duplicate names and callback order. Deterministic tests cover cancellation before extraction and during copying, with mixed streamed/streamless entries in solid and non-solid archives. Pending names now pass through session-owned validation and known rename mapping; reporting neither probes/reserves destinations nor scans unread TAR contents.

Both reported bugs were reproduced with failing regression tests before fixing them. The filename/entry-type boundary is explicitly documented as provisional: native-name and link semantics remain for the decoder compatibility evaluation. Buffer reuse remains a separate measured optimization, and there is no new decoder trait hierarchy.

The PR description and diff-size explanation have been updated. The earlier owner-reported smoke test predates these follow-up commits.

wmfeht (approved):

wmfeht:
## QA / verification pass (post-merge)

**Ref used:** detached `656b59e15960d8567acd09418d79c90e654d3a51` (PR head via `starting_ref`). Did **not** use merge commit `231c8f8` / current `origin/main` (local `origin/main` on this snapshot is still `2001656` / v0.13.0).

**Environment:** Cursor Cloud Agent VM, Ubuntu 24.04.4, GTK 4.14.5, rustc 1.98.1, 4 CPUs / 15 GiB. GTK and E2E ran on private Xvfb (not the Cloud Agent `DISPLAY=:1` desktop). Docker e2e used the host-network wrapper required on this VM (`bridge: none`).

No follow-up PR. Prefer reporting; none of the issues below are a clear one-line production fix I would land as a draft without more isolation.

---

### 1. `./scripts/check.sh` — **fail** (exit 101)

`cargo fmt`, `cargo check`, and `cargo clippy --all-targets --all-features -- -D warnings` succeeded. `cargo test` (script unsets `DISPLAY`/`WAYLAND_DISPLAY` as designed):

- **1017 passed**, **1 failed**, **19 ignored** in 247.76s
- Failure: `sandbox::tests::running_thumbnail_process_trees_are_stopped_on_timeout_and_cancellation`
  - panic: `thumbnail provider descendant did not start` (`src/sandbox/tests.rs`)
  - Unrelated to archives. Isolated retries **passed twice** (`exit 0`, ~2.4s each).
  - Full-suite log is noisy with leftover `writer: cannot create /tmp/strata-preview-…/timeout-marker` after the panic.
- `cargo-deny` / `typos` **not reached** (`set -e` after the test failure). They are also not installed on this image.

### 2. E2E

**Canonical gate** `PATH=/tmp/docker-hostnet:$PATH ./scripts/e2e.sh`:

- **1 failed, 335 passed** in 446.31s (0:07:26)
- Failure: `tests/e2e/scenarios/test_marquee_scrolling.py::test_scrolling_extends_marquee_without_losing_earlier_files[edge-icons]` (`TreeTimeout` waiting for marquee to reach `060.txt`). Unrelated to archives; overlapped a native archive UX run on the same 4 CPUs.
- **Retry (serial, container):** same test **passed** (`1 passed, 5 deselected` in 13.83s).

**Archive dialog scenarios** (container, serial):

```text
./scripts/e2e.sh -n 0 tests/e2e/scenarios/test_dialogs_and_menus.py -k "compress or extract"
# 3 passed, 14 deselected in 9.97s  (exit 0)
```

Covers ZIP compress via Enter, invalid `../escape` name, and Extract to….

**Native archive unit/UI tests** under private Xvfb:

```text
xvfb-run -a env -u WAYLAND_DISPLAY GDK_BACKEND=x11 GTK_A11Y=none NO_AT_BRIDGE=1 STRATA_REQUIRE_GTK_TESTS=1 \
  cargo test --all-features --bin strata adapters::local_operations::archive
# 48 passed

cargo test … ui::browser::archive     # 2 passed
cargo test … ui::browser::context_menu  # 13 passed (includes Extract here / Extract to… menu cases)
```

Skipped: `./scripts/e2e-native.sh` as a substitute for the full container gate (ran native only for extra archive UX).

### 3. Manual compress / decompress (Strata UI on private Xvfb + AT-SPI harness)

Drove the real `target/debug/strata` binary the same way e2e does.

| Scenario | Result |
| --- | --- |
| Nested folder (`payload/` + empty dir + empty file + nested file) → ZIP | Pass — `bundle-zip.zip` created (699 B in one run) |
| Extract to… into a new folder | Pass — top-level `payload/` restored; UI **navigates into the destination** (documented `pending_navigate`) |
| Extract here with existing `payload/` | Pass — creates `payload (2)`, original untouched |
| Multi-file ZIP (`readme.md` + `todo.txt`) → Extract to… | Pass — `pair-out/` contains both files |
| Empty folder ZIP | Pass — `empty-dir.zip`; extract preserves empty `archive/` |
| Compress collision | Pass — “File already exists” / Replace for `replace-me.zip` |
| Password-protected ZIP | Partial — compress dialog + Extract password prompt appear; one run did not find `payload/item.txt` after submitting the password |
| ZIP / 7Z / TAR / TAR.GZ UI roundtrip | Incomplete — harness clicks on format toggles did not leave a focused name field; recording stayed on ZIP |
| Cancel mid-extract (~16×4 MiB stored JPEG ZIP, progress ~12/17) | **Did not interrupt** — destination still had all 16 files |
| Password 7z / TAR UI | Not completed in UI (see format-toggle note). Decoder unit tests cover all four formats + passwords |

Worker-level cancellation (`cancelling_extraction_waits_for_the_worker_and_reports_incomplete_output` and the 7z identity tests) **passed**. The gap is UI cancel vs a fast extract, not the session policy tests added in this PR.

### Ranked findings

1. **P2 — Extract Cancel may not stop remaining members in the UI.** Repro: compress a bulky folder (~64 MiB stored ZIP, 16 files), Extract to…, wait for dialog `Working`, click `Cancel` around 12/17. After idle, all 16 files were present; no leftover `*.part`. Same mechanism (`LoadHandle` drop) works in the unit test when cancellation is requested at `ArchiveStarted`. Worth a dedicated UI e2e that asserts an incomplete destination (slow 7z or many members).
2. **P3 — Extract password dialog focus / AT-SPI.** Prompt title `Extract` / `secret-zip.zip` appears after encrypted ZIP extract-without-password (good). The field is `password text` and was **not** `focused` in the dump, so the shared e2e `editable_field()` helper (role `text` + focused) times out. Code calls `grab_focus()`; may be a11y-only.
3. **P3 / harness — Compress format segmented control.** Clicking `7Z` / `TAR.GZ` / `TAR` in the harness moved focus off the name field; the recording appeared to keep `ZIP` selected. Existing e2e only submits ZIP. Not confirmed as a pointer/GTK bug vs AT-SPI. Decoder tests already round-trip ZIP, 7z, TAR, TAR.GZ (plain + password where supported).
4. **Flakes, not this PR:** `check.sh` sandbox thumbnail spawn; container marquee `[edge-icons]` (passed on retry).

### Expected UX (not bugs)

- Extract to… navigates into the destination when the operation completes, so a second extract of the same archive requires returning to the parent.
- Repeat extract into a folder that already has the top-level member uses `name (2)` rather than overwrite — matches the PR test plan.

Refs #557.

PR #567, opened 2026-09-07
fix(ui): create items immediately and finalize renames safely

### LINKED ISSUE #566: New folder click-away freezes the whole app
**Strata version**
0.12.1-rc.1 (release binary). Also reproduced on current `main` when built from source (commit `af7bbef`).

**Installation method**
GitHub release archive (binary under `~/.local/bin`).

Arch Linux (Omarchy), Hyprland 0.55. Reproduced on a plain Xvfb display too, so the compositor is not a factor.

**What happened?**
Creating a new folder (or file) by shortcut (`Ctrl+Shift+N`), typing a name, and then **clicking away onto another file/folder or a sidebar entry without pressing Enter** freezes the entire app. The window stops accepting any input and one CPU core spins at ~100% until the process is killed.

It only happens in **List** and **Icons** views. In Columns view the same action works fine.

**Reproduction steps**
1. Open any directory.
2. Press `Ctrl+Shift+N` to create a new folder.
3. Type a name but do **not** press Enter.
4. Click another file, another folder, or a sidebar location.

**Expected behavior**
Final owner-approved behavior, consistent in Columns, List, and Icons:

- New Folder immediately creates `new folder`; New File immediately creates an empty `new file`. If occupied by any item, use the first free `new folder (1)` / `new file (1)`, `(2)`, etc. Select the allocated default name for editing.
- For new and existing files/folders, Enter or clicking away anywhere—including empty space—commits a valid name. Keyboard focus loss follows the same rule.
- Escape or empty/invalid input keeps the original name. Cancelling the initial rename leaves the newly created item under its allocated default; it does not delete it.
- Preserve file contents and existing items; no overwrite on a naming conflict. Existing files retain extension-aware name selection. Keep every view responsive without GTK parent-assertion spam.

This replaces the interim cancellation-only policy discussed earlier. Creation now happens before editing, so click-away finalizes a rename rather than submitting a temporary creation prompt.

**Actual behavior**
The app freezes on one core at ~100% CPU and stops reacting to all input. `top` shows the `strata` main thread in `R` state; a `gdb` backtrace shows it stuck in GTK's `synthesize_focus_change_events` walking a freed widget's parent chain forever (a debug build logged `Gtk-CRITICAL **: gtk_widget_get_parent: assertion 'GTK_IS_WIDGET (widget)' failed` millions of times).

Pre-fix root cause: when the field lost focus, `finish_mode_new_entry` destroyed the ListView row that owned the real keyboard focus *inside* the `focus-out` handling while GTK was mid-flight in `gtk_window_root_set_focus`. GTK then resumes walking the dangling widget to the window root (NULL != root → infinite loop). Columns view only hides its row, which is why it is unaffected.

**Relevant logs**
No application-level logs are produced; the app hard-freezes before writing anything. The only output is the GTK-CRITICAL above, produced in a tight loop (~30 million occurrences per minute in a debug build). The fix is tracked in #567, now updated through `d1edb6c` (folder behavior in `1f7a1c8`, matching file behavior and shared implementation in `d1edb6c`). Temporary creation rows have been removed. The owner manually verified folders in all three views, then reviewed the matching file functionality and approved this update. This issue remains open until the PR is merged.

**Additional context**
This looks specific to Strata's two-`Grid`-of-`ListViews` mode views (List/Icons) where the newly created entry is rendered as an in-place row. In-process unit tests cannot reproduce it because it requires real GTK input focus; it was reproduced and verified with a live XTEST input session under Xvfb.

Deoxizn:
## Description

Replace temporary creation rows with atomic creation followed by the normal rename editor, eliminating the row-removal path behind #566. File and folder creation share collision handling; renames are dispatched after GTK's focus walk, with pending-editor and offscreen-focus guards.

The final owner-approved behavior is consistent in Columns, List, and Icons:
- New Folder / New File immediately creates `new folder` / an empty `new file`, choosing the first free numbered alternative on conflict and selecting its default name.
- Enter or click-away commits a valid rename. Escape or empty/invalid input keeps the original name—including the allocated default for a new item.
- Existing items are not overwritten and file contents are preserved. Existing files retain extension-aware name selection.

This supersedes the earlier cancellation-only iteration of this PR.

## Visual evidence

Before: [original List freeze screenshot from #566](https://raw.githubusercontent.com/wmfeht/strata/1e40ced4bce9f4e8de3b02787a26356bc97edcc1/docs/repro/566/list_view_click_away_cpu100.png).

After a real pointer click on empty space, `renamed.item` is committed and the editor closes:

| View | New file | New folder |
| --- | --- | --- |
| Columns | [Screenshot](https://raw.githubusercontent.com/Deoxizn/strata/d1edb6c3a2a8e36069d638598412c4a92ffa388b/docs/screenshots/566/columns-new-file-rename.png) | [Screenshot](https://raw.githubusercontent.com/Deoxizn/strata/d1edb6c3a2a8e36069d638598412c4a92ffa388b/docs/screenshots/566/columns-new-folder-rename.png) |
| List | [Screenshot](https://raw.githubusercontent.com/Deoxizn/strata/d1edb6c3a2a8e36069d638598412c4a92ffa388b/docs/screenshots/566/list-new-file-rename.png) | [Screenshot](https://raw.githubusercontent.com/Deoxizn/strata/d1edb6c3a2a8e36069d638598412c4a92ffa388b/docs/screenshots/566/list-new-folder-rename.png) |
| Icons | [Screenshot](https://raw.githubusercontent.com/Deoxizn/strata/d1edb6c3a2a8e36069d638598412c4a92ffa388b/docs/screenshots/566/icons-new-file-rename.png) | [Screenshot](https://raw.githubusercontent.com/Deoxizn/strata/d1edb6c3a2a8e36069d638598412c4a92ffa388b/docs/screenshots/566/icons-new-folder-rename.png) |

## How to test

1. In a disposable directory, repeat in Columns, List, and Icons: use Ctrl+Shift+N and background menu → New File. Verify the selected default names; repeat with occupied defaults to check `(1)`, `(2)`, etc.
2. Type a valid name without Enter, then click a file, folder, sidebar location, or empty space. Repeat using Enter and with existing items renamed through F2.
3. Try Escape after typing, an empty name, spaces only, `.`/`..`, `bad/name`, and a conflicting name. Check that the original item and its contents remain intact; cancelling a new item's rename must not delete it.
4. Confirm inside-field clicks preserve editing and creation brings the editor into view in large or filtered directories.

**Expected result:** Files and folders are created immediately, valid renames stick on Enter/click-away, and cancelled or invalid edits retain the original name. All views stay responsive without GTK parent-assertion spam.

The owner manually verified folders in all three views and reviewed the matching file functionality before approving this update.

## Related issue

Closes #566


wmfeht:
## Merge blocker

**Do not merge.** The P0 freeze from #566 is still present on `41e03cfd9dfa1eedf86a8bf840a0e1e99437061f`.

The diagnosis is right: List/Icons destroy the focused ListView/GridView row while GTK is inside `gtk_window_root_set_focus` / `synthesize_focus_change_events`. Deferring `placeholder.splice` with `glib::idle_add_local_once` does **not** stop a real pointer click-away from wedging that walk.

Columns remains fine because it hides a persistent row (`new_entry_row.set_visible(false)`) instead of destroying a focused ListItem.

## Live XTEST repro (this head)

Private Xvfb + XTEST (not the Cloud Agent desktop). `Ctrl+Shift+N`, type a name, click `todo.txt`.

- **Icons and List:** the inline prompt stays on screen, no folder is created, AT-SPI drops (“the Strata application is not on the accessibility bus”).
- Logs: **~7.2M** `gtk_widget_get_parent` CRITICALs in List (~781 MiB) and **~8.0M** in Icons (~861 MiB) within ~8s — the same spin as #566.
- After-click frames still show `clicked-away-icons` / `clicked-away-list` in the prompt. The main loop is stuck before a successful submit/redraw.

In-process tests only `emit_by_name("leave")` then `grab_focus()`. That never enters `gtk_window_root_set_focus`, so they cannot catch this. Canonical E2E uses Enter/Escape, not click-away. The PR text says the freeze was validated with live XTEST; that does not match this head.

`idle_add_local_once` is the wrong tool here: a real click runs `leave` *inside* GTK’s focus bookkeeping. Either the idle still runs too soon, or ListView/GridView unparents the focused row on pointer focus-change independently of the splice.

A fix needs:

1. A real XTEST or e2e **click-away** gate in List **and** Icons.
2. A live widget in the focus chain until GTK finishes the walk — same idea as Columns, or move focus off the row **before** teardown.

## Other findings

Not merge-blocking by themselves:

- **Re-open skip is fragile.** Skip is `if field.upgrade().is_some() { splice }`. List/Icons factory widgets are recycled (`icons-new-entry-wired`, `ensure_rename_field`). `begin_new_entry` always `cancel_new_entry()` first, so a WeakRef to the old Entry can still upgrade and wipe the new prompt. A generation token is the reliable guard.
- **`if let (Some(placeholder), Some(stack))`** now requires both. Today every `ActiveModeNewEntry` sets both; a placeholder-only path would leak the row.
- **Empty-dir `show_status` is captured at finish time** and applied later. `create_directory` is async, so this is usually the old empty flash; if a listing lands first, idle can flip the stack back to `"status"`.
- **Tests:** no Icons case; list field wait is an unbounded `loop` + `iteration(false)`; leftover `eprintln!("STEP: …")`.
- **Escape/Enter** now share the deferred splice. Canonical e2e covers those and passed; they are not the freeze path.
- **PR is behind `origin/main`** (includes `chore: release v0.13.0`). Touched files do not conflict with current main.
- GitHub: Metadata policy passed; `[code]smith` skipped; fork CI `action_required`.

Window key handling returns `Proceed` while `new_entry_is_active()`, so `Ctrl+Shift+N` does not call `create_new_folder` again while the prompt is focused. Escape-then-immediate New Folder (button / `begin_new_entry`) can still race the idle splice.

## Test-suite results

Head: `41e03cfd9dfa1eedf86a8bf840a0e1e99437061f`

| Check | Result |
| --- | --- |
| `cargo fmt --all --check` | **pass** |
| `cargo clippy --all-targets --all-features -- -D warnings` | **pass** |
| `xvfb-run` + `STRATA_REQUIRE_GTK_TESTS=1` `cargo test --all-targets --all-features` | **994 passed**, 0 failed, **19 ignored**. New synthetic leave tests passed. |
| `./scripts/e2e.sh` (host-network Docker) | **329 passed** in 396.89s |
| Native XTEST click-away, **List** | **FAIL** — freeze, `gtk_widget_get_parent` loop, no folder |
| Native XTEST click-away, **Icons** | **FAIL** — same |
| GitHub Actions (fork) | Metadata policy success; test workflows **not run** |

Ignored Rust tests match this VM (GVfs Trash, `dbus-run-session`, `xdotool` mapped-window cases). Harmless `libEGL DRI3` noise on Xvfb.

## Residual risk

Even after a correct teardown, **no automated gate exercises pointer click-away**. E2E only uses Enter and Escape (`test_create_folder_from_the_keyboard`, `test_creating_a_folder_can_be_cancelled`). A green `./scripts/e2e.sh` does not mean #566 is fixed.

ogarza:
Review follow-up, now pushed as 0768638:

- I could not confirm the claim that idle deferral itself fails at the original PR head (`41e03cf`). Nine real XTEST folder click-away cases (file/folder/sidebar targets across all three views) passed on both pinned GTK 4.14.5 and host GTK 4.22.4. Temporarily restoring synchronous removal reproduced the parent-assertion loop in List and Icons while Columns passed. The negative-control run confirms these scenarios exercise the original freeze.
- Confirmed the cleanup concerns with failing unit tests: a replacement prompt could be removed, a stale empty-list snapshot could hide content, and cleanup could be skipped when the Entry died. Cleanup now checks the placeholder item's identity and current listing, independently of Entry lifetime or the presence of a status stack.
- Replaced misleading synthetic-click test names/debug logging and the unbounded field wait. Real-pointer coverage now includes empty-space dismissal, with additional validation, Enter/Escape/Tab, collision, and empty-directory reopen coverage.

**Owner-approved behavior change:** click-away now **cancels**, rather than submits. Only Enter with a valid nonblank name creates a file/folder; empty or whitespace-only Enter cancels. This is consistent across Columns, List, Icons, and the shared chooser controls. The PR description and #566 now reflect that decision.

@ogarza manually verified the updated functionality in all three views and approved this push. Fresh pre-push verification passed formatting, Clippy, 998 Rust tests (19 ignored) under the fully isolated display/session-bus runner, and all 406 pinned-container E2E tests. The 98 native entry-management E2E cases also passed in the final native run.

ogarza (dismissed):

ogarza:
Thanks @Deoxizn, this fixed a real bug, confirmed with @l0gicgate on the behavoir change.

ogarza:
Pushed the owner-reviewed final behavior in two commits:
- `1f7a1c8`: immediate folder creation and folder rename finalization.
- `d1edb6c`: matching file behavior, shared atomic creation/rename handling, and removal of the obsolete creation placeholders.

New items now exist immediately as `new folder` / `new file` (first free numbered alternative on conflict). Valid Enter/click-away renames stick; Escape or invalid input preserves the original/default name and contents. This supersedes the earlier cancellation-only iteration. Existing file-stem selection is retained.

The final implementation also covers pending-editor cancellation/navigation, deferred GTK focus handling, and scrolling the created item into view in large directories. The PR description, issue, documentation, and screenshots now describe this final policy.

The owner verified folders manually in all three views, subsequently reviewed the file functionality, and approved this update. Fresh pre-push checks passed formatting, Clippy, 1001 Rust tests (19 ignored) under the fully isolated display/session-bus runner, and all 637 pinned-container E2E tests.

ogarza:
Merged current main (`9c01639`) in `cd6a93a` and resolved the conflicts while preserving the reviewed file/folder creation and rename behavior. The shared inline-edit handling now lives in the new window-composition input wiring; current filtered-preview and delayed-loading behavior is retained. The keyboard fixture now waits for its first rendered row before testing F2.

Local verification passed formatting, Clippy, 1061 Rust tests (20 ignored) under the private display/session-bus runner, and all 674 pinned-container E2E tests. GitHub reports the PR as conflict-free and CI is now running. Review approval remains required.

l0gicgate (approved):

PR #587, opened 2026-09-08
fix(browser): treat dropping a folder onto itself as a no-op in all three views

### LINKED ISSUE #424: fix(browser): treat dropping a folder onto itself as a no-op in all three views
## Strata version

0.11.1 (source, `92a5953`)

## Installation method

Built from source

Arch Linux (Omarchy), Hyprland

## What happened?

Dragging a folder in Miller, grid, or explorer and dropping it on itself starts a move. Large folders make this obvious because the “Moving items” dialog appears while the tree is walked.

## Steps to reproduce

1. Open a folder that contains a large subdirectory.
2. Drag that subdirectory and drop it on itself (Miller, grid, or explorer).
3. Observe the move progress dialog.

## Expected behavior

A self-drop is a no-op: no transfer, no dialog.

## Screenshots or video

https://github.com/user-attachments/assets/bcf7d67b-44c0-41e0-84a0-8009083fea2b

## Relevant logs

N/A

## Additional context

Folder rows/cards are drop targets (`src/ui/browser/columns/rows.rs`, `install_list_drag_drop` in `src/ui/browser_modes.rs`) and never reject a self-drop. `start_transfer` already knows this is not a collision (`transfer_has_collision` early-return) but still queues the item, so `TransferStarted` shows the dialog. The adapter’s `transfer_is_noop` skip runs only after `transfer_sizes` has walked the tree.

One guard in `start_transfer` that drops no-op sources (and returns if none remain) would cover Miller, grid, explorer, pane, and sidebar drops. Not a duplicate of #395.

spandan11106:
## Description

Dragging a folder in Miller, grid, or explorer and dropping it on itself (or back onto its own parent, or onto one of its own descendants) started a real transfer: `ViewState::start_transfer` only skipped sources that collided with an existing destination item, so a self-drop was queued, its tree walked for size, and the "Moving items" dialog shown for a transfer that changes nothing.

All three views' drops funnel through the same `transfer_handler` closure into `ViewState::start_transfer`, so one guard there covers Miller, grid, explorer, pane, and sidebar drops. `start_transfer` now filters out sources that are a no-op relative to the destination (dropped onto itself, onto its own current parent when moving, or onto one of its own descendants) before anything is queued, and returns early if none remain. Same-folder copies still reach the adapter to create numbered duplicates.

## Visual evidence

[Current-build screenshots: Duplicate before/after the action and folder self-drop](https://github.com/spandan11106/strata/blob/ade9f810d7d701eab17d9a84d6e39035810bc8f3/docs/screenshots/587/README.md).

## How to test

1. Open a folder that contains a large subdirectory.
2. Drag that subdirectory and drop it on itself, in each of Miller, grid, and explorer view.

**Expected result:** Nothing happens — no "Moving items" dialog, no transfer.

3. Select a file and press Ctrl+D, or copy and paste it into its current folder.

**Expected result:** The original remains and a numbered copy is created.

## Related issue

Closes #424


wmfeht:
## Merge Blocker

`ViewState::start_transfer` now filters every source that already lives in the destination folder via `transfer_is_noop`, including **copies**. That silently breaks Ctrl+D Duplicate, paste into the current folder, Copy-to the current folder, and Ctrl-drop onto the parent / pane background — those never reach the adapter, which is supposed to treat same-folder copy as a rename (`photo.jpg` → `photo (1).jpg`).

Self-drop and descendant-drop filtering are fine. Parent-drop must be a no-op **only for moves**.

Fix: treat `source.equal(&target)` as a no-op only when `move_sources` is true (match the adapter’s `is_duplicate` carve-out). Update the new unit test so parent+copy is not asserted as a no-op.

spandan11106:
Good catch — fixed. `transfer_is_noop` now only treats the same-parent case (\`source.equal(&target)\`) as a no-op when \`move_sources\` is true, matching the adapter's \`is_duplicate\` carve-out; self-drop and descendant-drop stay unconditional no-ops as before. Updated the unit test to assert parent+copy is not a no-op while parent+move still is. Pushed as 127fa31.

l0gicgate:
Merged the latest `main` and the concurrent duplicate-copy fix without rewriting the branch history. Same-folder copies remain enabled; self/descendant drops and moves back to the parent are filtered before transfer progress starts.

Added coverage for mixed valid/no-op selections, files and folders, local/file/SFTP locations, and Duplicate/paste across all three views. Simplified the guard and removed redundant added comments.

GUI verification in the pinned container: Ctrl+D and Copy to the current folder created numbered copies with intact contents; folder self-drop left the filesystem unchanged in Columns, Icons, and List.

Current-build captures ([all captions and before-action image](https://github.com/spandan11106/strata/blob/ade9f810d7d701eab17d9a84d6e39035810bc8f3/docs/screenshots/587/README.md)):

| After Ctrl+D | After folder self-drop |
| --- | --- |
| Numbered duplicate alongside the original | Folder self-drop with no transfer dialog |

PR #591, opened 2026-09-08
fix(rename): keep the Columns caret inside the visible viewport

### LINKED ISSUE #394: fix(rename): keep the caret visible when editing long filenames
Split from the grouped rename and copy/paste report ([#354](https://github.com/lgse/strata/issues/354), item 1 of 5).

## Strata version

Reproduced on Strata 0.10.0, debug build at commit `183d744`, GTK 4.14.5 on X11, isolated fixture `HOME`/XDG dirs.

## What happened?

When renaming a file whose name is wider than the available column name cell, the inline rename field does not scroll horizontally to keep the caret visible. Editing at the end of a long name happens off-screen.

**Reproduced in Columns view only.** Grid and Explorer scroll correctly after End so the caret and trailing text stay visible. Columns uses `BrowserState::begin_rename` in `src/ui/browser.rs`; Grid/Explorer use the mode-view path in `src/ui/browser_modes.rs`.

Also observed in Columns: on F2, stem-only selection is applied but the field is already scrolled toward the stem end, so the *start* of the name is not visible either (issue text assumed the start would be visible).

## Steps to reproduce

1. Open a directory in Columns view.
2. Select a file whose name is wider than the column (e.g. `synthetic-quarterly-report-with-a-very-long-descriptive-basename-2026.txt`).
3. Start inline rename (`F2`).
4. Move the caret to the end of the filename (`End`).
5. Observe that the entry does not scroll and the caret is off-screen.
6. Type a few characters; the field may nudge only slightly while the caret and typed text stay off-screen.
7. Repeat in Grid and Explorer — those views scroll correctly (contrast).

## Expected behavior

The rename field keeps the caret visible for long names, including when the caret is at the end of the name. Stem-only initial selection is unchanged. This must work in Columns (and remain correct in Grid / Explorer).

## Screenshots or video

Repro recording (Columns fails; Grid/Explorer OK):

https://cursor.com/artifacts/v/art-09ec12ca-2f1d-4ebb-90ea-b04aab79a98f

Keyframe sequence from Columns (zoomed on the rename entry):

1. After F2 — stem selected but field scrolled to stem end (`-descriptive-basename-2026` visible; name start hidden).
2. After End — no scroll; caret past `.txt` is off-screen (same visible slice).
3. After typing — only ~2 characters of scroll (`scriptive-basename-2026.tx…`); typed text and caret still off-screen.
4. After Escape — rename cancelled; original label restored (`synthetic-quarterly-report-…`).

## Additional context

Inline rename is a `gtk::Entry` with class `inline-rename`. If the entry is allowed to take the natural width of the full name, the parent row/card/cell clips it. GTK then treats the caret as on-widget even though it is off-screen. Constraining the entry to the visible editor width so its inner `GtkText` scrolls, then ensuring caret and selection are scrolled into view after `select_region` and on later caret moves, is the likely fix path for Columns.

Existing coverage only checks stem vs extension; there is no test that a long name stays caret-visible.

### Acceptance criteria

- [ ] In Columns (and still in Grid / Explorer), the rename field scrolls as needed to keep the caret visible for names wider than the editor.
- [ ] Moving the caret to the end of a long name (End / Right) brings the caret on-screen without changing stem-only initial selection.
- [ ] Typing, deleting, and clicking inside a long name keep the caret visible.
- [ ] Automated tests cover long-name rename caret visibility (or an equivalent constrained-width editor assertion) in addition to stem selection.

spandan11106:
## Description

Fix Columns rename when the column is wider than the visible browser area. Constrain the editor to that visible area and follow viewport changes while editing, so GTK can keep the caret on-screen without changing stem-only selection.

Replace the vacuous regression checks with non-zero allocation, overflowing-text, and browser-relative caret assertions, plus real F2/End, typing, deletion, arrow-key, and pointer scenarios.

## Visual evidence

| Before End fix | After End fix |
| --- | --- |
| Extension and caret clipped | Extension and caret visible |

[Capture details and List/Icons comparisons](https://github.com/spandan11106/strata/blob/ff8f54ac9a03557f531be577d6e03f78fe3e1501/docs/screenshots/394/README.md).

## How to test

1. Create `synthetic-quarterly-report-with-a-very-long-descriptive-basename-2026.txt` in a temporary directory.
2. Open it in Columns with the sidebar visible and a narrow window (420×300 reproduces the reported clipping). Select the file and press F2; only its stem should be selected.
3. Press End, type a suffix, delete it, click inside the name, and try Left/Right and End again. Resize the window while editing.
4. Compare List and Icons at a width that accommodates their minimum table/card sizes, such as 640×300 with the sidebar open.

**Expected result:** the caret and extension stay visible after End, editing stays on-screen, and stem-only initial selection is unchanged.

## Related issue

Closes #394


spandan11106:
image

wmfeht:
## Merge Blocker

This PR is test-only and should not close #394. Columns rename still fails the user-facing acceptance: in a 420×300 window, after F2 and after End the editor stays on the stem-end slice (`.txt` never scrolls into view). List and Icons already keep the caret on `.txt` after End.

The new helper never presses End (it uses `set_position(-1)` on a headless BrowserView), does not assert that the name overflows the editor, and can pass vacuously when width is 0. Drop `Closes #394`, or fix Columns End in the real UI and add overflow + non-zero-width preconditions before claiming the issue is done.

spandan11106:
The scroll bug is already fixed — it went away when the rename code got consolidated into the shared `inline_edit` module, before this PR. Re-verified directly: in a real 420×300 window, after F2 and End the caret sits at the visible edge of the field (e.g. x=252 of a 253px-wide field) in Columns, matching List and Icons pixel-for-pixel in relative terms. No off-screen caret in any mode.

This PR is test-only because there's nothing left to fix — it just adds the regression test the issue's acceptance criteria called for. `Closes #394` is accurate, not premature.

l0gicgate:
@wmfeht addressed in ff8f54a. This now includes a production fix, not just tests: the Columns editor is constrained to the visible viewport, including resize/scroll changes, so GTK can bring the caret and `.txt` into view after End. Stem-only selection is preserved.

The Rust fixture now requires mapped, non-zero-width editors and overflowing text, and checks the caret against the browser bounds as well as GtkText. The real-application scenario sends F2, End, typing, Backspace, Left/Right, and inside-field clicks. Disabling the constraint makes the Columns scenario fail after End (306px editor versus 210px visible browser area); List/Icons comparisons still pass. Before/after screenshots and their capture conditions are in the updated description.

One separate finding from an extra GTK 4.14 Rust-suite run: the existing `new_entries_scroll_into_view_when_the_default_name_sorts_past_the_initial_viewport` test times out with 2000 entries. I reproduced that failure with both PR production changes removed, so it is not introduced here. The new caret fixture passes on GTK 4.14 and 4.22, and all required local gates pass.

Ready for another look at the narrow-window Columns case; keeping `Closes #394` now that the actual clipping is fixed.

PR #599, opened 2026-09-08
fix(browser): offer keep-both for cross-folder paste conflicts

### LINKED ISSUE #395: fix(file-operations): unify same-folder and cross-folder paste
Split from the grouped rename and copy/paste report ([#354](https://github.com/lgse/strata/issues/354), item 5 of 5).

Sibling issues already filed for other slices of that report: copy undo ([#390](https://github.com/lgse/strata/issues/390)), conflict dialog **X** ([#391](https://github.com/lgse/strata/issues/391)), destination selection ([#392](https://github.com/lgse/strata/issues/392)), rename caret ([#394](https://github.com/lgse/strata/issues/394)). This issue is the remaining product inconsistency between pasting into the current folder and pasting into a different folder — naming / collision policy — not those completion behaviors.

## Strata version

Reproduced on Strata debug build at commit `183d744` with fixture `source/report.txt` and `dest/report.txt`.

## What happened?

Copy/paste does not behave the same way when the destination is the current folder versus a different folder.

In practice:

- Pasting a file into the folder it already occupies silently creates `name (1).ext` (then `(2)`, …) and never asks.
- Pasting that same file into another folder that already has `name.ext` raises “File already exists” with Replace / Skip / Cancel, and has no “keep both / unique name” choice.

That makes it hard to predict whether a paste will auto-rename, prompt, replace, or no-op.

## Steps to reproduce

Use a writable fixture with `source/report.txt` and `dest/report.txt` (same basename, different files).

### Same-folder paste (auto-duplicate)

1. Open `source/`.
2. Copy `report.txt` and paste into `source/`.
3. Observe a new `report (1).txt` with no dialog.

### Cross-folder paste (conflict dialog)

1. Copy `source/report.txt`.
2. Open `dest/` and paste.
3. Observe the “File already exists” dialog instead of a generated unique name.
4. Choose Replace, Skip, and Cancel on separate runs and compare outcomes.

## Expected behavior

Same-folder and cross-folder paste share consistent naming and collision policy (auto unique name, prompt, or a prompt that includes “keep both”), and share the same dialog actions when a prompt is required.

Duplicate and clipboard paste into the source’s parent must stay on that shared path so Ctrl+V in the current folder and an explicit duplicate do not diverge again.

Selection / undo / dialog **X** belong to the sibling issues above; the unified paste path should use those once they exist.

## Screenshots or video

Full walkthrough (~91s) — same-folder unique name, conflict dialog Cancel/Skip, Miller context check:

https://cursor.com/artifacts/v/art-86a1f3a7-9cee-47d1-8ac9-a381187a9b2b

Clean side-by-side (~28s) — silent `report (1).txt`, then conflict dialog → Replace:

https://cursor.com/artifacts/v/art-785334c0-28ba-46e3-b57f-36e626141119

## Additional context

Paste enters `BrowserState::start_transfer`. `transfer_has_collision` returns false when the source equals the would-be target, so same-folder pastes bypass the dialog and take the path where `local_operations` allocates `name (1).ext` via `duplicate_target`. Only genuine cross-folder collisions reach `confirm_replace_conflict`, which offers Replace / Skip / Cancel with no generate-unique-name choice.

`TransferConflict` only has `FailIfExists` and `ReplaceExisting` — there is no keep-both / unique-name option for the cross-folder dialog.

### Acceptance criteria

- [ ] Same-folder and cross-folder paste use one naming and collision policy, including generated unique names and conflict-resolved names.
- [ ] Duplicate, clipboard paste into the source folder, and clipboard paste into another folder share completion, selection, and undo behavior (via the sibling fixes).
- [ ] The conflict dialog, when shown, offers the same actions regardless of which path produced the collision; Cancel / Escape / **X** do not modify files.
- [ ] Automated tests cover: same-folder generated names, cross-folder unique-name or replace/skip outcomes under the chosen policy, and mixed multi-item pastes.

spandan11106:
## Description

Same-folder copy paste silently generates a unique name (`name (1).ext`) on a collision and never prompts, while cross-folder paste only offers Replace/Skip/Cancel with no way to keep both copies. This adds a `KeepBoth` conflict option, reusing the existing unique-name generation logic, and surfaces it as a "Keep Both" button in the cross-folder conflict dialog for copy paste.

Move operations (cut/paste, drag-move) keep their prior Replace/Skip/Cancel behavior only — Keep Both is not offered there, because undo and post-move reveal currently assume a move never renames its destination; offering it there would need separate follow-up work to carry the renamed target through undo/reveal tracking.

## Visual evidence

N/A from me directly — this only adds one new button ("Keep Both") to the existing "File already exists" conflict dialog, no other layout changes. I don't have a way to capture a screenshot from this environment; happy to attach one if needed.

## How to test

1. Create two folders with a same-named, differently-contented file, e.g. `source/report.txt` and `dest/report.txt`.
2. Copy `report.txt` from `source/` and paste into `source/` again.
3. Copy `report.txt` from `source/`, navigate to `dest/`, and paste.
4. Cut `report.txt` from `source/` (Ctrl+X), navigate to `dest/`, and paste.

**Expected result:**
- Step 2 (same-folder copy): unchanged — `report (1).txt` is created silently, no dialog.
- Step 3 (cross-folder copy): the conflict dialog now offers Skip / Keep Both / Replace / Cancel. Choosing Keep Both creates `report (1).txt` in `dest/` without touching the existing file.
- Step 4 (cross-folder move): the conflict dialog offers only Skip / Replace / Cancel, same as before — Keep Both is not shown for moves.
- Cancel/Escape/the dialog's close button never modify any file, for either dialog.

## Related issue

Closes #395


spandan11106:
image

l0gicgate:
Merged `main` (`c62354f`) and fixed two review blockers:

- Enter now activates the focused conflict action instead of always choosing Replace—even when **Keep Both** was focused.
- Keep Both no longer bypasses the backend's self/descendant-transfer no-op guard.

Added regression coverage for keyboard/pointer Keep Both, numbered-copy selection and undo, mixed apply-to-all pastes, and copy/move cancellation. Audited every added comment and removed or tightened redundant prose.

Exercised the real GTK application on isolated Xvfb: Shift+Tab to Keep Both, then Enter creates and selects the numbered copy without changing the existing file. Pointer activation, undo, and Cancel/Escape/close also preserve the expected files.

GitHub still requires approval from someone other than the last pusher.

Captured current dialog/result screenshots, but this session has no GitHub attachment-upload tooling; these files were **not uploaded**:
- `/tmp/strata-599-review-media/keep-both-keyboard-focus.png`
- `/tmp/strata-599-review-media/keep-both-result.png`

Issue #550, opened 2026-09-07
chore(quality): track CodeScene setup and Code Health improvements


l0gicgate:
## Problem or use case

We need one living tracker for Strata's CodeScene setup and code-quality improvements, with focused issues/PRs underneath it. Improve maintainability and frequently changed code—not just the displayed score. This issue is the progress source of truth; durable design decisions belong in `docs/architecture.md`.

### Verified Cloud baseline

[Strata project](https://codescene.io/projects/84395) · [baseline analysis](https://codescene.io/projects/84395/jobs/7478387/results)

| Metric | Baseline |
| --- | --- |
| Analysis time | 2026-09-07 14:56:33 UTC |
| Analyzed commit | `83e4d7f062079da786e0108b7d55281fe7dc27a6` |
| Overall Code Health | **6.76 / 10** |
| Hotspot Code Health | **4.10 / 10** |
| Analyzed files / files with Code Health | 247 / 242 |

At the initial review, architectural components and hotspot goals were empty, and PR safeguards reported **zero PR analyses** between August 7 and September 7. Zero analyses is not evidence of clean PRs. Month/year comparison values are unavailable, not zero.

### Latest verified Cloud analysis

[Analysis 7506640](https://codescene.io/projects/84395/jobs/7506640/results), completed **2026-09-09T16:49:51Z**, analyzed `a4dcfaae66be4a385351b6925235917b136e8bc2` on `main`, verified to contain #683. Overall **7.46**, hotspot **4.97**, scope **347 / 340** analyzed/scored files. Compared with 7497338, overall **7.54 → 7.46** and hotspot **5.03 → 4.97**; this spans many later changes, not just #683.

**#683 is merged and Cloud verified:** `inline_edit.rs` **4.96 → 5.45**, existing tests unchanged at **6.54**, new lifecycle tests **10.00**. These three blobs are unchanged between #683 and this analysis. Exact first-parent CLI comparison confirms reveal complexity **38 → 21**. Remaining `begin_rename_item` complexity **21** is the next bounded target; focused follow-up [#710](https://github.com/lgse/strata/pull/710) is open, with local CLI implementation score **5.45 → 5.77**; not merged or Cloud verified yet.

### Previous Cloud checkpoint (7497338)

[...]

PR #601, opened 2026-09-08
refactor(browser): simplify staged row publication

l0gicgate:
## Description

Separate staged row publication from the application browser controller. An owned publication plan carries request identity, row count, selection, and terminal data; streaming progress stays separate. Share selection/completion dispatch across inline publication, idle tails, and synchronous draining, and separate chunk calculation from observer calls.

Preserve prefix/chunk limits, idle priority, cooperative time budgets, cancellation, and event order. Sort paths still capture their publication data before preference observers run. Ten focused tests cover these boundaries; sorting workers, remote coalescing, and operation callbacks remain separate follow-ups.

Local CodeScene CLI 1.0.40 against `ed5e1bd`: `app/browser.rs` **3.94 → 4.04**; publication implementation/tests **10.00**. This removes the complex tail-dispatch and argument-heavy publication findings, not the controller's remaining debt.

## Visual evidence

N/A: internal behavior-preserving refactor; no intended visual changes or baseline updates.

## How to test

1. Open a directory containing several thousand files. Check that the first rows appear promptly and the rest arrive without losing the final selection.
2. Select a file near the end, change sorting (Name, Size, Modified), and switch between Columns, Icons, and List. Check that selection follows the same file and loading/sorting indicators settle.
3. Refresh or navigate away while rows are arriving. Return to the directory and create or rename a file; check that the listing converges without stale rows or a stuck indicator.

**Expected result:** Unchanged row ordering, selection restoration, and completion behavior; superseded work does not publish into the current directory.

## Related issue

Refs #550 (partial initiative; keep the umbrella open).

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -171,8 +171,26 @@ stale work is still checked against the owning directory or peek request.
 validated against directory identity and row-position/location tokens. Metadata chunks
 never complete a sort; `MetadataFinished` retains that responsibility. Both modules release
 state and routing borrows before synchronous observer dispatch, allowing observers to
-navigate safely. Sorting, publication budgets, timers, and cancellation remain in the
-browser controller; this extraction does not change those policies.
+navigate safely. Sorting, metadata scheduling, and cancellation remain in the browser
+controller; event routing does not change those policies.
+
+### Browser staged publication
+
+`app/browser/publication.rs` owns staged row publication on the same `Browser`, not a
+second controller. A `PublicationPlan` captures request identity, row count, selection,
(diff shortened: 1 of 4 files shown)


l0gicgate:
Cloud verification and the refreshed execution plan are recorded in #550. Analysis [7483415](https://codescene.io/projects/84395/jobs/7483415/results) analyzed merge revision 0bd3870: app/browser.rs is 4.04, and publication implementation/tests are 10.00. Overall health is 7.48 and hotspot health 4.98 across all intervening commits. Next is alternate-mode event handling (browser_modes.rs 2.75), while active operation-policy fixes remain separate.

PR #605, opened 2026-09-08
fix(archives): bound extraction to free space and declared size

### LINKED ISSUE #479: security(archives): extraction has no size, ratio, or entry-count limits (zip bomb)
### Strata version

`main` / skills-branch review (source review; extraction path present in current tree)

### Installation method

Built from source

N/A (source / design review; affects Extract here / Extract to…)

### What happened?

**Severity: Medium** (confirmed in security review)

Archive extraction writes decompressed bytes until EOF with **no cap** on:

- total uncompressed size
- compression ratio
- free space at the destination
- entry count (count is used for progress only)

A small malicious archive (“zip bomb”) extracted via **Extract here** / **Extract to…** can expand until the disk is full — local resource-exhaustion DoS.

Affected helpers share `copy_with_big_buf` (~`src/adapters/local_operations.rs:3655`), used by zip / tar / 7z extract paths.

### Steps to reproduce

1. Obtain or craft a highly compressible archive that expands to a very large size (or many entries).
2. In Strata, use **Extract here** or **Extract to…** on that archive.
3. Observe extraction continuing without aborting on size / ratio / free-space / entry limits until disk pressure or failure.

### Expected behavior

- Track cumulative uncompressed bytes across entries and abort past a ceiling (fixed max and/or a multiple of free space from `statvfs` at the destination).
- Optionally cap entry count and reject extreme compression ratios.

### Screenshots or video

Source review evidence (security-review skill, agent `bc-295dd478-2da2-477b-9f81-d852102b7138`):

```rust
// copy_with_big_buf loop — total counted for progress, never bounded
loop {
    ...
    total += n as u64;
}
```

Used by `extract_zip_from_archive`, `extract_tar`, `extract_7z_from_reader`.

Note: path traversal / symlink recreation in archives was reviewed separately and found hardened; this issue is specifically **resource exhaustion**, not path escape.

### Additional context

- Finding ID from review: **VULN-002**
- Confidence: High

wmfeht:
## Description

Archive extraction previously copied decompressed bytes until EOF with no check against destination free space or the uncompressed size the archive claimed. A small malicious archive could expand until the disk was full ([#479](https://github.com/lgse/strata/issues/479)).

After looking at how other file managers handle this, they generally do **not** guard against zip-bomb attacks beyond letting the user cancel, or verifying that the bytes actually extracted match what the archive header claims. This change follows that model rather than adding ratio or entry-count caps.

Extraction now:

- Queries unprivileged free space with `fstatvfs` on the pinned destination root and refuses a claimed total that would not fit
- Copies at most the declared uncompressed size and remaining free space, then fails if the stream overruns or undershoots the header
- Preflights ZIP and 7z from the advertised archive total so an oversized archive is refused before any member is written
- Checks TAR members as they arrive, so extraction stops at the free-space boundary and members already written stay in place
- Skips the free-space checks on filesystems that report no capacity (`f_blocks == 0`, including some FUSE mounts), while still enforcing the declared-size match

The `zip` crate only bounds compressed input, not inflated output, so a ZIP member that lies about its size is stopped by that match. Highly compressible archives whose headers are truthful still extract in full; cancellation continues to work as before.

## Visual evidence

N/A — this is extract-path hardening, not a layout change. Failures use the existing extract error path.

## How to test

1. Extract a normal ZIP, tar/tar.gz, and 7z with **Extract here** / **Extract to…**. Contents should extract as before.
2. Extract a highly compressible but truthful archive (for example a large file of zeros). It should complete; there is no ratio cap.
3. Cancel mid-extract and confirm the current member is removed when possible and remaining members are reported as not attempted.
4. Optional: extract onto a nearly full volume, or a ZIP whose uncompressed-size header has been patched smaller than the payload. Extraction should fail with a declared-size or free-space error instead of filling the disk.

Unit coverage is in `extraction/tests.rs` (preflight, remaining space, overflow, shortfall, unreported capacity) and `decoders/tests.rs` (ZIP members that lie about size, highly compressible ZIP/7z/tar.gz still extracting).

**Expected result:** Legitimate archives extract. Archives that claim more than free space, or whose members do not match their declared size, fail closed without leaving the overflowing member on disk. The user can still cancel.

## Related issue

Closes #479


wmfeht:
Local verification on `753dfd9` (`fix/479-prevent-zip-bombs`):

- `./scripts/check.sh` passed (fmt, check, clippy `-D warnings`, tests, cargo-deny, typos)
  - `cargo test --all-targets --all-features`: 1090 passed, 20 ignored, 0 failed
  - cargo-deny: advisories / bans / licenses / sources ok
- `./scripts/e2e.sh` passed: **707 passed** in 172.30s (12 workers)

Host toolchain: rustc 1.98.0

l0gicgate:
Merged the latest main and completed the extraction review. Removed redundant comments and doc comments while retaining the pinned-root safety, filesystem compatibility, sequential-format, and extra-byte-probe rationale. No functional blockers found. The isolated local quality and E2E suites passed. No UI changes or separate manual GUI verification were made; visual evidence is not applicable.

ogarza (approved):

l0gicgate:
Review verdict: **ready to merge; no blocking issues found.**

Extraction enforces declared member sizes and the destination’s initial free-space budget. Oversized or truncated members fail with partial-file cleanup, while truthful highly compressed archives remain supported. Coverage includes dishonest ZIP headers, capacity limits, size mismatches, and filesystems that do not report capacity.

I merged current main and removed redundant comments; no functional fixes were needed.

Scope limitation: this is not comprehensive zip-bomb protection. There are no ratio or entry-count caps, and the initial free-space snapshot does not account for concurrent writes, block rounding, or inode exhaustion. This matches the documented scope.

Local quality passed (1,150 tests passed, 20 ignored), E2E passed (655 tests), and GitHub checks are green at e6ea2c9. No separate manual GUI verification was performed. No UI changes were made; visual evidence is not applicable.

PR #616, opened 2026-09-08
fix(browser): reveal archive outputs and keep selection visible during context menu

### LINKED ISSUE #611: fix(browser): reveal archive outputs and keep selection visible during context menu
### Strata version

0.13.0 source build at commit `916095a`

### Installation method

Built from source

Arch Linux, Hyprland / Wayland (tests run under an isolated Xvfb X11 display)

### What happened?

Two related selection/visibility bugs in the browser views:

**1. Archive output not scrolled into view after compression**

After Compress finishes, Strata refreshes the directory and selects the new archive, but the archive is not reliably scrolled into the visible viewport. In a directory with enough entries to require scrolling, users can finish compression without seeing the output file.

**2. Right-click in Columns view does not keep the prior selection highlighted**

When right-clicking a folder in Columns view to open the context menu, the previously selected row loses its highlighted state because the column no longer has focus-within. The selection is still active internally, but the visual feedback disappears while the menu is open.

### Steps to reproduce

**Archive reveal:**

1. Open a directory containing enough entries to require vertical scrolling.
2. Select a file or folder near the top of the listing.
3. Choose **Compress…**, give the archive a name that sorts below the visible entries, and complete compression.
4. Wait for the directory to refresh.
5. Observe that the archive can be selected outside the visible viewport.

**Context-menu highlight:**

1. Switch to Columns view.
2. Select an entry in a column.
3. Right-click another folder to open the context menu.
4. Observe the previously selected row loses its highlighted background while the menu is open.

### Expected behavior

- When compression or extraction completes and Strata selects the resulting archive/output, the active Columns, Icons/Grid, or List/Explorer view automatically scrolls far enough to reveal the selected item.
- While a context menu is open, the prior selection stays visually highlighted in the column even though the column no longer has focus-within.

### Screenshots or video

N/A — behavioral fixes; automated GTK regression tests can capture selection state and viewport position.

### Relevant logs

No application errors were reported.

### Additional context

`BrowserEvent::ArchiveCompleted` queues the output name and reloads the active directory. `LoadFinished` selects it through `select_entries_by_name`, so the intended reveal relies on the selection event reaching each collection after reload/allocation.

The context-menu highlight is governed by the `:not(:focus-within)` CSS rule on `.directory-column` and `.file-list-mode`; adding a `context-selection` class while the popover is open keeps the selection visible without focus.

JoeJoeflyn:
## Summary

- Track the archive destination so the browser navigates back to it once compression completes, and reveal the focused entry after selection restoration across Columns, Icons/Grid, and List/Explorer views.
- Add a `context-selection` CSS class toggled on popover show/close so right-click context menus keep the prior selection highlighted in Columns view while the column lacks focus-within.
- Add a layout-driven `reveal_collection_after_layout` helper that materializes virtualized rows before scrolling, used by both the archive reveal and general selection restoration.
- Add e2e regression coverage for the archive reveal across all view modes, including Columns edge cases (active child, deep nesting, non-active parent).

Closes #611.

#### Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo test --all-targets --all-features` under isolated Xvfb
- [x] `./scripts/e2e.sh` (CI container)
- [x] Manual: compress a file in a long directory and confirm the new archive scrolls into view in each view mode
- [x] Manual: right-click a folder in Columns view and confirm the prior selection stays highlighted while the menu is open
screenrecording-2026-09-08_16-45-55


wmfeht:
## Merge Blocker

Head `f46f296`:

**`pending_archive_destination` is not cleared when compression is cancelled (or when reveal never succeeds).** The flag is set as soon as compress starts. Cancel emits empty-name `ArchiveCompleted` then `OperationCancelled` — neither takes the flag (`OperationFailed` does). While it stays set, Columns `LoadFinished` skips the shared restore path used by paste reveal / `pending_select`, so a later paste in the same miller path is not selected. On reveal miss, `reveal_pending_archive_at` can `retry_column` with no cap.

Fix: take `pending_archive_destination` on `OperationCancelled`, `OperationCompletedWithErrors`, empty-name `ArchiveCompleted`, and `TransferReveal`. On reveal miss, wait for publication with at most one bounded retry — do not loop `retry_column`. Add a GTK unit test that cancel/empty-name completion clears the flag and a following `LoadFinished` still restores a transfer selection.

wmfeht:

Agent: https://cursor.com/agents/bc-f421634d-90f0-47a9-b2c2-02139b569659  
Head: `528ada81e0b76251d67a1c74569ce4ec2a8d0790` (pinned, clean)

**Result: clean pass — no confirmed candidate issues.**

Checked: compress/extract reveal + scroll-into-view (Columns/List/Icons, child/parent columns); selection stays visible under context menu (single/multi); cancel compress then paste restore; `./scripts/test-headless.py events::tests` — 13 passed. Prior cancel-leak path did not reproduce.

PR #619, opened 2026-09-08
fix(browser): preserve submitted names while renames complete

### LINKED ISSUE #617: fix(browser): prevent old-name flash when committing inline renames
### Strata version
Reported against the current development checkout (0.13.0; cd6a93a). Exact running application version not confirmed.

### Installation method
Development/source checkout; reporter’s running installation method not confirmed.

Omarchy (Arch Linux), Hyprland.

### What happened?
When committing an inline file or folder rename by clicking elsewhere, the name briefly switches back to the old name before settling on the new name.

### Steps to reproduce
1. Start renaming a file or folder.
2. Type a different valid name.
3. Click outside the editor.
4. Watch the item label while the rename completes. Repeat for files and folders in each browser view.

### Expected behavior
Click-away commits the edit and the new name stays visible without an intermediate old-name flash. Escape continues to cancel. Failed or invalid renames must not leave a misleading name displayed.

### Screenshots or video
Not yet captured.

### Relevant logs
No logs supplied; no crash reported.

### Additional context
The rename editor appears to restore the stale label before the deferred filesystem operation refreshes the listing.

ogarza:
## Description

Keep the submitted file or folder name visible after click-away or Enter, including while rows are rebound during refresh. Track the pending rename through success, failure, cancellation, and superseding operations so stale labels do not flash and failed edits roll back. Preserve Escape cancellation before submission.

Also keep the hidden-file filter installed when restoring the directory model, preventing stale row labels during synchronous rebinding.

## Visual evidence

Before and after: pending file and folder renames

Private-Xvfb widget captures compare base `916095a` with this fix. A delayed rename provider holds the normally brief pending state for both file and folder screenshots.

## How to test

1. In Columns, List, and Icons, rename a file and a folder, typing a valid new name and clicking another item to commit. Repeat using Enter and with newly created items.
2. Try Escape before submitting, an invalid name, and an existing destination name.

**Expected result:** Committed names stay visible without flashing back to the original. Escape and invalid input retain the original name; failed operations restore it and report the error. Subsequent renames continue working.

## Related issue

Closes #617


ogarza:
Addressed the architecture and test review:

- Replace coupled rename flags with explicit lifecycle states.
- Remove the generic operation-start event; guard queued dispatch with a central operation token.
- Store weak bound-label handles instead of traversing production widget layouts.
- Exercise delayed success/failure, synchronous completion/cancellation, refresh replacement, and wrong/stale callbacks through providers.
- Find visible labels independently in tests and force unrelated-entry → original-entry rebinding. Removing the pending-name bind override causes the regression test to fail independently in Columns, List, and Icons (mutations restored).

The larger test fixture replaces direct state manipulation with protocol-level coverage. Full local Rust checks and all 729 container E2E scenarios pass.

wmfeht:
## Merge Blocker

Head `a2aceff` (prior `d42f52e` cancel/reload blockers look fixed):

**Pending rename can stick after a Columns folder rename.** Renaming the selected open folder rebases via native `DirectoryChange::Move` → `restore_path` (Reset + ColumnAdded) without splicing the parent. If the monitor fires while state is still `Running`, those loads are not recorded; `complete_pending_rename` then enters `AwaitingRefresh` with an empty request list, so LoadFinished/EntriesSpliced never clear pending. `begin_rename` refuses F2 and auto-refresh stays skipped until the parent leaves the column stack.

Fix: on miller `restore_path` / Reset while rename is Running or AwaitingRefresh, record the new load IDs, treat path restore as completion when the rebound parent already shows `new_location`, or otherwise drop/timeout the stuck pending. Add a miller-folder monitor-before-completion test (single-column fixtures miss this path).

l0gicgate:
## Guided manual testing and review

Reviewed head: `d42f52e6fbb19c8f13bd41a24cf4d22ec200fb65`.

### Methodology and observed results

Created an isolated temporary fixture containing `original.txt`, `other.txt`, and `Original Folder`. Instructions were to quit the installed/running Strata instance and launch this checkout with `cargo run -- <fixture-directory>`. The user exercised the UI interactively and reported results; these are user-reported observations, not automated measurements.

- **Columns:** file and folder renames submitted by clicking another item; Enter submission; Escape before submission; navigate away/return to check persistence. Reported pass, no old-name flashing.
- **Conflict and recovery:** attempt to rename onto `other.txt`, decline replacement if prompted, then perform a valid rename. Reported pass.
- **List and Icons:** click-away file rename, Enter folder rename, and Escape before submission. Both reported pass, no flashing.
- **Newly created items:** in each view, create/name folders using Enter and click-away, rename again, then navigate away/return. User reported all good. New-file repetition was conditional on the UI offering it; separate new-file coverage was not recorded.

### Coverage limits and source review

These happy-path results do **not** resolve the merge-blocker comment. We did not deliberately delay operations, navigate/switch views during pending work, measure loading/blank frames, test invalid basenames, or test active-filter preservation. No automated checks were rerun during this review.

Source inspection supports both concerns in the existing comment:

1. `ViewState::cancel_rename` (`src/ui/browser/inline_edit.rs:642-668`) clears queued pending state and cancels the operation handle for running renames. `truncate` calls it unconditionally (`src/ui/browser/columns.rs:1271`). Thus editor/view teardown can discard an already submitted rename; the queued idle callback requires that pending state to dispatch. Editor dismissal should be separated from cancellation of submitted work.
2. Rename success now refreshes native parents too (`src/app/browser.rs:1974-1982`). `reload_column` clears entries (`src/app/navigation.rs:492-505`), and `ColumnReloaded` clears the visible model and shows loading (`src/ui/browser/events.rs:212-233`). This introduces a full-list reload path rather than just a label update; perceptible duration was not measured in our manual session.

**Additional concrete consequence:** that same `ColumnReloaded` handler clears `filter_entry` and search results. A successful rename in a filtered native Columns listing now discards the user's filter. Preserve the filter as well as the listing when reconciling a rename.

Recommended regressions before landing: delayed/queued submitted rename survives folder navigation, Columns rebuild, and view switching; Escape-before-submit remains cancellation; successful native rename retains visible rows and filter text/results without a loading reset. Current manual passes are useful evidence for label behavior, but not merge clearance for these paths.

l0gicgate (dismissed):

PR #627, opened 2026-09-08
feat(browser): reveal newly created items after naming

### LINKED ISSUE #618: feat(browser): keep newly created items selected and visible after naming
### Problem or use case
When creating a file or folder near the bottom of a directory and naming it something that sorts near the top, the new item can disappear from the viewport. Users should not have to search for the item they just created.

### Proposed solution
After naming a newly created file or folder finishes and the directory refresh/sort settles, retain selection and keyboard focus on that item and scroll only enough to make it fully visible. Leave the viewport unchanged if it is already fully visible. Support Columns, List, and Icons views, descending sorts, and folders-first ordering. Avoid revealing a stale pre-sort position.

### Alternatives considered
Manually searching or scrolling for the item; always centering it (unnecessary viewport movement).

### Visual examples
Reproduction scenario: scroll to the bottom of a long alphabetically sorted directory, create a file or folder, name it a-first, and press Enter. The final item should remain selected and appear onscreen.

### Strata version
Source commit 916095a.

### Additional context
Scope is newly created files and folders only; renaming existing items is not included. Related but distinct: #392 (copy destinations), #611 (archive outputs), #613 (existing GTK 4.14 new-entry fixture failure).

ogarza:
## Description

Reveal and focus the naming editor when creating files and folders, then keep the final item selected and visible after naming in Columns, List, and Icons. Wait for directory updates and GTK layout to settle; preserve the viewport when the item remains visible. Existing-item rename behavior is unchanged.

## Visual evidence

Before submitting the new folder name (scrolled down, naming editor visible):

Before naming

After naming it `a-final` (selected and revealed at its sorted position):

After naming

## How to test

1. Open a directory containing enough files and folders to scroll, sorted by name. Create a file or folder while its default name would land outside the viewport.
2. Confirm the naming editor scrolls fully into view with the default name selected. Type `a-first` immediately and press Enter. Repeat in Columns, List, and Icons, including creation from near the bottom.
3. Repeat with descending sort and with a final name whose sorted position is already visible.

**Expected result:** The naming editor is visible and ready for typing as soon as it opens. After confirmation, the final item stays selected and keyboard-focused. Scroll only enough to reveal it fully; leave the viewport unchanged if it is already visible.

## Related issue

Closes #618


ogarza:
Fixed the follow-up Columns footer overlap in 96857ad. Reveal now measures the unobscured area above the Keyboard/Pointer · Paste here bar and corrects GTK scrolling after row allocation, for both the naming editor and final selected item. No fixed pixel gap is used.

To exercise: create a file or folder in a long Columns listing, verify its naming editor is fully above the paste footer, then give it a name sorting to the bottom and confirm. The selected item content should remain above the footer.

Naming editor:

Editor above paste footer

After confirmation:

Selected item above paste footer

ogarza:
Follow-up in d0b1e01 fixes the reproduced `zzz` case: with numbered default-name collisions and a tall Columns listing, reveal could finish on a transient GTK layout frame and leave the renamed item offscreen. Track the actual file Location across row recycling/sorting, distinguish repeated requests from new targets, and wait for visibility to settle before the selection/focus handoff.

Manual reproduction: in a long directory containing `new file`, `new file (1)` through `(10)`, `y`, and `yy`, create another file and type `zzz`, then press Enter without scrolling. The final item should remain selected and visible above the paste footer.

The new realistic regression fails on 96857ad and passes with this change on both the native private-display run and the pinned container.

ogarza:
Direct follow-up in 6b43c05: the application selected the renamed Location, but `ListView::grab_focus()` restored GTK’s previous keyboard cursor. Merely checking that the list held focus missed this mismatch, including when the renamed item moved upward but stayed inside the viewport.

The completion path now focuses the native list-item widget bound to the final Location and verifies that exact widget owns keyboard focus. No additional scroll retries or pixel offsets were added.

To exercise: create a file in a long Columns listing; name it just above its default position while still visible, then repeat with a name sorting beyond the viewport. After Enter, the selected item, keyboard cursor, and revealed item must all be the same file. Regression coverage checks GTK focus directly and real-input behavior in both directions; the previous commit fails the new regression.

PR #662, opened 2026-09-08
feat(browser): keep newly named items visible without unnecessary scrolling

### LINKED ISSUE #618: feat(browser): keep newly created items selected and visible after naming

ogarza:
## Description

Keep newly created files and folders selected, keyboard-focused, and fully visible when their naming editor opens and after the final name sorts into place in Columns, List, and Icons. Resolve the live GTK item by identity, account for the bottom hint bar, and leave the viewport unchanged when the final item is already visible—whether it moves upward or downward.

Wait for successful rename completion before revealing; cancel stale requests on failure, cancellation, supersession, or navigation. Stabilize editor/row heights to avoid layout jumps, while retaining local incremental updates rather than adding a full-view refresh. Existing-item rename reveal remains out of scope.

## Visual evidence

Before confirming the name: the newly created file's editor is fully above the destination hint.

Before naming

After confirming `000-renamed`: selected and focused at its final sorted position.

After naming

## How to test

1. Open a directory with enough entries to scroll. Create a file or folder whose default name lands offscreen; type immediately into its naming editor.
2. Confirm a name that sorts outside the viewport, then repeat with names that move upward and downward but remain inside it.
3. Repeat in Columns, List, and Icons, with descending sort and folders-first ordering. Also try Escape and a conflicting name.

**Expected result:** The naming editor and final item are fully visible, unobscured by the bottom hint. Successful naming preserves selection and keyboard focus. Already-visible destinations do not scroll; offscreen destinations move only enough to become fully visible. Failed or cancelled naming never selects an unrelated destination.

## Related issue

Closes #618


ogarza:
Fixed in #619

PR #697, opened 2026-09-09
fix(filter): open results independently of click preferences

### LINKED ISSUE #681: fix(filter): open results independently of navigation click preferences
### Strata version
User version not supplied; investigation targets main at 01d348be8701e593bbbc5fbb8e6dd9e96a131fa3.

### Installation method
Not supplied.

Reported from Linux; exact reporter environment not supplied.

### What happened?
Filtered/search results follow file navigation click preferences in all browsing modes.

### Steps to reproduce
1. Configure navigation to require double-click or preview on click.
2. Filter for files and folders in each browsing mode.
3. Single-click a result or press Enter once on the highlighted result.

### Expected behavior
One unmodified single click or one Enter opens the result: navigate into folders and launch files. Filtering activation must not depend on ordinary file/folder click or preview preferences. Preserve normal unfiltered navigation preferences and modifier selection behavior.

### Screenshots or video
Not supplied; interaction behavior report.

### Relevant logs
None supplied.

### Additional context
Related to #620, but distinct from result stability, preview, and context-menu handling. Cover all browsing modes and local/recursive filtering with regression tests.

l0gicgate:
## Description

Open local and recursive filter results with one unmodified click or Enter, regardless of normal file/folder click and preview preferences. Preserve unfiltered navigation and modifier selection behavior. Choosers retain the exact nested file path when accepting filtered results.

## Visual evidence

Before and after one click on a filtered folder with double-click preferences enabled (interaction captures, not a styling change):

Before activation
After one click

## How to test

1. Set file/folder activation to double-click and enable single-click previews.
2. In Columns, Icons, and List, filter for files and folders with recursive filtering both enabled and disabled.
3. Single-click a result; repeat by pressing Enter on a highlighted result.
4. Clear the filter and verify ordinary entries still follow click preferences.
5. Ctrl-click or Shift-click a filtered result and verify selection without opening. In a file chooser, accept a nested filtered file and verify its exact path is returned.

**Expected result:**

Filtered results open with one plain click or Enter in every browsing mode. Unfiltered preferences and modifier selection remain intact; chooser files are accepted only through Enter/Open.

## Related issue

Closes #681


wmfeht:
## Merge Blocker

Head `1582bb1`:

1. **Sticky `chooser_location` can accept the wrong file.** `accept()` prefers `browser.chooser_location()` over the visible selection. That side channel is set on unmodified click of a recursive search file and cleared only in `navigate` / `select`. Arrow keys in the search list, filter dismiss, and Ctrl/Shift clicks do not clear it — Open can return a path that is no longer selected (or only the remembered path under multi-select). Fix: drop the sticky field and read current selection/result on Open, or bind/clear it on filter dismiss, keyboard moves, and modifier selection. New chooser tests stub `select`/`set_chooser_location` and miss this.

2. **Recursive Icons/List file click can launch twice on double-click.** Single-click `GestureClick` opens the file and sets `suppress_row_activation`, cleared on `idle_add_local_once`. Idle often runs before the second click; ListBox then emits `row-activated` and launches again. Fix: claim the click, suppress `row-activated` for the whole gesture, or treat `press_count >= 2` as already consumed — do not clear suppress on idle.

l0gicgate:
Addressed both blockers in 4f4effe.

- Removed the sticky chooser path. Open reads live result selection, including all selected recursive Columns files. Inactive search and active-but-empty selection are distinct, so deselection/no matches cannot accept a hidden prior file. Existing focus-scoped preview behavior is preserved.
- Recursive Icons/List pointer gestures now claim the sequence directly; no idle-timed suppression remains. Keyboard activation stays separate, and a real pointer double-click regression verifies exactly one file launch.

Added widget-level chooser regressions for changing the selected result, filter dismissal, multiple selection, deselection, and no-match queries.

Validation: final Rust suite 1,180 passed (20 declared ignored), formatting and Clippy passed. Full canonical E2E passed 683 tests before the final chooser-selection refinements; the subsequent relevant preview/modifier/double-click selection passed 10 tests. Final chooser refinements were covered by the full Rust rerun.

PR #698, opened 2026-09-09
fix(ui): round before choosing file size unit

### LINKED ISSUE #657: fix(ui): 999,950 bytes is shown as "1000 kB" instead of "1 MB"
### Strata version

### Installation method

Built from source

### What happened?

Both `format_file_size` implementations (browser entry and preview) pick the unit before rounding, so a value that rounds to 1000.0 is printed in the smaller unit: `999_950` → "1000 kB", `999_950_000` → "1000 MB". The preview variant also prints `9_960` as "10.0 kB" while `10_000` prints "10 kB".

### Steps to reproduce

1. `head -c 999950 /dev/zero > /tmp/f`
2. Look at its size in the list and in the quick preview.

### Expected behavior

"1 MB" / "1.0 MB".

### Screenshots or video

### Relevant logs

_No response_

### Additional context

**Root cause:** Unit chosen from the unrounded value; the fix compares the rounded tenths and formats from them.

**Fix available:** branch [`fix/file-size-unit-rounding`](https://github.com/lgse/strata/compare/main...nixfred:strata:fix/file-size-unit-rounding) on my fork, with the regression test `file_sizes_use_compact_decimal_units (boundary cases) and formats_preview_file_sizes` that fails on current main and passes with the change. I have not opened a PR because the change was developed with an AI coding agent, which CONTRIBUTING.md excludes; please take the branch, or the analysis, in whatever form suits the project.

JoeJoeflyn:
## Summary
- Both `format_file_size` implementations picked the unit from the unrounded value, so a value that rounds up to 1000 stayed in the smaller unit: `999_950` → "1000 kB", `999_950_000` → "1000 MB"
- Extracted a shared `rounded_size_and_unit` helper that rounds to one decimal first, then promotes to the next unit if the rounded value crosses 1000
- Both the browser entry and preview formatters call the shared helper

Fixes #657.

#### Test plan
- [x] `cargo test --bin strata -- file_sizes formats_preview preview_file_sizes` — 4/4 pass
- [x] `cargo fmt --all --check` — clean
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean
- [ ] Manual: create a 999,950-byte file, verify it shows "1 MB" in list and "1.0 MB" in preview

https://github.com/user-attachments/assets/caa68ab2-64e0-4a7c-b341-9a6cb2b9f869

### DIFF
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -63,7 +63,7 @@ pub(super) use crate::ui::browser::context_menu::{
 pub(super) use crate::ui::browser::desktop::{launch_terminal, open_location};
 pub(super) use crate::ui::browser::entry::{
     FOLDER_TYPE_GROUP, entry_filter, entry_icon, entry_model_value, format_file_size,
-    metadata_needs_fill, model_type_group,
+    metadata_needs_fill, model_type_group, rounded_size_and_unit,
 };
 pub(super) use crate::ui::browser::inline_edit::{
     queue_rename, rename_stem_end, reveal_rename_row, update_basename_validation,
--- a/src/ui/browser/entry.rs
+++ b/src/ui/browser/entry.rs
@@ -12,18 +12,19 @@ use std::path::Path;
 use std::rc::Rc;
 
 pub(in crate::ui) fn format_file_size(bytes: u64) -> String {
-    let (value, unit) = rounded_size_and_unit(bytes, &["B", "kB", "MB", "GB", "TB"]);
+    const UNITS: [&str; 5] = ["B", "kB", "MB", "GB", "TB"];
+    let (value, unit) = rounded_size_and_unit(bytes, &UNITS);
     let formatted = format!("{value:.1}");
-    format!("{} {}", formatted.trim_end_matches(".0"), unit)
+    format!("{} {}", formatted.trim_end_matches(".0"), UNITS[unit])
 }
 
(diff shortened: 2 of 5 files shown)


wmfeht:
## Merge Blocker

Head `4e26d3c` (prior `bcce526` preview `5.0 B` / `1000 kB`@999450 blockers look fixed):

1. **Preview double-rounds with `f64::round()`.** `rounded_size_and_unit` already rounds to one decimal; preview then uses `value.round()` (half away from zero) on the integer path. That stacks tenths rounding and diverges from main’s `{value:.0}` (ties-to-even). Example: `10_499` / `10_500` → helper `10.5` → `"11 kB"` (main preview `"10 kB"`). Same window at other even `N.5` sizes. Format the displayed integer with `{value:.0}`, then promote if that result is 1000 — `.round()` is not needed for the 999450 → `1.0 MB` case.

2. **Preview can still print `"1000 GB"`.** Unit table still stops at GB; when the 0-decimal value is ≥ 1000 and `unit + 1` is out of range, the follow-up leaves `"1000 GB"`. Includes a regression at `999_450_000_000` (main `"999 GB"` → PR `"1000 GB"`) and still-broken `999_950_000_000` (list now `"1 TB"`, preview `"1000 GB"`). Give preview the same `TB` unit as the listing helper, or refuse to print 1000 when promotion is impossible. Add tests around `10.5 kB` and the GB/TB boundary.

PR #705, opened 2026-09-09
fix(columns): only slide_out source row when drag actually moved it

### LINKED ISSUE #598: fix(columns): source row slide-out animation runs even when a drag is cancelled
### Strata version

0.13.0 (source build at commit `200165606d007622d014e4119233ec96f0f0ce65`, `main`)

### Installation method

Built from source

Ubuntu 24.04.4 LTS, XFCE 4 / X11 (`xfwm4`), GTK 4.14.5

### What happened?

In Columns view, every file `DragSource` end runs `slide_out` on the source row, including cancelled drags and copies. The source row plays the 240 ms slide-up / fade as if it had been removed, then is still in the listing. During the drag a ghost of the source row also overlaps the next column.

Closed #337 asked to use `slide_in_down` after a **completed** drop instead of `slide_out`; it was closed not_planned. This report is narrower: `slide_out` should not run when the drag did not remove the source (cancel, or copy that leaves the original).

### Steps to reproduce

Isolated temp `HOME` / XDG dirs. Columns view, synthetic `files/{a.txt…h.txt, nested/}`.

1. Open Columns view on the fixture.
2. Drag `a.txt` by its icon/name (content hit) across into the next column, **without dropping on a folder** — cancel by releasing on empty space / the other column's non-destination area.
3. Watch the source `a.txt` row as the drag ends.

### Expected behavior

`slide_out` should run only when the drag actually removes the source (a successful move). A cancelled drag, and a copy that leaves the original, should not play the removal animation. GTK's `drag_end` `delete_data` flag is the usual signal for that.

### Screenshots or video

Live hold frame during a Columns content drag of `a.txt`: drag ghost of `a.txt` overlapping the `nested` column, with a faint duplicate `a.txt` / `17 B` in the source column. After cancel, the file list was unchanged (`dest/` still `[.keep]`, `a.txt` still in `files/`). The 240 ms post-cancel animation is easy to miss in stills; the `drag_end` handler is unconditional.

### Relevant logs

No application errors. After cancel:

```
RESULT 7 dest=['.keep'] files=['a.txt', 'b.txt', 'c.txt', 'd.txt', 'dest', 'e.txt', 'f.txt', 'g.txt', 'h.txt', 'nested']
```

### Additional context

Related: closed #337 (wanted `slide_in_down` after drop; not this cancel/copy case). Not a duplicate.

**Code observation (no patch in this report)**

`src/ui/browser/columns/rows.rs` `drag.connect_drag_end`:

```rust
drag.connect_drag_end(move |_, _, _| {
    if let Some(row) = dragged_row.upgrade() {
        row.remove_css_class("dragging");
        slide_out(&row);
    }
});
```

The third `drag_end` argument (`delete_data: bool`) is ignored. `slide_out` in `src/ui/modal.rs` is the removal animation.

No pull request; filing only.

JoeJoeflyn:
## Description
The Columns view `drag_end` handler ignored the `delete_data` argument and ran `slide_out` unconditionally, playing the removal animation on every drag end — including cancelled drags and copies that leave the original in place. Gate `slide_out` on `delete_data` so only actual moves animate.

## Visual evidence

screenrecording-2026-09-09_19-58-32

## How to test
1. Switch to Columns view.
2. Drag a file by its icon/name and cancel by releasing on empty space (not a folder).
3. Observe the source row.

**Expected result:** The source row stays in place without playing the slide-out/fade animation. Only actual moves (where the source is deleted) animate.

## Related issue
Fixes #598

### DIFF
--- a/src/ui/browser/columns/rows.rs
+++ b/src/ui/browser/columns/rows.rs
@@ -217,11 +217,13 @@ pub(super) fn column_rows(
                 }
             });
             let dragged_row = row.downgrade();
-            drag.connect_drag_end(move |source, _, _| {
+            drag.connect_drag_end(move |source, _, delete_data| {
                 source.set_actions(gtk::gdk::DragAction::COPY | gtk::gdk::DragAction::MOVE);
                 if let Some(row) = dragged_row.upgrade() {
                     row.remove_css_class("dragging");
-                    slide_out(&row);
+                    if delete_data {
+                        slide_out(&row);
+                    }
                 }
             });
             row.add_controller(drag.clone());


wmfeht:
## Merge Blocker

Gating Columns `slide_out` on GTK `delete_data` does not mean the source was removed — it means GTK accepted a `MOVE` drop. Empty-pane / same-directory no-op drops still return true (folder drops delay `start_transfer`; directory drop targets accept local MOVE), so the source can animate out and then snap back. Live check on head `8120c95`: empty-space drop of a file still plays slide-out while the file stays on disk.

Fix: stop calling `slide_out` from `drag_end` (match List/Icons); clear `dragging` on every end and let the listing drop the row when the transfer actually removes it. `delete_data` cannot prove removal.

PR #706, opened 2026-09-09
fix(operations): fail fast on special files instead of blocking a copy

### LINKED ISSUE #641: fix(operations): copying a folder that contains a named pipe hangs forever
### Strata version

### Installation method

Built from source

### What happened?

Copying (or cross-device moving) a directory that contains a FIFO never completes and cannot be cancelled. The copy worker opens every non-directory, non-symlink entry with a blocking read-only `open`, which for a FIFO waits for a writer that never comes. A character device such as `/dev/zero` in the tree would instead be copied without end.

### Steps to reproduce

1. `mkdir -p /tmp/src && mkfifo /tmp/src/pipe && echo hi > /tmp/src/a.txt`
2. In Strata, copy `/tmp/src` and paste it into another folder.
3. Observe the transfer never finishes; Cancel has no effect.

### Expected behavior

The copy fails with a clear error naming the entry that cannot be copied (as `compress` already does for special files), or skips it.

### Screenshots or video

### Relevant logs

_No response_

### Additional context

**Root cause:** `open_local_copy_source` in `src/adapters/local_operations.rs` treats every non-symlink, non-directory entry as a regular file and opens it without `O_NONBLOCK`, on a `spawn_blocking` thread. `open_archive_source` in `compression.rs` already guards against this.

**Fix available:** branch [`fix/copy-special-files`](https://github.com/lgse/strata/compare/main...nixfred:strata:fix/copy-special-files) on my fork, with the regression test `copying_a_tree_with_a_named_pipe_fails_instead_of_blocking` that fails on current main and passes with the change. I have not opened a PR because the change was developed with an AI coding agent, which CONTRIBUTING.md excludes; please take the branch, or the analysis, in whatever form suits the project.

JoeJoeflyn:
## Description

Copying a folder containing a FIFO (or other special file) hung forever: `open_local_copy_source` treated every non-symlink, non-directory entry as a regular file and opened it with a blocking `openat2`, which waits for a FIFO writer that never comes. Cancel could not interrupt the blocked worker.

The copy now only opens regular files (with `O_NONBLOCK` plus an `fstat` re-check against TOCTOU swaps) and fails fast with an error naming the entry for anything else, matching the existing `compress` behavior for special files.

## Visual evidence

https://github.com/user-attachments/assets/4f762094-f9cf-4c60-aa40-4020c7f84ccd

## How to test

1. `mkdir -p /tmp/src && mkfifo /tmp/src/pipe && echo hi > /tmp/src/a.txt`
2. In Strata, copy `/tmp/src` and paste it into another folder.
3. Observe the transfer fails promptly with an error naming `pipe` instead of hanging; Cancel is no longer needed.

**Expected result:** the copy fails fast naming the special entry; no hang, no unkillable progress dialog.

## Related issue

Closes #641


JoeJoeflyn:
Superseded: branch is now pushed directly to origin as fix/641-copy-special-files (collaborator, no fork needed). Will reopen from there.

PR #713, opened 2026-09-09
fix(sidebar): keep pinned places when the GTK bookmarks file has bad UTF-8

### LINKED ISSUE #648: fix(sidebar): one non-UTF-8 byte in the GTK bookmarks file drops every pinned place
### Strata version

### Installation method

Built from source

### What happened?

`load_pinned_places` reads the bookmarks file with `read_to_string`, so a single invalid byte in any label (GTK itself never validates the file) makes the whole file parse as empty and the sidebar shows no pinned places. The next pin then writes a one-line file over all the others.

### Steps to reproduce

1. `printf 'file:///tmp/a A\nfile:///tmp/b \xff\n' > ~/.config/gtk-3.0/bookmarks`
2. Start Strata: no pinned places are shown.
3. Pin any folder: the file now contains only that folder.

### Expected behavior

The bookmarks with valid URIs are shown; only the label is decoded leniently.

### Screenshots or video

### Relevant logs

_No response_

### Additional context

**Root cause:** Byte-safe parsing (split on `\n`, tolerate CRLF, `from_utf8_lossy` on labels, skip only lines whose URI is invalid UTF-8) fixes it.

**Fix available:** branch [`fix/bookmarks-byte-safe`](https://github.com/lgse/strata/compare/main...nixfred:strata:fix/bookmarks-byte-safe) on my fork, with the regression test `gtk_bookmarks_survive_non_utf8_labels_and_windows_line_endings` that fails on current main and passes with the change. I have not opened a PR because the change was developed with an AI coding agent, which CONTRIBUTING.md excludes; please take the branch, or the analysis, in whatever form suits the project.

JoeJoeflyn:
## Description

`load_pinned_places` read the GTK bookmarks file with `read_to_string`, so a single non-UTF-8 byte in any label (GTK itself never validates the file) made the whole file parse as empty and the sidebar showed no pinned places. The next pin then wrote a one-line file over all the others.

The file is now read as bytes (`std::fs::read`) and `parse_pinned_places` splits on `\n`, tolerates CRLF, decodes labels with `from_utf8_lossy`, and skips only lines whose URI is not valid UTF-8. Valid URIs survive a bad byte anywhere else in the file.

## Visual evidence
Before:
image
After:
image

## How to test

1. `printf 'file:///tmp/a A\nfile:///tmp/b \xff\n' > ~/.config/gtk-3.0/bookmarks`
2. Start Strata.

**Expected result:** both `/tmp/a` (label "A") and `/tmp/b` (label U+FFFD) appear in the sidebar; no pinned places are lost to the single invalid byte.

3. Pin any folder.

**Expected result:** the bookmarks file gains the new entry alongside the existing two, instead of being overwritten with only the new one.

## Related issue

Closes #648


l0gicgate:
Merged upstream main at f8ab4fedb10fb82ce3e8bb892678ddbba03eb6c0 and reviewed the complete updated diff. Added a two-window regression covering real bookmark loading, pinning, and saving with an invalid UTF-8 label, so reverting to read_to_string cannot evade parser-only coverage. Tightened the sole new comment to its compatibility rationale.

Pinned local formatting, Clippy, and full Rust suite passed (1,328 passed; 20 ignored), plus the canonical bookmark pin/unpin E2E scenario. Full local E2E was omitted for this bounded change. Host Xvfb was unavailable; GTK validation passed using private Xvfb/D-Bus in the pinned container. No manual desktop interaction or new media: review edits affect only tests/comments.

PR #748, opened 2026-09-10
fix(installer): read interactive pacman prompts from the terminal

### LINKED ISSUE #682: website installer
pacman is reading from the installer pipe instead of the terminal, the upstream fix is to change the interactive pacman invocation to:
`sudo pacman -S --needed -- "$@" </dev/tty`

separately, it unnecessarily requests `github-cli` because it doesn't recognize the mise-installed `gh`.

spandan11106:
## Description

When `install.sh` is piped through `curl | bash`, stdin is bound to the installer pipe rather than the terminal. The interactive `pacman -S --needed --` invocation inherited that stdin, so pacman's confirmation prompt read from the pipe instead of the terminal, instead of the `/dev/tty` the rest of the installer's own prompts already use.

Separately, `github-cli` was always checked with `pacman -Q`, so a `gh` installed another way (e.g. via mise) was not recognized and the installer offered to install `github-cli` through pacman unnecessarily.

## Changes

- `run_pacman`'s interactive branch now redirects from `/dev/tty` (the installer already verifies `/dev/tty` is readable/writable before entering interactive mode, so this is always available on that path).
- `install_arch_dependencies` skips the `github-cli` package requirement when `gh` is already resolvable on `PATH`.
- Added focused tests in `scripts/test_installer.py` covering both cases.

## Visual evidence

N/A — shell script fix, no UI change.

## How to test

1. `python3 -m unittest scripts.test_installer -v`
2. Manually: `curl -fsSL <install.sh url> | bash` on an Arch-based system with an existing pacman package pending confirmation, and with `gh` installed via a non-pacman method (e.g. mise).

**Expected result:** pacman's confirmation prompt reads from the terminal instead of hanging/failing on the installer pipe, and `github-cli` is not listed as a required package when `gh` is already available.

## Related issue

Closes #682

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -192,13 +192,14 @@ run_pacman() {
     sudo -n pacman -S --needed --noconfirm -- "$@" \
       || die "Non-interactive package installation failed; passwordless sudo or cached credentials may be required."
   else
-    sudo pacman -S --needed -- "$@"
+    sudo pacman -S --needed -- "$@" </dev/tty
   fi
 }
 
 install_arch_dependencies() {
   local missing=() package
   for package in "${REQUIRED_PACKAGES[@]}"; do
+    [[ $package == github-cli ]] && command -v gh >/dev/null 2>&1 && continue
     pacman -Q "$package" >/dev/null 2>&1 || missing+=("$package")
   done
 
--- a/scripts/test_installer.py
+++ b/scripts/test_installer.py
@@ -2,6 +2,7 @@
 
 import os
 import pathlib
+import shutil
 import subprocess
 import tempfile
 import unittest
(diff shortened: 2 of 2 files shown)


wmfeht:

Agent: https://cursor.com/agents/bc-d0af6395-ab2d-4df1-94f2-d03037027860  
Head: `ed49013d5db929b21504b89187b5bddf4ad6c2c2` (pinned, clean)

**Result: clean pass — no confirmed candidate issues.**

### Context
Interactive `run_pacman` now uses `</dev/tty` (curl|bash stdin fix for #682); skips `github-cli` when `gh` is on PATH. Non-interactive `--noconfirm` path unchanged.

### Checked
- Pipe+PTY harness: new path reads TTY answers; pre-fix body consumed piped stdin
- Piped script after pacman still reaches marker / script tail
- mise-style `gh` shim on PATH drops only `github-cli` from missing list; no-`gh` PATH still offers it
- `--noconfirm` argv/guard unchanged vs base
- `python3 -m unittest scripts.test_installer -v` → 21 passed

Probes used sudo/pacman shims, not a destructive Arch install.

l0gicgate (approved):

PR #749, opened 2026-09-10
fix(keyboard): scope Ctrl+A to the entry while renaming

### LINKED ISSUE #687: fix(keyboard): Ctrl+A in the rename field also selects every file
### Strata version

v0.14.0-rc.1 (commit `01d348b`)

### Installation method

Built from source

### What happened?

With the caret in an inline rename field, Ctrl+A selects the rename text **and** every row/tile in the pane. After Escape or commit, that file selection can remain, so a later Delete would trash the whole listing.

While renaming, `Dispatcher::inline_editing` returns `Propagation::Proceed` for keys other than F2/Escape (`src/ui/window/keyboard/commands.rs`). The window controller is in the capture phase, so Proceed lets GTK’s `list.select-all` on ListView/GridView run as well as the Entry binding. Filter already special-cases this (`filter_has_focus()` → Proceed **without** calling `view.select_all()`); rename does not.

This is **not** a duplicate of #650, which is Ctrl+C/V/X/A/D/Delete inside **preview** text.

GTK test `ctrl_a_during_rename_proceeds_to_gtk`: F2 then Ctrl+A returns Proceed (does not Stop).

### Steps to reproduce

1. Open a folder with several files.
2. Select one file and press F2.
3. Press Ctrl+A.
4. Observe the rename text and the pane selection.
5. Press Escape, then consider Delete.

### Expected behavior

Ctrl+A inside the rename field should select only the editor text. Pane selection must stay on the item being renamed. Delete after cancelling rename must not offer to trash every file.

### Screenshots or video

N/A — dispatcher behaviour confirmed by GTK test; ListView/GridView then apply their default select-all binding.

### Relevant logs

```
GTK test ctrl_a_during_rename_proceeds_to_gtk: ok
inline_editing → Propagation::Proceed while rename_is_active()
clipboard_command only skips Strata select-all when filter_has_focus()
```

### Additional context

Verified on tag **v0.14.0-rc.1**. Found in a Fable 5.1 exploratory pass (agent `bc-1f7e335f-ac08-49b6-b77f-2b028e19f117`). Do not close as a duplicate of #650.

pdwaldrop:
## Description

`Dispatcher::inline_editing`'s catch-all returned `Propagation::Proceed` for every key except F2/Escape while a rename (or new-entry naming) was active, so it never got a chance to apply `filter_has_focus()`-style scoping the way `clipboard_command` already does for the filter field.

For Ctrl+A specifically, that `Proceed` let the event continue past this window-level capture controller to the surrounding ListView/GridView's own default select-all binding on its way to the entry -- selecting every row in the pane along with the entry's text. That pane-wide selection can survive canceling the rename, so a later Delete would offer to trash the whole listing.

Ctrl+A is now handled directly while inline editing is active: select the active rename field's text and stop, instead of proceeding and letting GTK own it. Every other key while renaming is unaffected -- confirmed against the existing `inline_editing_owns_filter_keys_but_not_global_search` test, which already covers Ctrl+F and Ctrl+K during rename.

## Visual evidence

Fixed behavior, Columns mode: after F2 then Ctrl+A, only the rename field's text is selected -- `file1.txt`-`file4.txt` are not highlighted.

Ctrl+A during rename selects only the entry text

I also tried to reproduce the original bug live -- pre-fix and fixed builds, Explorer and Columns mode -- via F2 -> Ctrl+A -> Escape -> Delete, checking the delete-confirmation dialog's item count. On this machine (current GTK, Hyprland/Wayland) the pre-fix build's dialog only ever said "Permanently delete 1 item?" in either view mode; the whole-pane selection described in the issue didn't manifest visually here, likely a GTK-version or X11-vs-Wayland difference from the original report's environment (Ubuntu 24.04, XFCE, GTK 4.14.5, X11). That doesn't change the dispatcher-level root cause: `ctrl_a_during_rename_selects_only_the_entry_text` presses F2 then Ctrl+A through the actual window-level dispatcher (bypassing GTK's own event delivery, so it isn't sensitive to that environment difference) and asserts the rename entry's selection bounds cover its full text while the pane's file selection stays on just the renamed item. Reverting only the `commands.rs` change makes it fail exactly as described (`press` returns `Proceed` instead of `Stop`), confirming the test isolates this fix specifically.

## How to test

1. Open a folder with several files.
2. Select one file and press F2.
3. Press Ctrl+A.
4. Observe the rename text and the pane selection.
5. Press Escape, then consider Delete.

**Expected result:** Ctrl+A selects only the text in the rename field. The pane selection stays on the item being renamed. After Escape, Delete does not offer to trash every file.

## Related issue

Closes #687


l0gicgate:
Reviewed and merged upstream main at afef6c701cf5e649f3cbaad0431621539d10d1af (merge commit 9adfcde). Follow-up d19ce3c tightens the capture-phase rationale, removes redundant test documentation, and expands Ctrl+A coverage to Unicode names across view modes and new-entry naming.

Validation is incomplete: the expanded Icons test initially failed; after adding a rendered-row wait, the targeted rerun hung and was stopped. Native headless testing could not start because Xvfb was unavailable. Formatting, Clippy, and canonical E2E were not run. No new manual UI verification or media was produced; these follow-up changes affect comments/tests, not UI implementation.

Pushed at the owner’s explicit direction despite these local validation gaps. This is not a merge-ready approval: the expanded test remains unverified and required CI checks must pass.

PR #755, opened 2026-09-10
fix(startup): classify open arguments asynchronously

### LINKED ISSUE #726: fix(startup): classify local and network arguments asynchronously
### Strata version

v0.14.0; code review of PR #725 at `72c1ba0` (follow-up to #722).

### Installation method

Built from source

Review host: Omarchy Linux. Existing isolated GUI validation uses Ubuntu 24.04 / GTK 4.14.5 under Xvfb. Live network mounts and authentication were not exercised; the blocking risk below is identified from code, not a reproduced outage.

### What happened?

Startup argument classification in `src/main.rs::open_requests` has two related limitations:

- Native paths are classified using synchronous filesystem metadata on the GTK main thread. NFS/CIFS mounts (and explicitly supplied FUSE paths) are native paths too, so a slow or unavailable server can stall startup or an already-running application's open handler.
- Non-native SMB/SFTP file URIs are deliberately not probed synchronously. They continue through directory navigation rather than revealing the file in its parent. A GVfs FUSE mirror must not be mistaken for a genuinely native URI.

PR #725 intentionally fixes local file arguments only. Resolve both limitations with asynchronous argument classification rather than expanding synchronous probing.

### Steps to reproduce

Suggested network reproduction (not run on the review host):

1. In an isolated test environment, mount an NFS/CIFS share and make the test server slow or unavailable.
2. Pass a path on that mount to Strata, both on cold launch and while a window is already open; observe whether the UI remains responsive during metadata lookup.
3. With an available SMB/SFTP share, pass a file URI such as `sftp://test-host/share/file.txt` to Strata.
4. Check whether Strata opens the parent and selects the file instead of attempting to enumerate it as a directory.

### Expected behavior

Argument metadata lookup must not block the GTK main thread. Local/network directories open normally; files are revealed in their parent, preserving the original URI/path and argument ordering. Remote queries should handle mounting/authentication, cancellation, and failure without silently treating every unknown target as a file or leaking a GVfs mirror path.

Recommended feedback: delayed and non-blocking, rather than silence or an immediate dialog.

- Open the requested window immediately and classify the argument asynchronously, keeping the UI responsive.
- If still waiting after approximately one second, show a spinner and “Connecting to location…” with **Cancel**.
- On success, open the directory or reveal the file in its parent.
- On failure, show an inline error with **Retry**, not a modal dialog.
- If the operation remains pending, keep cancellation available. Elapsed time alone should not cause failure: network mounts and authentication can legitimately take time.
- Use GIO's existing mount/authentication prompts where needed.

For NFS/CIFS, cancellation may not immediately terminate a blocked filesystem call. **Cancel must stop the UI from waiting and ignore late results**, without assuming the underlying operation has terminated.

**New navigation wins:** If the user navigates elsewhere while an open request is pending, cancel/invalidate that request immediately and dismiss its loading feedback. Ignore late successes and failures: no unexpected navigation, selection changes, stale errors, or delayed loading indicators. Closing the window must also invalidate its pending request. Scope cancellation to that window so other windows' requests continue independently.

Use a per-window navigation generation/request ID (prefer existing navigation infrastructure) and apply callbacks only while the request is still current and the window exists. Cancellation alone is insufficient when a blocked filesystem operation can complete later.

Keep the first implementation minimal: no progress percentages, configurable timeouts, or automatic retries.

### Screenshots or video

N/A — follow-up identified through code review; no network reproduction captured.

### Relevant logs

None collected for live network mounts.

### Additional context

Related: #722 and #725. Keep the existing native-location distinction in `src/adapters/gio_location.rs`; `gio::File::path()` alone does not establish that a location is native. Local symlink handling stays in #725; async local/network classification belongs here.

spandan11106:
## Description

`open_requests` (`src/main.rs`) classified every launch/open-request argument with a synchronous `std::fs::metadata` call on the GTK main thread, and only ever probed native paths — a native path that's actually a slow or dead NFS/CIFS mount could stall startup or an already-running instance's open handler, and a non-native file URI (`sftp://`, `smb://`) was never revealed in its parent at all.

This replaces that with fully async classification, using the same `query_info_future` primitive already used elsewhere for non-native location validation:

- The window for a launch argument now presents immediately (`ui::present_open`), before its file-vs-directory kind is known.
- Classification runs off the main thread as a normal async query. On resolution it either opens the location directly or redirects to reveal the target in its parent.
- Cancellation safety reuses the existing per-window navigation-generation counter (no second mechanism): navigating elsewhere or closing the window invalidates a pending classification, and a stale result is simply ignored rather than assumed to have terminated.
- If classification is still pending after ~1s, an inline (non-modal) status banner shows a spinner and "Connecting to location…" with Cancel. On failure, an inline error with Retry replaces it.
- The D-Bus `org.freedesktop.FileManager1` handler (`ShowItems`/`ShowFolders`) is untouched — it already knows file-vs-directory from the method name and keeps using `present_reveal` directly.

## Visual evidence

Missing launch argument before (the path is opened as a directory):

Before: missing argument opened as a directory

After (centered plain error with the requested location and Retry):

After: centered unavailable-location feedback with Retry

## How to test

1. Build the app, then launch it with a plain directory argument (`strata /some/directory`) — window opens and navigates directly, no banner.
2. Launch it with a plain file argument (`strata /some/file.txt`) — window opens immediately and reveals the file selected in its parent directory.
3. Launch it with a path to a broken symlink — still reveals the link itself in its parent, matching prior behavior.
4. Launch it against a slow/unavailable network mount (e.g. an unresponsive NFS/CIFS share) — the window appears immediately and stays responsive; after about a second, "Connecting to location…" with Cancel appears; Cancel returns to the original location without hanging the UI.
5. Trigger a classification failure (e.g. a path that disappears mid-flight, or a permission error) — centered plain error feedback with the requested location and Retry appears instead of a modal dialog; Retry re-runs classification.

**Expected result:** the GTK main thread never blocks on argument classification; directory arguments open normally; file arguments (native or remote) are revealed in their parent; a slow/dead network target shows delayed, cancellable feedback instead of freezing or silently misclassifying.

## Related issue

Closes #726


l0gicgate (dismissed):

l0gicgate:
Review follow-up completed:

- Unmounted remote arguments now use GIO mount/authentication before metadata classification retries.
- Navigation and window close invalidate pending results, cancel delayed status timers, abort mount prompts, and release window references.
- Fast failures retain the inline **Retry** state instead of being overwritten by the delayed connecting spinner.
- Error text no longer exposes credentials embedded in launch URIs.
- Rust regressions cover fast-failure timing, cancellation/navigation, window teardown, mount abort replies, local files/directories, and symlinks; the full canonical E2E suite also passed.

Manual missing-path evidence: [before](https://raw.githubusercontent.com/spandan11106/strata/933f3ed4b0785e2d69c7f8a9e8931d99251328f9/docs/screenshots/726/before.png) / [after](https://raw.githubusercontent.com/spandan11106/strata/6c8abb4b972a34989f395d350d338a9a2a7d2891/docs/screenshots/726/after-error.png).

No live SMB/SFTP server validation was available; mount/authentication behavior was reviewed and exercised through lifecycle regressions, not a live remote endpoint.

l0gicgate:
CI follow-up: merged current `lgse/strata@9c4ccd86fa150370479c87a743e8cb7f8e572638`. The `src/ui/window/tests/keyboard_dispatch.rs` conflict was resolved to preserve main’s `GtkInscription` rendering fix; the open-argument implementation remained intact. Final local pinned-container results: 1,359 Rust tests passed (20 ignored), 747 canonical E2E tests passed, and fmt/Clippy passed.

l0gicgate:
UI correction follow-up:

After: centered unavailable-location feedback

The unavailable-location state is now centered in the browser content area with plain themed text, the safely displayed requested path/URI, and a normal **Retry** button below—without the danger badge or card container. Retry still re-runs metadata classification and mounting. Regression coverage restores initially missing directory and file targets and verifies Retry opens/reveals them successfully; the image above is the fresh canonical manual capture.

PR #760, opened 2026-09-10
fix(columns): preserve new-entry selection, focus, and child paths

### LINKED ISSUE #690: fix(columns): New Folder from the parent background leaves the child column stale
### Strata version

v0.14.0-rc.1 (commit `01d348b`)

### Installation method

Built from source

### What happened?

In Columns, with a child column already showing some folder, creating a **New Folder from the parent column background** highlights the new folder in the parent, but the **child column still shows the previously opened folder** and the **breadcrumb stays on that old path**. Keyboard focus and the Miller “opened” folder are out of sync with the highlighted row.

`rename_created_entry` waits for the new row, then calls `browser.select(pending.depth, position)` and starts the rename editor. `select()` only updates that column’s selection and emits `FocusChanged`. It does not `descend_with_selection` / `activate_focused`, so the next column and breadcrumbs are not rebuilt. `sync_active_location` is not run from that `FocusChanged` handler.

### Steps to reproduce

1. In Columns, open a directory that contains at least one folder.
2. Select that folder so a child column lists its contents.
3. Right-click empty space in the **parent** column → New Folder (or Ctrl+Shift+N while that column is focused, if it still targets the parent).
4. Observe the new folder highlighted in the parent, the child column, and the breadcrumb.

### Expected behavior

Selecting the brand-new folder should open it as the child column (or clear the previous child) and update the breadcrumb to match the highlighted folder, consistent with clicking that row.

### Screenshots or video

N/A — behaviour follows `ViewState::rename_created_entry` (`src/ui/browser/inline_edit.rs`) calling `Browser::select` rather than descending.

### Relevant logs

```
rename_created_entry → browser.select(depth, position)
NavigationState::select updates selected_locations only; no ColumnAdded / restore_path
FocusChanged handler updates row highlight, not breadcrumbs
```

### Additional context

Verified on tag **v0.14.0-rc.1**. Found in a Fable 5.1 exploratory pass (agent `bc-1f7e335f-ac08-49b6-b77f-2b028e19f117`).

pdwaldrop:
## Description

Fix the create-and-rename lifecycle in Columns mode so the selected item, keyboard focus, child columns, and breadcrumb stay consistent. This builds on the original stale-child-column fix and the follow-up preventing New File from opening an external application.

- Reveal an already-created entry without starting a second asynchronous navigation that could cancel its rename editor. New folders replace stale descendants; new files close them and never activate a file handler.
- On a successful folder rename, relocate only the affected open columns rather than resetting the entire path. Preserve parent selection, outside-control focus, and column preferences; handle URI paths and completion without a watcher notification.
- Ignore detached virtualized row bindings and recover GTK's removed-row/pane-stack keyboard-focus fallback, so Enter followed by F2 continues editing the correct item.

The guided walkthrough on a local optimized release candidate confirmed creation/Enter/F2, click-away selection, folder conflicts, Escape/empty/invalid names, Home/Back navigation, and unchanged List/Icons behavior. Everything exercised now appears to work correctly. Live SMB/SFTP shares were not manually exercised; remote lifecycle behavior has controlled-provider coverage.

## Visual evidence

The contributor's before/after captures demonstrate the original stale-child-column fix and the new-file editor regression. The follow-up preserves these outcomes and adds the selection/focus behavior described above.

Before: the child still shows the previously opened `existing` folder.

Before: stale child column

After: the child follows the newly created folder.

After: child column follows the new folder

Before: New File leaves a stale child and never opens its rename editor.

Before: new file rename editor missing

After: the stale child closes and the new-file rename editor appears.

After: new file rename editor appears

## How to test

Use a build containing this PR and a disposable directory containing an `existing` folder with some contents and a `taken.txt` file.

1. In Columns mode, open `existing`, then right-click the **parent column's background → New Folder**. The default folder must exist immediately, its whole name must be selected in the parent editor, and its child/breadcrumb must replace the old path.
2. Type `alpha`, press **Enter**, then **F2 without clicking elsewhere**. The child/header/breadcrumb should follow `alpha`; the same parent item must remain selected and editable. Escape closes the editor without deleting it.
3. From the parent background choose **New File**. Stale descendants should close and the breadcrumb should return to the parent, with no application launch or opening error. Rename it to `beta.txt`, Enter, then F2; it should remain the edit target.
4. Create another folder, type `gamma`, and click `taken.txt` instead of Enter. The folder rename should complete, but `taken.txt` must remain selected and be the F2 target. Repeat using **Home**, then **Back**: no late editor or unwanted navigation, and the returned path should be valid.
5. Try a conflicting folder name (`existing`). Dismiss the error and verify both the pre-existing contents and the newly created default folder remain intact, with consistent child/header/breadcrumb labels. Also check a file conflict against `taken.txt`.
6. During a new folder's rename, try **Escape**, an empty name, and `bad/name`. Each should finish/cancel editing while retaining the default item and its actual path. An unchanged name should simply finish editing.
7. Repeat creation without an existing child and in a narrow window. In **List and Icons/Grid**, create and rename both files and folders: stay in the parent directory, retain selection for F2, and do not automatically navigate or open a file.

**Expected result:** creation, rename completion, cancellation, and errors keep the UI consistent with the filesystem without losing the parent selection, stealing focus, reopening abandoned paths, or launching new files.

## Related issue

Closes #690


ogarza:
Thanks for the fix! Updating the child column makes sense, but reviewing and trying it exposed a few interactions we need to address before merging. This touches more of the create-and-rename lifecycle than it initially appears. 

While testing I found that the child column was updated correctly, but the folder was no longer highlighted and trying to create a new file following the same instructions was showing an error message in the UI.

If you notice on the main release, when you create a folder or a new file, and confirm it by pressing enter, the file or folder is highlighted, this is working as designed. The highlight once it is open is different than when you have not activated the child column and on some themes it can be hard to spot, but it should be there.

Where I agree with you is that the UI is out of sync, and for a folder, we need to update the child column's location and breadcrumb if the rename was succesful.

 The expected behavior is:

 ### Creation

 - Create the file/folder immediately on disk with an available default name.
 - Select and reveal it in the parent column, then start inline rename with the entire default name selected.
 - New folder: replace the previous child column with the new folder’s contents and update the breadcrumb, while keeping focus in the parent’s rename field.
 - New file: close stale descendant columns and keep the breadcrumb at the parent. Do not open the file or launch an application. Currently, this produces an opening error.

 ### Completing rename

 - Enter: submit once and close the editor, keeping the item selected and visible. Enter must not also activate it.
 - For folders, update the child column’s location/header and breadcrumb after a successful rename.
 - An unchanged name simply finishes editing.
 - Click outside: submit a valid changed name, then allow the clicked target’s normal action. If the user selects something else or navigates away, rename completion must not steal focus, restore the previous selection, or reopen columns.

 ### Cancellation and failures

 - Escape: cancel only the rename; retain the already-created item under its default name.
 - Empty/invalid names should preserve the existing behavior: finish editing and retain the original name.
 - Rename conflicts or failures must retain the item, report the error, and leave labels, columns, and breadcrumbs consistent with its actual location.
 - Creation failures should report an error without opening an editor or navigating into a nonexistent item.
 - Navigation or cancellation before creation finishes must prevent a late editor from appearing or pulling the user back.

 Please cover files and folders, with and without an existing child column, including Enter, outside-click, Escape, failures, and navigation during pending operations. Tests should verify actual selection and keyboard focus, not just editor presence, and that New File never requests opening. List/Grid behavior should remain unchanged.

 You’re welcome to continue with these adjustments. Equally, if you’d prefer to hand off the follow-up, someone from the team can pick it up, let me know. Thanks again for identifying the issue and getting the fix started!

ogarza (dismissed):
Please see my previous comment

pdwaldrop:
@ogarza Thanks for the thorough review — you were right, this needed more than the child-column fix alone. Pushed an update (rebased onto current main, which had drifted a fair bit):

- **Fixed the regression you found**: `activate()` was firing for any new entry, not just folders, so creating a new file tried to open it and surfaced an error. It's now gated on `is_directory` — new files select and close any stale child column instead.
- **Child-column-follows-a-rename**: turns out this is already handled generically by `path_after_external_change` (rebases any open path under a renamed directory) — no extra code needed, just added a test proving it works for the just-created-then-renamed case specifically.
- **Broader lifecycle coverage**: added tests for Enter vs. outside-click completion (both keep the right item selected without reactivating it), an unchanged name on outside-click (just closes the editor, no rename dispatched), and a rename conflict (retains the item under its original name).

Full details and updated before/after shots (including one for the new-file regression) are in the PR description. Let me know if you'd still like this split up differently or want to take it over from here.

ogarza:
@pdwaldrop Thanks again for finding #690 and getting the initial fix and new-file regression coverage in place. We followed up on your commits and pushed **780841f** to this PR.

The remaining issues were in the surrounding lifecycle rather than the basic child-column update:

- The generic path restore updated the renamed folder's path, but reset parent selection and moved keyboard focus. We now relocate only the affected open columns, preserving selection, focus, and column preferences.
- Activating a newly created remote folder could finish validation after its editor opened and cancel the editor. Created entries now have a dedicated reveal path, without a second asynchronous navigation or file activation.
- URI rebasing and successful rename publication now cover open remote paths without depending on a watcher notification.
- Native GTK testing also exposed detached row bindings and pane-stack focus fallback. The follow-up resolves live rows and restores the correct keyboard cursor without taking focus from outside controls.

Added state/widget and real keyboard/pointer regressions, then walked through the optimized release candidate: Enter → F2, New File with stale children, click-away selection, folder conflicts, Escape/empty/invalid names, Home/Back, and List/Icons behavior were all confirmed. **Everything we exercised now seems to be working correctly.** The final affected GUI selection passed 227 cases, and the exact release binary passed its 13 create/rename scenarios. Live SMB/SFTP shares were not manually exercised; those paths have controlled-provider coverage.

I've updated the PR description and manual test steps to reflect the final implementation and results, retaining your original visual evidence. The PR remains in draft for final review. Thanks for the original contribution and for working through the earlier feedback with us!

ogarza (approved):

PR #761, opened 2026-09-10
fix(dnd): preserve filesystem identity through network aliases

### LINKED ISSUE #734: fix(dnd): preserve filesystem identity through network mount aliases
### Strata version

0.14.0 source build at `aacda2b5c8cdf8e526bd237fde0ff39d394cc506`; affected code verified unchanged in merged main `f8ab4fedb10fb82ce3e8bb892678ddbba03eb6c0` (#502).

### Installation method

Built from source

Omarchy / Arch Linux. Verification used private Xvfb/X11 and D-Bus with disposable HOME/XDG, real temporary directories/symlinks, and synthetic NFS mountinfo. No real NAS or user files were modified.

### What happened?

Two paths on the same native network/FUSE filesystem can be classified as different volumes when one is reached through a symlink outside the mount's lexical path.

`Directory::classify` captures `is_remote` from the lexical path. GIO then follows the symlink and obtains the correct filesystem ID, but `VolumeIdentity::matches` also requires the original `is_remote` booleans to match.

Example: `/home/me/nas -> /mnt/nas`, where `/mnt/nas` is NFS. Dropping `/mnt/nas/a/file` into `/home/me/nas/b` can produce matching GIO filesystem IDs but a `Different` relation. Always Ask then incorrectly claims a device difference; Always Copy copies instead of following the expected same-volume move policy. No data loss was demonstrated.

### Steps to reproduce

Deterministic regression using the existing volume-test helpers:

1. Create `<fixture>/nas/a`, `<fixture>/nas/b`, and `nas/a/file` in a temporary directory.
2. Create `<fixture>/alias` as a symlink to `<fixture>/nas`.
3. Build synthetic mountinfo describing `/` as an ordinary local filesystem and `<fixture>/nas` as `nfs`.
4. Use `DropVolumeQuery::with_mounts` for source `nas/a/file` and destination `alias/b`; resolve with `resolve_with_mounts` on a private main context.
5. Assert matching returned filesystem IDs and `VolumeRelation::Same`.

The IDs match, but the relation is `Different`. This exercises the production path/identity logic without needing a live NFS server. A real NAS/FUSE alias smoke test remains useful separately.

### Expected behavior

Equivalent native filesystem identities should remain comparable through directory aliases. Once lookup resolves, same-volume plain drops should follow the same move policy regardless of which alias was used.

Separate **lookup blocking risk** from **stable filesystem/backend identity**. Preserve genuine backend identity distinctions; do not blindly remove all namespace distinctions or assume unresolved lookups are same-volume.

### Screenshots or video

N/A — reproduced at the volume-classification layer; no live-NAS dialog recording was captured.

### Relevant logs

Sanitized review-probe result; the particular ID string is incidental:

```text
DropVolumeLookup {
  dest: Some(VolumeIdentity { filesystem_id: "l56", is_remote: false }),
  sources: [Some(VolumeIdentity { filesystem_id: "l56", is_remote: true })],
  relation: Different
}
assertion failed: actual Different, expected Same
```

### Additional context

Follow-up to merged #502 and the behavior introduced for #248. Keep this focused on identity comparability, with no changes to the default strategy, dialogs, or restore confinement.

Relevant code at the merge:
- [Lexical classification](https://github.com/lgse/strata/blob/f8ab4fedb10fb82ce3e8bb892678ddbba03eb6c0/src/adapters/volume.rs#L147-L159)
- [Captured remote flag passed into the async result](https://github.com/lgse/strata/blob/f8ab4fedb10fb82ce3e8bb892678ddbba03eb6c0/src/adapters/volume.rs#L218-L230)
- [Identity comparison](https://github.com/lgse/strata/blob/f8ab4fedb10fb82ce3e8bb892678ddbba03eb6c0/src/services/transfer_action.rs#L10-L22)

Acceptance:
- [ ] Add the alias regression and verify `Same` after the asynchronous lookup completes, including the reversed source/destination spelling.
- [ ] Verify the resulting plain-drop policy is Move even when the configured cross-device strategy is Copy or Ask.
- [ ] Preserve genuinely different/unknown results, backend distinctions, modifier overrides, and asynchronous handling of potentially blocking paths.

ogarza:
## Description

Separate lookup blocking risk from GIO filesystem identity. Native paths on network mounts and their symlink aliases now share the `file` backend namespace, while other URI backends remain distinct. Add regressions for both alias directions, Copy/Ask policies, and modifier overrides.

Side note: fix pre-existing keyboard-test formatting and teach its filename helper to recognize `GtkInscription`, preventing the icon-view fixture timeout. No keyboard behavior changes.

An unrelated updater ownership-probe test failed once during validation, then passed in isolation and in the final full run; its intermittent cause remains unverified.

## Visual evidence

N/A — filesystem identity classification and test-helper changes; no interface layout changes. A live-NAS smoke test has not been performed.

## How to test

1. With an existing NFS/FUSE mount, create disposable `a` and `b` directories and a symlink alias to the mount outside its lexical path. Put a test file in `a`.
2. Set the cross-device strategy to Always Copy, then drag the file from the direct `a` path to `b` through the alias without modifiers. Repeat with Always Ask and with the direct/alias spellings reversed.
3. Repeat with Ctrl held to request a copy.

**Expected result:** Plain drops move the file without a device-difference prompt in either direction. Ctrl-drops copy and retain the source.

## Related issue

Closes #734

### DIFF
--- a/src/adapters/volume.rs
+++ b/src/adapters/volume.rs
@@ -139,7 +139,6 @@ fn lookup_drop_volumes_with_mounts(
 
 struct Directory<'a> {
     location: &'a Location,
-    is_remote: bool,
     synchronous: bool,
 }
 
@@ -155,7 +154,6 @@ impl<'a> Directory<'a> {
                 .is_some_and(|path| !native_query_may_leave_mount(path));
         Self {
             location,
-            is_remote,
             synchronous,
         }
     }
@@ -215,9 +213,10 @@ impl PendingVolumeLookup {
                 state.remaining -= 1;
                 continue;
             }
-            let is_remote = directory.is_remote;
+            let file = gio_file_for_location(directory.location);
+            let backend = file.uri_scheme().unwrap_or_default().to_string();
             let state = state.clone();
-            gio_file_for_location(directory.location).query_info_async(
+            file.query_info_async(
                 gio::FILE_ATTRIBUTE_ID_FILESYSTEM,
                 gio::FileQueryInfoFlags::NONE,
                 glib::Priority::DEFAULT,
@@ -225,7 +224,7 @@ impl PendingVolumeLookup {
                 move |result| {
                     let identity = result
                         .ok()
-                        .and_then(|info| identity_from_gio(&info, is_remote));
+                        .and_then(|info| identity_from_gio(&info, &backend));
                     PendingState::resolve(&state, index, identity);
                 },
             );
@@ -326,7 +325,7 @@ fn native_volume_identity_with(
             None::<&gio::Cancellable>,
         )
         .ok()?;
-    identity_from_gio(&info, false)
+    identity_from_gio(&info, "file")
 }
 
 pub(crate) fn location_is_remote(location: &Location) -> bool {
@@ -339,13 +338,13 @@ pub(crate) fn location_is_remote(location: &Location) -> bool {
     }
 }
 
-fn identity_from_gio(info: &gio::FileInfo, is_remote: bool) -> Option<VolumeIdentity> {
+fn identity_from_gio(info: &gio::FileInfo, backend: &str) -> Option<VolumeIdentity> {
     let filesystem_id = info.attribute_string(gio::FILE_ATTRIBUTE_ID_FILESYSTEM)?;
     if filesystem_id.is_empty() {
         return None;
     }
     Some(VolumeIdentity {
         filesystem_id: filesystem_id.to_string(),
-        is_remote,
+        backend: backend.into(),
     })
 }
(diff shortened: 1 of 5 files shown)


l0gicgate (commented):
Reviewed the complete diff, alias lookup and drop-policy regressions, backend namespace separation, and the keyboard fixture adjustment. No merge-blocking code findings identified. The sole added doc comment records the non-obvious native-path/file-URI namespace invariant and is retained. The latest fetched main is already included in this branch; no base merge or review fixes were necessary.

At the owner’s explicit request, publication proceeds without completing local validation. Full local validation is not claimed as passed. A live NAS/manual GUI smoke test was not performed; no UI changes or new media were produced during review.

l0gicgate:
Validation clarification: the already-running pinned Rust test suite completed before validation was stopped: 1341 passed, 0 failed, 20 ignored. Formatting, Clippy, canonical E2E, and manual NAS verification were not completed during this review, per the owner override. No code changes were needed; the remote head remains 32c3d393382f1fc6285e8c398a94ca4a3e5b2e0c.

PR #765, opened 2026-09-10
fix(preview): release media pipeline on clear

### LINKED ISSUE #764: Media previews retain GPU memory after replacement
### Strata version

Post-PR #318 (`553b788`), reproduced by the reporter on a build containing the accelerated quick-preview changes.

### Installation method

Not reported.

Linux with hardware-accelerated media previews enabled; exact distribution, desktop environment, and GPU driver were not reported.

### What happened?

Repeatedly previewing GIFs or videos causes Strata's RAM and process-attributed VRAM to climb. The reporter observed RAM rising approximately 300 → 500 → 700 MB and VRAM reaching approximately 2 GB after 3–4 previews. The memory is released only when Strata exits.

The active preview is paused in `src/ui/preview.rs::clear_content`, but its `GtkMediaFile` is not explicitly cleared. This may leave the GTK/GStreamer decode pipeline and GPU surfaces alive after the preview is replaced.

### Steps to reproduce

1. Enable hardware-accelerated media previews.
2. Open quick preview for a large GIF or video.
3. Move through 3–4 distinct GIF/video files while monitoring Strata's RAM and process-attributed GPU memory.
4. Close the preview while leaving Strata running.
5. Observe that memory accumulates across previews and is released only when Strata exits.

### Expected behavior

Replacing or closing a media preview should tear down its media pipeline. GPU memory should return to a stable baseline or bounded allocator plateau instead of increasing linearly with each preview.

### Screenshots or video

N/A — memory-lifecycle defect; no capture was provided.

### Relevant logs

No relevant logs were provided.

### Additional context

PR #318 did not introduce the pause-only teardown, but faster cycling made the existing lifecycle problem more visible. Its new preview cache is bounded to 128 MiB, so cache retention should be evaluated separately from unbounded VRAM growth.

JoeJoeflyn:
## Description

Explicitly clear the active `GtkMediaFile` when preview content is replaced or closed. Pausing alone leaves the input stream and GTK/GStreamer pipeline attached, which can retain hardware decode resources until Strata exits.

Adds focused lifecycle coverage proving teardown detaches the media source.

## Visual evidence

image

## How to test

1. Enable hardware-accelerated media previews and open a large GIF or video.
2. Move through several distinct GIF/video files, then close the preview while leaving Strata running.

**Expected result:**

Process-attributed GPU memory reaches a stable plateau instead of increasing linearly with every preview, and closing/replacing a preview releases its media pipeline.

## Related issue

Closes #764

### DIFF
--- a/src/ui/preview.rs
+++ b/src/ui/preview.rs
@@ -869,52 +869,7 @@ impl PreviewState {
                     }
                 });
 
-                let picture = gtk::Picture::for_paintable(&media);
-                picture.add_css_class("preview-media");
-                picture.set_content_fit(gtk::ContentFit::Contain);
-                picture.set_hexpand(true);
-                picture.set_vexpand(true);
-                picture.set_cursor_from_name(Some("grab"));
-                install_preview_drag(&picture, self);
-
-                let overlay = gtk::Overlay::new();
-                overlay.set_child(Some(&picture));
-                overlay.set_focusable(true);
-                overlay.set_can_target(true);
-
-                let center_play = gtk::Button::new();
-                center_play.add_css_class("preview-media-center");
-                center_play.set_halign(gtk::Align::Center);
-                center_play.set_valign(gtk::Align::Center);
-                center_play.set_visible(false);
-                let center_icon = crate::assets::primary_icon(crate::assets::icons::PLAY, 48);
-                center_play.set_child(Some(&center_icon));
-                overlay.add_overlay(&center_play);
-
-                let media_for_center = media.clone();
-                center_play.connect_clicked(move |_| {
-                    if media_for_center.is_playing() {
-                        media_for_center.pause();
-                    } else {
-                        media_for_center.play();
-                    }
-                });
-
-                let media_for_click = media.clone();
-                let overlay_for_focus = overlay.downgrade();
-                let click = gtk::GestureClick::new();
-                click.connect_pressed(move |_, _, _, _| {
-                    if let Some(overlay) = overlay_for_focus.upgrade() {
-                        overlay.grab_focus();
-                    }
-                    if media_for_click.is_playing() {
-                        media_for_click.pause();
-                    } else {
-                        media_for_click.play();
-                    }
-                });
-                picture.add_controller(click);
-
+                let (overlay, center_play) = self.build_media_view(&media);
                 self.content.append(&overlay);
 
                 if is_gif {
@@ -978,6 +933,56 @@ impl PreviewState {
                 );
             }
         }
+    }
+
+    fn build_media_view(self: &Rc<Self>, media: &gtk::MediaFile) -> (gtk::Overlay, gtk::Button) {
+        let picture = gtk::Picture::for_paintable(media);
+        picture.add_css_class("preview-media");
+        picture.set_content_fit(gtk::ContentFit::Contain);
+        picture.set_hexpand(true);
+        picture.set_vexpand(true);
+        picture.set_cursor_from_name(Some("grab"));
+        install_preview_drag(&picture, self);
+
+        let overlay = gtk::Overlay::new();
+        overlay.set_child(Some(&picture));
+        overlay.set_focusable(true);
+        overlay.set_can_target(true);
+
+        let center_play = gtk::Button::new();
+        center_play.add_css_class("preview-media-center");
+        center_play.set_halign(gtk::Align::Center);
+        center_play.set_valign(gtk::Align::Center);
+        center_play.set_visible(false);
+        let center_icon = crate::assets::primary_icon(crate::assets::icons::PLAY, 48);
+        center_play.set_child(Some(&center_icon));
+        overlay.add_overlay(&center_play);
+
+        let media_for_center = media.clone();
+        center_play.connect_clicked(move |_| {
+            if media_for_center.is_playing() {
+                media_for_center.pause();
+            } else {
+                media_for_center.play();
+            }
+        });
+
+        let media_for_click = media.clone();
+        let overlay_for_focus = overlay.downgrade();
+        let click = gtk::GestureClick::new();
+        click.connect_pressed(move |_, _, _, _| {
+            if let Some(overlay) = overlay_for_focus.upgrade() {
+                overlay.grab_focus();
+            }
+            if media_for_click.is_playing() {
+                media_for_click.pause();
+            } else {
+                media_for_click.play();
+            }
+        });
+        picture.add_controller(click);
+
+        (overlay, center_play)
     }
 
     fn render_pdf_viewer(
--- a/src/ui/preview.rs
+++ b/src/ui/preview.rs
@@ -1439,6 +1439,9 @@ impl PreviewState {
                 stream.disconnect(handler);
             }
             stream.set_playing(false);
+            if let Some(media_file) = stream.downcast_ref::<gtk::MediaFile>() {
+                media_file.clear();
+            }
         }
         self.media_toggle_mute.replace(None);
         self.media_volume_slider.replace(None);
--- a/src/ui/preview.rs
+++ b/src/ui/preview.rs
@@ -901,10 +901,12 @@ impl PreviewState {
                 });
 
                 let media_for_click = media.clone();
-                let overlay_for_focus = overlay.clone();
+                let overlay_for_focus = overlay.downgrade();
                 let click = gtk::GestureClick::new();
                 click.connect_pressed(move |_, _, _, _| {
-                    overlay_for_focus.grab_focus();
+                    if let Some(overlay) = overlay_for_focus.upgrade() {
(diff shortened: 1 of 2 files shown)


l0gicgate (commented):
Reviewed at `51c8a16` (implementation self-review, not an independent approval).

The fix now addresses both resource teardown and ownership:
- `MediaFile::clear()` detaches the active media source on replacement/close.
- Weak ancestor captures break the overlay/click-controller and seek-scale/drag-controller reference cycles.
- Lifecycle tests exercise shared production widget construction and assert overlay, picture, and media finalization after close and repeated replacement. They no longer depend on decoder behavior, which caused the previous CI failures before ownership assertions ran.

Regression strength was checked by restoring each strong capture: the overlay mutation fails with `overlay must finalize`; the seek mutation fails replacement with `media must finalize`. Both mutations were reverted.

Local validation: pinned Rust suite 1,368 passed / 0 failed / 20 ignored; targeted preview suite 20 passed; canonical quick-preview E2E 3 passed; formatting, Clippy, and diff checks passed.

No blocking findings in the final diff. Remaining limits: these lifecycle tests do not validate media decoding or quantify hardware VRAM reclamation. GitHub CI is still in progress on this head and must pass before merge.

l0gicgate:
### Why I added these changes

The original `MediaFile::clear()` fix is useful: pausing a preview does not detach its media source, while clearing it tears down that source. I kept this fix.

Code review also found two strong-reference cycles that clearing the source alone does not break:
- Overlay → picture → click controller → callback → overlay.
- Seek scale → drag controller → callback → seek scale; the callback also retains the media object.

I changed those ancestor captures to weak references so removing a preview can destroy its widgets and media object, rather than leaving an unreachable hierarchy alive.

I added close/replacement lifecycle tests to check that the overlay, picture, and media object actually finalize. The first version used decoder-backed fixtures and passed locally, but failed in the pinned CI environment before reaching the ownership assertions. To make these ownership tests portable, I extracted the existing production widget construction into `build_media_view` and test that same construction with an empty `MediaFile`. The production callbacks are preserved, and the separate source-detachment test remains. This tests ownership without pretending to test codec playback.

Validation confirms the tests catch both bugs: restoring the strong overlay capture fails overlay finalization; restoring the strong seek capture fails media finalization on replacement. The final pinned Rust suite passes (1,368 passed, 20 ignored), as do targeted preview tests, quick-preview E2E, formatting, and Clippy.

Before/after benchmarks are being collected separately. No hardware VRAM reduction is claimed yet; I will post the measured results and their limitations in a follow-up.

l0gicgate:
### Before/after ownership benchmark

Measured fixed head `51c8a16` against reconstructed pre-fix and clear-only variants, with the same instrumentation and environment. Five fresh isolated GTK processes per variant; one initial preview, 100 replacements, then close.

| Variant | Live overlays / pictures / MediaFiles after close | Mean RSS increase from initial preview |
|---|---:|---:|
| Reconstructed pre-fix: strong captures, pause-only teardown | 101 / 101 / 101 | 1,818.4 KiB |
| Clear-only: strong captures, `MediaFile::clear()` retained | 101 / 101 / 101 | 1,740.8 KiB |
| Fixed `51c8a16` | **0 / 0 / 0** | 660.0 KiB |

Object counts were identical across all five repetitions. During replacement, fixed retained only the active hierarchy; both leaking variants retained every hierarchy. Fixed RSS was approximately flat after cycle 10 (mean delta 645.6 KiB at cycle 10 versus 660.0 KiB at cycle 100). RSS includes allocator/GTK caching and is not a byte-for-byte leak measurement.

**What this establishes:** both widget reference cycles are eliminated in the production widget/control construction path. Clearing the media source alone does not eliminate those cycles.

**Important limits:** this uses codec-independent empty `MediaFile` objects, shared production builders and teardown, private Xvfb/private D-Bus, and process `VmRSS`. It does not decode/play media, measure decoder resources, or measure GPU VRAM. The pre-fix variants are explicitly reconstructed source variants, not historical executable benchmarks. These results therefore do not yet establish that the reported 1.7–2 GB VRAM growth is resolved during real playback.

[Full report, per-cycle statistics, raw CSVs, instrumentation, variant patches, and reproduction scripts](https://gist.github.com/l0gicgate/dc268bdd818bd12f509686d9ad636f75). When downloading the flat Gist files, place the CSVs in `raw/` and the two scripts in `scripts/` as used by the report; make scripts executable and use your local Xvfb path.

All 15 measured runs succeeded. Temporary benchmark changes were removed; the worktree is clean.

ogarza:
## Real-playback NVIDIA before/after comparison

We manually reproduced #764 on a real GPU, then repeated the workflow with this PR's changes. **The large per-preview VRAM retention is substantially addressed; the remaining RAM/FD/small-VRAM growth is tracked separately in #779.**

### Environment and provenance

- Strata 0.14.0 native debug builds; Omarchy/Arch + Hyprland; GTK **4.22.4**; RTX **5090**; NVIDIA **610.57.04**.
- Actual backend **`GtkGstMediaFileBuiltin`**, renderer **`GskVulkanRenderer`**.
- Unfixed build: upstream **`63f1063`** plus opt-in action/cache/backend/finalization traces.
- Fixed build: that **same baseline plus #765 at `51c8a16`**, retaining the same instrumentation. This is not a pristine executable of the PR's older tree.
- User-operated desktop playback, private D-Bus and private settings/cache profile. External `nvidia-smi -q -x` includes graphics clients; `/proc` measurements distinguish the main Strata PID from helpers. No GPU polling was added to the UI thread.
- **Both measured sessions used software conversion**, with zero GPU devices exposed to the media helpers. Disabling this preference does not disable GTK playback/rendering GPU use.

### Results (main Strata process only)

| Measurement | Unfixed | With this PR |
|---|---:|---:|
| VRAM before first preview | 60 MiB | 60 MiB |
| Added retained VRAM across **five MP4 opens** | +2,832 MiB | **+26 MiB** |
| Final closed-preview VRAM | 5,577 MiB | **114 MiB** |
| Observed session playback peak | 5,577 MiB | 680 MiB |
| Media objects finalized / created | 0 / 10 | **9 / 9** |
| Pictures finalized / created | 0 / 10 | **9 / 9** |
| Overlays finalized / created | 0 / 10 | **9 / 9** |
| Pre-preview → final RSS | 252 → 1,560 MiB | 184 → 682 MiB |

The fixed media hierarchies finalized **58–65 ms after cleanup began**. The old roughly half-GiB staircase was replaced by playback allocations that drop when the preview closes. Fixed VRAM stayed at 114 MiB for about **20 seconds after the final video closed**, including the subsequent image sequence; the last reported sample preceded process teardown.

Only **two media FFmpeg conversions per session** occurred (initial GIF and MP4); subsequent media opens were cache hits. Encoded outputs matched at **490,175 bytes** and **4,634,021 bytes**, with matching prepared dimensions. The encoded cache ended around **6 MiB**. Still-image-only phases did not grow VRAM. This implicates playback lifecycle, not repeated conversion or cache capacity, in the dominant original symptom.

### Limits and remaining investigation

These were two manual runs, not repeated identical-work benchmarks: unfixed had **5 GIF + 5 MP4 + 25 image opens**; fixed had **4 GIF + 5 MP4 + 29 image opens**, with different timing/order. Final totals are descriptive. The equal-count five-MP4 phase is the strongest comparison: approximately **99.1% less retained VRAM increase**, not a universal memory-reduction claim.

The fixed run still shows smaller retention:
- Post-GIF-close RSS: **338.4 → 374.9 → 417.1 → 451.4 MiB**.
- Corresponding FDs: **111 → 120 → 129 → 138**.
- Final FDs **180**, versus **71** before previews; final RSS **682 MiB**.
- Closed-preview VRAM rises by a few MiB per cycle (GIFs **75, 79, 84, 88 MiB**), rather than ~0.5 GiB.

Nine cycles do not prove whether this residual growth plateaus, and RSS alone does not prove a leak. #779 will identify surviving FD/thread/resource owners after GTK object finalization before proposing another fix. These residuals do **not** negate the substantial improvement from this PR.

Sampling was roughly 500 ms (individual query windows approximately 90–300 ms). Both captures had **zero trace-parse/GPU-query errors**. Missing GPU records and unavailable values were not converted to zero. Raw application logs/private settings are not uploaded because they can contain private paths.

**Recommendation:** retain these lifecycle fixes; include software-mode reproduction in the manual test instructions, and track the smaller residual issue separately rather than claiming all memory growth is resolved.

### Durable evidence and follow-up

[Sanitized report, numeric samples, structured traces, and build provenance](https://github.com/ogarza/strata/blob/73aa97e96b3d1c761ad9f5eec5807548f90059ec/docs/evidence/779/README.md) are preserved in draft **#782**, stacked on this PR. That follow-up adds opt-in FD/thread-category and PSS snapshots for **#779**; it does not claim to fix the residual growth or close that issue. Raw application logs and private settings remain local.

PR #782, opened 2026-09-10
chore(preview): add diagnostics for residual resource retention

### LINKED ISSUE #779: Media previews still accumulate RAM and file descriptors after player teardown
### Strata version

Strata 0.14.0, native debug source build: upstream `63f1063589b4a523612ef606a847be05fa6ffce3` plus the production changes from #765 at `51c8a16b5154887caec3b4e29206c653df123cd1` and opt-in diagnostic instrumentation. This is an otherwise-equivalent comparison build, not a pristine PR-head executable.

### Installation method

Built from source.

Omarchy (Arch Linux), Hyprland, Linux `7.2.3-arch1-3`; GTK 4.22.4; NVIDIA GeForce RTX 5090, driver 610.57.04. Actual media backend `GtkGstMediaFileBuiltin`; GTK renderer `GskVulkanRenderer`. User-operated desktop playback, private D-Bus session, private settings/cache profile. Both measured runs used **software media conversion**, with zero GPU devices exposed to each conversion helper.

### What happened?

Follow-up to #764 / #765. #765 removes the dominant ~0.5 GiB-per-open VRAM retention in real playback, but smaller VRAM growth and substantial RAM/file-descriptor retention remain after the preview closes and **all observed media/widget objects finalize**.

Fixed-build results after repeatedly previewing the same GIF:

| After close | RSS MiB | Open FDs | VRAM MiB |
|---|---:|---:|---:|
| GIF 1 | 338.4 | 111 | 75 |
| GIF 2 | 374.9 | 120 | 79 |
| GIF 3 | 417.1 | 129 | 84 |
| GIF 4 | 451.4 | 138 | 88 |

Across the fixed session (4 GIF, 5 MP4, 29 still-image opens), pre-preview → final closed-preview RSS increased **184 → 682 MiB**, FDs **71 → 180**, and VRAM **60 → 114 MiB**. Several consecutive media cycles leave roughly nine additional FDs. Thread counts stay elevated, but shared worker pools complicate attribution. The encoded cache ends at only **6.1 MiB**.

All nine media players, nine pictures, and nine overlays finalized within **58–65 ms** of cleanup starting. The small GPU/RAM/fd growth must therefore be investigated separately from the GTK object reference cycles already addressed in #765. Counts alone do not establish whether GTK/GStreamer/driver resources, allocator high-water retention, or another owner is responsible. Nine cycles do not prove a bounded plateau or an unbounded leak.

### Steps to reproduce

1. Use a build containing #765; start a fresh process with software preview conversion selected. Keep toolkit, driver, view, and media constant.
2. Record a pre-preview baseline for the main Strata PID, not the launcher or desktop GPU total.
3. Open/play/close the same GIF repeatedly, allowing a closed-preview settling interval after each open. Repeat with the same MP4. Confirm reopenings hit the encoded cache and do not relaunch FFmpeg.
4. Switch repeatedly between JPEG/WebP previews as a control, then close the preview while leaving Strata alive.
5. Compare settled RSS, FD/thread identities, and graphics-inclusive per-PID VRAM. Extend to 30–50 media cycles to test whether the residual growth plateaus.

### Expected behavior

After warm-up, closing/replacing previews should leave bounded resources rather than increasing the post-close baseline on every cycle. Identify surviving resources and their owner before selecting a fix. Do not declare RSS growth a leak or allocator caching without evidence, and do not mask growth by changing the cache limit or disabling GPU use globally.

### Screenshots or video

N/A — resource-lifecycle measurements, no intended visual change.

### Relevant logs

Sanitized trace/measurement summary from the fixed run:

```text
backend=GtkGstMediaFileBuiltin renderer=GskVulkanRenderer
preview-media helpers=2; policy=software; gpu_devices=0
GIF output=490175 bytes; MP4 output=4634021 bytes
media_created=9 media_finalized=9
picture_created=9 picture_finalized=9
overlay_created=9 overlay_finalized=9
encoded_cache_bytes=6385054
final_closed_vram_mib=114 rss_mib=682.016 fds=180
```

Capture: 568 valid trace records, 163 external samples at roughly 500 ms intervals, no GPU query/trace parse errors. Final 114 MiB persisted for approximately 20 seconds after the last video close, including subsequent image previews. Missing GPU records were not treated as zero. No private filenames, settings, raw application logs, or core dumps are attached.

### Additional context

**What #765 already improves:** across five MP4 opens, retained VRAM increases +2,832 MiB before versus +26 MiB after. Final session VRAM is 5,577 versus 114 MiB; baseline had 10 media opens, fixed run 9, so final totals are descriptive rather than an equal-work benchmark. Still-image phases remain flat. The acceleration setting controls conversion, not all GTK playback/display GPU use.

**Investigation plan / completion criteria:**
- Preserve opt-in lifecycle tracing and an external sampler; no NVIDIA dependency or polling in the GTK UI thread.
- Add privacy-preserving FD categories and thread-name groups, with counts of unreadable/racing entries. Capture them at closed-preview checkpoints and collect PSS when available, not only RSS.
- Verify process identity across PID reuse; distinguish host PIDs from sandbox namespace PIDs. Do not read or publish arbitrary file contents/paths or thread stack dumps.
- Repeat one-file cycles long enough to distinguish warm-up from per-cycle retention; correlate surviving resources with actual backend, cache-hit, source-clear, and finalization events.
- If evidence identifies independent CPU and GPU owners, split fix tickets then. Currently RAM/FD growth and small VRAM increments co-occur, so they belong in one investigation rather than speculative separate fixes.
- A diagnostic PR should reference this issue, **not close it**. Resolution requires measured bounded behavior and owner-specific regression coverage.

Related: #764, #765. This follow-up should not obscure the demonstrated improvement from #765.

ogarza:
## Description

Investigate the residual RAM/FD/small-VRAM growth in #779 without guessing at another fix:

- Preserve opt-in action, cache, conversion-backend, and GTK-object lifetime traces. Helper diagnostics use a bounded pipe (64 KiB/job), not a writable host log descriptor.
- Add an external NVIDIA/`/proc` sampler; `--resources` includes path-free FD/thread categories and PSS at periodic checkpoints. No NVIDIA dependency or polling in the GTK UI thread.
- Commit sanitized before/after evidence and an investigation plan in `docs/evidence/779/README.md`.

The real-playback comparison supports #765's large improvement, but fixed-build RSS still rose **184 → 682 MiB**, FDs **71 → 180**, and post-close VRAM **60 → 114 MiB** despite all nine media hierarchies finalizing. A longer run must distinguish bounded warm-up from residual retention and identify the owner. RAM/FD and small GPU growth remain one investigation until evidence establishes independent causes.

**Rebased onto `main` after #765 merged.** This PR adds reusable diagnostics, not a fix for #779. Historical captures used the same `63f1063` baseline with/without #765 at `51c8a16`; they do not claim measurements of this PR's new resource snapshots.

## Visual evidence

N/A — non-visual diagnostics and sanitized resource measurements. No intended UI changes; raw application logs and private settings are excluded.

## How to test

1. Build this branch with `cargo build --locked` on a machine supporting its GTK media backend and NVIDIA process accounting.
2. In your desktop terminal, run `python3 scripts/preview-memory.py --resources`. It launches a private D-Bus session and separate settings profile. Keep software conversion selected for the first comparison.
3. Wait ten seconds, then repeat the same GIF/MP4 open/play/close sequence 30–50 times, allowing at least five seconds closed between cycles. Use still images as a separate control.
4. Close the debug window. Inspect the new directory under `target/preview-memory/`, aligning `object_finalized` events with main-PID RSS/PSS, FD/thread categories, and VRAM. See `docs/preview-memory-tracing.md` for interpretation and privacy limits.

**Expected result:**

Traces identify cache hits and player destruction; periodic resource snapshots identify which categories remain after close without exposing raw FD paths or arbitrary thread names. Unavailable measurements are explicit, not zero. The run supplies evidence for the next fix; continued growth does not by itself fail this diagnostic PR or establish an allocator/backend cause.

## Related issue

Refs #779. Builds on merged #765; follows up on #764. Does not close #779.


l0gicgate:
Local reproduction on head `73aa97e`: the native debug app crashed with SIGSEGV after closing the fifth preview of the same MP4. Environment: RTX 5080, NVIDIA 610.57.04, GTK 4.22.4, GLib 2.88.3, GStreamer 1.28.6; GskVulkanRenderer / GtkGstMediaFileBuiltin. Owner manually operated the desktop; native build and manual capture were explicitly authorized.

The capture contains five media opens/closes and all 15 media/picture/overlay finalization events. The fifth cleanup completed immediately before the crash. The crashing GstPlay thread runs through `gst_element_change_state` / `gst_pad_set_active` → notification emission → libgstplay → `gst_message_new_custom` → `gst_object_ref` → `g_object_ref` → `g_type_check_instance_is_fundamentally_a`. The main thread was idle in its GLib event loop. This suggests an invalid object reference during asynchronous playback teardown; the precise ownership fault is not yet established (internal libgstplay frames remain unsymbolized). No OOM kill or NVIDIA Xid appeared in the inspected kernel interval.

Important comparison limits: this run used the inherited **automatic** conversion policy, not the proposed software control, and reopen gaps were shorter than five seconds. There was one conversion and four cache hits. Resource snapshots show increasing eventfd/NVIDIA-device counts and GstGL context threads, but this interrupted run cannot establish an unbounded leak or attribute the crash to this diagnostics PR versus its base/toolkit. A base/trace-disabled comparison is still needed.

Capture and core remain local; no private media, raw logs, settings, or core dump uploaded.

l0gicgate:
Follow-up control: built the crashed revision’s exact parent, `51c8a16`, natively and launched it with `STRATA_PREVIEW_TRACE=0`, the same private profile/settings, and the external resource sampler. The owner reports 10 repeated MP4 preview cycles with no crash. Closing the application completed normally (exit code 0).

The early baseline versus settled post-cycling samples still show RSS about 223 → 560 MiB, main-PID VRAM 43 → 74 MiB, and FDs 78 → 189. The final resource snapshot includes 11 `gstglcontext` threads; eventfds increased 4 → 38 and NVIDIA-device FDs 39 → 98. These are observed resource increases, not proof of unbounded leakage. No action traces exist in this control, so the cycle count comes from the owner.

This shows the residual resource growth exists without the diagnostic instrumentation. A single non-crashing base run does **not** establish that tracing caused the earlier SIGSEGV. The next isolation step is the original PR binary (`73aa97e`) with tracing disabled. This comparison concerns the original crashed revision, not the subsequently updated PR head.

l0gicgate:
Second control completed: launched the **identical binary** from the original crashing `73aa97e` capture (SHA256 verified), with only preview tracing disabled (`STRATA_PREVIEW_TRACE=0`), reusing the private profile and external resource sampler. The owner reports 10 MP4 preview cycles without a crash; application shutdown completed with exit code 0. The capture contains 169 samples and zero GPU-query errors.

Early baseline → post-cycling checkpoint: RSS approximately 215 → 583 MiB, main-PID VRAM 43 → 69 MiB, FDs 78 → 171. The latter resource snapshot includes nine `gstglcontext` threads, 32 eventfds (baseline four), and 88 NVIDIA-device FDs (baseline 39). Cycle count is owner-reported because action traces were disabled.

Results so far: original PR binary with tracing enabled crashed after the fifth close; exact parent without tracing and identical PR binary with tracing disabled each survived one owner-reported 10-cycle run. Resource retention is present without tracing. These single runs do not prove that instrumentation causes the crash; a timing-sensitive playback-teardown fault remains a hypothesis. No raw logs, private profile, media, or core uploaded.

l0gicgate:
Further local investigation found two independent toolkit lifetime defects and produced experimental local-library patches (nothing installed or merged):

1. **GTK 4.22.4 GL context retention:** `gtk_gst_sink_dispose()` does not release the `gst_context` obtained from `gst_gl_display_create_context()`. In a standalone GtkMediaFile/Xvfb/software-GL probe, the unchanged library adds two FDs and two threads per cycle despite media finalization. Releasing that context kept FDs/threads flat through 50 cycles. The context-only NVIDIA test still crashed on the third close, so it was not a crash fix.
2. **GStreamer 1.28.6 premature worker-thread finalization:** an API message can own the last GstPlay reference and release it on the playback worker. Disposal skips joining its own thread and finalization runs before `gst_play_main()` finishes accessing the object. A standalone non-GUI forced-order probe aborts with both installed and locally compiled unpatched GStreamer. A candidate retaining the object through worker teardown passed 10 forced-order trials. Both library candidates together passed 50 isolated GTK playback cycles.

**Latest owner-operated NVIDIA capture, tracing enabled:** the original `73aa97e` app loaded both candidate libraries (verified in process maps), completed **12 prepared media cycles**, finalized all **36 media/picture/overlay objects**, and exited normally (code 0). Baseline VRAM/FDs were **43 MiB / 78**; available closed-preview samples and the final pre-shutdown resource checkpoint were **47 MiB / 91**, with no remaining `gstglcontext` threads at that checkpoint. No GPU-query errors.

**Still unresolved:** final pre-shutdown RSS/PSS remained approximately **553 / 490 MiB**. Rapid reopenings leave many closes without isolated samples; a longer settled run is needed before claiming bounded RAM or complete resolution of #779. These are experimental upstream-library patches, not Strata production changes or proof against every crash ordering. Private captures and cores remain local.

l0gicgate:
Independent-review follow-up and revised manual capture:

- Fixed the candidate GstPlay deferred-reference path to acquire its extra reference only once on repeated dispose.
- Corrected the standalone reproducer to leave the API bus flushing before dropping the application reference, with an already-in-flight message. It now joins the worker and asserts exactly one finalization rather than relying on a sleep. This documented-usage unpatched case still aborts; the candidate passed 10 trials each for main-thread, worker-thread, and repeated worker disposal. The revised combined libraries also passed another 50 isolated GTK cycles.
- Latest NVIDIA capture of the revised candidates, tracing enabled: **14 prepared media cycles**, all **42 objects finalized**, exit **0**, zero GPU-query errors. Available post-close measurements settled at **47 MiB VRAM / 88 FDs**, and later resource snapshots had **24 threads**, no `gstglcontext` threads, and five eventfds.
- **RAM is not resolved:** post-close RSS increased from approximately **326 MiB after cycle 1 to 531 MiB after cycle 14** (final PSS ~468 MiB). The requested 30–50 cycles and 30-second final settling interval were not completed; observed last post-close interval was about six seconds. Do not report this as a full plateau test or complete fix for #779.

The candidates remain local experimental toolkit builds, not installed system libraries or submitted upstream changes.

ogarza:
The memory/resource-retention claim is supported by GTK 4.22.4’s source.

 - GTK stores the newly created context in self->gst_context (https://github.com/GNOME/gtk/blob/4.22.4/gtk/media/gtkgstsink.c#L1054).
 - GStreamer documents that return value as “transfer full”: GTK owns a reference and must release it.
 - gtk_gst_sink_dispose() (https://github.com/GNOME/gtk/blob/4.22.4/gtk/media/gtkgstsink.c#L1134-L1147) releases the other contexts and resources—but omits gst_context.

 That leaves the context alive even after the media object finalizes. Its finalizer (https://github.com/GStreamer/gstreamer/blob/1.28.6/subprojects/gst-plugins-base/gst-libs/gst/gl/gstglcontext.c#L683-L752) normally stops the GL thread and releases associated resources.

 ### Can Strata force that release?

 Not safely through GtkMediaFile’s public API. The retained pointer belongs to GTK’s private sink implementation. Strata already calls MediaFile.clear(); repeating that cannot repair GTK’s missing internal release. General memory trimming
 cannot reclaim a still-referenced object either.

 The missing ownership cleanup belongs inside GTK, along the lines of:

 ```c
   g_clear_object (&self->gst_context);
 ```

 That needs appropriate teardown ordering and platform testing. We could use a patched GTK package/runtime, rather than manipulate private pointers from Strata.

 This confirms the source defect—not independently the reported 50-cycle measurements, nor that it explains all our remaining RAM growth.

ogarza:
Opened an issue on the GNOME gitlab

https://gitlab.gnome.org/GNOME/gtk/-/work_items/8406

ogarza:
They closed it as 'LLM Slop' but a MR has been approved. Turns out [@matthiasc](https://gitlab.gnome.org/matthiasc) is just an oldfart asshole and I was right.

l0gicgate:
We were able to get a PR into Gnome for the context resources not being released. Closing this now.
https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/10367

PR #639, opened 2026-09-08
fix(filter): preserve interactive results without flashing

kallupx:
## Description

Keep List and Icons filter rows attached across progressive results and query edits. Reuse rows and thumbnails by full path, preserve query focus and selection through reordering, and choose the next slot (or last remaining result) when the selected file disappears.

Connect the alternate search surface to the shared item menu and resolve recursive targets by their actual location in Columns as well. Preview, open, clipboard, rename, and other menu actions retain the result identity instead of using the underlying directory index. Filtered rename uses the existing themed form and validation helpers. The completed Space-preview routing remains in use.

The E2E image now includes bubblewrap, with runner settings for its nested sandbox, so the thumbnail scenario exercises the actual decoder. The product sandbox is unchanged. The existing marquee E2E helper now samples bound rows directly so it can observe selection while the view is scrolling; its assertions and timeout are unchanged.

## Visual evidence

List mode, right-clicking the same filtered result:

| Before: background menu | After: file menu |
|---|---|
| Before | After |

Recordings: [List before](https://github.com/kallupx/strata/raw/refs/heads/fix/620-stable-filter-results/docs/evidence/620/before/list.mp4) / [after](https://github.com/kallupx/strata/raw/refs/heads/fix/620-stable-filter-results/docs/evidence/620/after/list.mp4), [Icons before](https://github.com/kallupx/strata/raw/refs/heads/fix/620-stable-filter-results/docs/evidence/620/before/icons.mp4) / [after](https://github.com/kallupx/strata/raw/refs/heads/fix/620-stable-filter-results/docs/evidence/620/after/icons.mp4), [Columns control before](https://github.com/kallupx/strata/raw/refs/heads/fix/620-stable-filter-results/docs/evidence/620/before/columns.mp4) / [after](https://github.com/kallupx/strata/raw/refs/heads/fix/620-stable-filter-results/docs/evidence/620/after/columns.mp4).

[All screenshots, preview states, environment, and reproduction steps](https://github.com/kallupx/strata/blob/fix/620-stable-filter-results/docs/evidence/620/README.md). Captures use synthetic fixtures and an app-sized private display with popup transparency.

## How to test

1. In a disposable root, create `match-note.txt`, `alpha/match-note.txt`, and `beta/match-note.txt` with different contents, plus `match-note-other.md` and a generated PNG named `beta/thumb.png`.
2. Open the root in List. Press Ctrl+F, enter `match-note`, select the beta result, focus the query, and press Space.
[...]

PR #800, opened 2026-09-11
fix(filter): refresh results after rename or delete from the result menu

### LINKED ISSUE #720: fix(filter): refresh results after rename or delete from the result menu
### Strata version

v0.14.0-rc.2 (commit `065af2d`)

### Installation method

Built from source

### What happened?

Rename or delete from a filter-result context menu updates the real file, but the filter index/`search_results` snapshot is not refreshed. The stale row stays on screen. Clicking it then tries to open the old path (`No such file` / “Unable to open file”).

`#639` already routes Columns/Icons/List item menus at the result's real location (`search_result_entry`), and `test_filtered_rename_targets_the_nested_duplicate` checks that the nested file was renamed. Neither refreshes the result list after the mutation. `index_filter` is built once when the query starts; later `query()` only re-scores that snapshot. Directory `EntriesSpliced` events do not update `search_results`.

### Steps to reproduce

1. Open a folder with a file such as `report.txt`. Press Ctrl+F and filter for `report`.
2. Right-click the result → Rename to `renamed.txt` (or Move to Trash / Delete).
3. Confirm the file on disk changed. The filter still shows the old name.
4. Click the leftover result.

### Expected behavior

After a successful rename, delete, or move from the result menu, that hit should disappear or show the new path. Activating a result must not target a path that no longer exists. The same should hold in Columns, Icons, and List.

### Screenshots or video

N/A — stale identity after a filesystem change. Related coverage in `tests/e2e/scenarios/test_filter_results.py` asserts the rename hit the nested file, not that the result list was updated.

### Relevant logs

Clicking a deleted/renamed hit goes through `activate_recursive_search_result` → `open_location` of the old `SearchItem.path`, which then surfaces “Unable to open file” (or a missing-file GIO error).

### Additional context

Related to `#697` / `#616` but distinct: those cover activation vs click preferences and context-menu selection styling, not index freshness after a mutation. Not a duplicate of `#692` (directory-only symlink/hidden mismatch). Verified on tag **v0.14.0-rc.2** (`065af2d`). Found in a Fable 5.1 exploratory pass (agent `bc-6156188a-9492-4ab5-babc-7eadfba69ed5`).

pdwaldrop:
## Description

`Browser::search_result_entry` (via `#639`) already routes item-context-menu actions on a filter/search result at the result's real location, but nothing refreshed the result list itself afterward. `SearchHandle::query()` only re-scores the snapshot `index_filter` took when the search started, so renaming, deleting, or moving a hit through that result's own context menu left the stale row on screen. Activating it then targeted a path that no longer existed (`Unable to open file`).

Both search implementations now prune results whose path no longer exists whenever a rename, delete, or transfer completes: `search_results`/`search_model` for Columns, and the `InlineSearch` used by Icons and List. Checking real path existence (rather than tracking old/new locations per operation) makes this work uniformly for rename, delete, and move, and for recursive results several directories below the open column, without depending on the directory watcher's per-column, non-recursive coverage.

## Visual evidence

N/A — the observable change is a stale row disappearing from a filtered list after its target is renamed/deleted/moved, which doesn't show anything a screenshot would make clearer than the tests below. Reproducing it manually also needs a right-click on the result, and this machine has no mouse-simulation available (Wayland/Hyprland, keyboard-only `wtype`).

Instead, `renaming_a_filtered_result_clears_the_stale_hit_in_every_view_mode` and `deleting_a_filtered_result_clears_the_stale_hit_in_every_view_mode` drive the exact rename/delete completion through the real `Browser` API in Columns, List, and Icons, and assert the stale hit disappears. Both fail on unmodified `main` (confirmed by reverting just the fix and rerunning) and pass with it applied.

## How to test

1. Open a folder with a file such as `report.txt`. Press Ctrl+F and filter for `report`.
2. Right-click the result → Rename to `renamed.txt` (or Move to Trash / Delete).
3. Confirm the file on disk changed.

**Expected result:** The stale result disappears from the filtered list once the rename/delete/move completes. The same holds in Columns, Icons, and List.

## Related issue

Closes #720

### DIFF
--- a/src/services/search/tests.rs
+++ b/src/services/search/tests.rs
@@ -406,8 +406,12 @@ fn bounded_index_reaches_a_deep_file_while_broad_folders_compete() {
     let pictures = root.join("Pictures");
     roots.push(pictures.clone());
     fs::create_dir_all(&pictures).expect("create Pictures fixture");
+    // Keep root discovery small regardless of readdir order; this tests scheduling discovered branches.
     for position in 0..15 {
-        fixture_file(&pictures, format!("screenshot-{position:02}.png"));
+        fixture_file(
+            &pictures,
+            format!("screenshots/screenshot-{position:02}.png"),
+        );
     }
     let target = fixture_file(&pictures, "test/dsds/le-cat.jpeg");
 
--- a/src/services/search/tests/multi_root.rs
+++ b/src/services/search/tests/multi_root.rs
@@ -97,9 +97,10 @@ fn fair_directory_scheduling_makes_deep_progress_in_every_root() {
         }
         expected.push(fixture_file(root, "Documents/demo/Cats/wanted.jpg"));
     }
+    // Allow storage-first discovery in both roots, while still indexing less than half the fixture.
     for ordered_roots in [roots.to_vec(), roots.into_iter().rev().collect()] {
         let (search, events) =
-            index_trees_with_budget(ordered_roots, false, 40, 64, Duration::from_secs(10));
+            index_trees_with_budget(ordered_roots, false, 80, 64, Duration::from_secs(10));
         search.query("wanted");
         let SearchEvent::Results {
             items, coverage, ..
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -542,6 +542,22 @@ impl BrowserView {
     }
 
     pub fn begin_rename(&self) -> bool {
+        if self.filter_has_focus() || self.selected_search_results().is_some() {
+            let Some(entry) = self.selected_search_result() else {
+                return false;
+            };
+            if self.state.rename_operation_pending() {
+                return false;
+            }
+            self.state.cancel_new_entry();
+            context_menu::rename_context_entry(
+                &self.state,
+                self.state.destination_depth().unwrap_or(0),
+                None,
+                entry,
+            );
+            return true;
+        }
         self.state.begin_rename()
     }
 
--- a/src/ui/browser/columns.rs
+++ b/src/ui/browser/columns.rs
@@ -137,6 +137,33 @@ pub(super) fn set_column_busy(column: &ColumnView, busy: bool) {
         .update_state(&[gtk::accessible::State::Busy(busy)]);
 }
 
+/// Drops search results whose path no longer exists, e.g. after a rename, delete, or move
+/// dispatched from that result's own context menu. `query()` only re-scores the snapshot
+/// `index_filter` took when the search started, so a mutated hit otherwise lingers until the
+/// query itself changes.
+pub(super) fn prune_missing_search_results(column: &ColumnView) {
+    if column.search_handle.borrow().is_none() {
+        return;
+    }
+    let mut results = column.search_results.borrow_mut();
+    let before = results.len();
+    results.retain(|item| item.path.try_exists().unwrap_or(true));
+    if results.len() == before {
+        return;
+    }
+    let labels: Vec<_> = results.iter().map(|item| item.name.clone()).collect();
+    drop(results);
+    let labels: Vec<_> = labels.iter().map(String::as_str).collect();
+    column
+        .search_model
+        .splice(0, column.search_model.n_items(), &labels);
+    column.filtered_model.items_changed(
+        0,
+        column.search_model.n_items(),
+        column.search_model.n_items(),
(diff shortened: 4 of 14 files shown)


l0gicgate:
Review found two blockers in the proposed fix: queued/query-updated snapshots can resurrect removed hits, and `try_exists()` drops dangling symlinks that still exist. Local commit `22f5987` fixes both, extends Rust/E2E coverage, and removes redundant doc comments. It is **not pushed**: the required full local Rust suite fails two pre-existing bounded-search scheduling tests, reproduced on exact upstream main `f4e2a9b3c824ddc513f1f34e1e9f67887fe7a0c5` as well:

- `services::search::tests::multi_root::fair_directory_scheduling_makes_deep_progress_in_every_root`
- `services::search::tests::bounded_index_reaches_a_deep_file_while_broad_folders_compete`

The isolated real-GUI filter scenarios pass, including nested context-menu rename and changing the query without reviving the stale result. No desktop/manual interaction was used. Current screenshots were captured locally; upload tooling is unavailable, so they are not attached. Paths: `/home/l0gicgate/Documents/github/strata-worktrees/720-pr800-review/target/review-800/filter-after-{Columns,List,Icons}.png`.

Retaining review branch `fix/720-pr800-review` in that worktree pending resolution of the baseline failures; nothing was pushed or merged. Main was already the PR parent, so the requested main merge was a no-op.

l0gicgate:
Resolved and pushed as `22f5987` and `61b19c5`.

The bounded-search failures depended on directory enumeration order. Test fixtures now leave room for storage-first discovery and separate root discovery from deep-branch scheduling, while still requiring partial coverage and all expected deep hits. Both tests pass under native, alphabetical, and reverse-alphabetical traversal; production scheduling is unchanged.

The filter fixes also prevent queued/query-updated snapshots from restoring stale hits and preserve dangling symlinks. Isolated real-GUI verification exercised nested context-menu rename and subsequent query changes in all three views; the stale result stays absent. The complete local suites and CI on the pushed head pass.

Screenshots were captured but upload tooling is unavailable. Preserved local paths: `/home/l0gicgate/.cache/pi-tmp/pr800-completed.zo3ZcZ/target/review-800/filter-after-{Columns,List,Icons}.png`. These replace the earlier temporary-worktree paths. No media was uploaded. PR remains unmerged.

l0gicgate:
Fixed the owner-reported F2 / Ctrl+R failure during guided filter testing in `46c22e3`. Both shortcuts now rename the selected search result through the existing result-menu dialog, including while the filter field has focus. Empty/unselected searches cannot fall back to renaming the hidden directory selection.

Verified real keyboard input in Columns, List, and Icons, with result and filter-field focus; the nested file changes, same-named siblings remain untouched, and the stale result does not return. The F2 regression failed in all six focus/view combinations before the fix. Full local quality and E2E suites, and CI on this head, pass.

Current dialog screenshots are preserved at `/home/l0gicgate/.cache/pi-tmp/pr800-f2-completed.QwePvn/target/review-f2/rename-{Columns,List,Icons}.png`. Upload tooling is unavailable; these were not uploaded. Existing owner test window was left untouched and still runs the previous build.

l0gicgate:
Fixed the guided-testing Escape focus regression in `a5ab9d3`. Cancelling the filtered-result Rename dialog now restores selection and keyboard focus to the exact originating path after dismissal, in Columns, List, and Icons. Successful submission does not restore the old target.

Verified context-menu, F2, and Ctrl+R entry points, including filter-field focus: Escape returns to the nested result, then F2 immediately renames that same file. Native search navigation is kept separate from directory commands so restored focus cannot expose hidden directory selections to Delete and other generic shortcuts. This does not add keyboard support for those file operations on search results.

The regression failed in all three views before the fix. Full local suites and CI on the pushed head pass. Current screenshots are local at `/home/l0gicgate/.cache/pi-tmp/pr800-focus-completed.QmvhYd/target/review-focus/escape-{Columns,List,Icons}.png`; upload tooling is unavailable, so none were uploaded. The owner’s running window remains on the preceding build until restarted.

l0gicgate:
Owner completed guided desktop testing of the latest build and confirmed “Good to go.” Verified that the tested head `a5ab9d3` is already pushed to the PR branch and all CI gates remain green. PR left unmerged.

PR #827, opened 2026-09-11
fix(archives): drop pending navigate and leftover folder on Extract to… cancel

### LINKED ISSUE #808: fix(archives): cancelling Extract to… password prompt hijacks the next completion
### Strata version

0.15.0 (`66fb0e6`)

### Installation method

Built from source

Ubuntu 24.04.4 LTS, X11, GTK 4.14.5

### What happened?

Extract to… on a password-protected archive creates the destination, stores it in `pending_navigate`, and starts extract without a password. `OperationFailed` then opens the password dialog and **does not clear** `pending_navigate` (`src/ui/browser/events.rs`). Escape / Cancel only dismisses the modal (`build_archive_modal`); it never clears that pending destination. The next archive completion — including Extract here on a later plain zip — therefore navigates into the leftover empty folder.

Not a reopen of #691 (member-name `password` false positive) or #790 (wrong-password dialog reopen).

### Steps to reproduce

1. Compress a password-protected 7z (or use a known encrypted 7z).
2. Extract to… a new folder and confirm. Wait for the password dialog.
3. Press Escape.
4. Extract here a later unencrypted `later.zip` in the original folder.
5. The window navigates into the empty leftover Extract to destination instead of staying on the origin (or selecting the extracted files there).

### Expected behavior

Cancelling the password dialog should drop `pending_navigate` for that failed Extract to…. A later successful Extract here must not navigate into the abandoned destination.

### Screenshots or video

N/A — navigation target is the leftover empty folder; AT-SPI pane name after the later extract was `leftover-dest` with an empty directory and `later.txt` still created at the origin.

### Relevant logs

```
panes=['leftover-dest']
current='leftover-dest'
later_exists=True
leftover_exists=True
leftover_contents=[]
```

`OperationFailed` password path takes `pending_extract_retry` and returns without `pending_navigate.take()`. `ArchiveCompleted` then navigates if that cell is still set.

### Additional context

Found in the 2026-09-11 full QA suite sweep on `lgse/strata` main @ `66fb0e6` (via wmfeht/strata). Evidence agent: `bc-e6731602`. Reproduced with the encrypted 7z fixture plus a later zip, and with a GTK event-handler test that `OperationFailed` + `ArchiveCompleted` navigates to the leftover dest.

JoeJoeflyn:
## Description

"Extract to…" on a password-protected archive created the destination folder eagerly and stored it in `pending_navigate` before the password was known. When the passwordless extract failed, `OperationFailed` opened the password dialog without clearing `pending_navigate`, so cancelling the dialog left the abandoned destination behind. The next archive completion (e.g. a plain "Extract here" on a different archive) then consumed that stale value and navigated into the empty leftover folder.

Two fixes:

- **Stale navigation:** `OperationFailed`'s password branch now clears `pending_navigate` before showing the dialog; the retry-confirm handler re-arms it, so a successful retry still reveals the destination but a cancelled one leaves nothing behind.
- **Leftover folder:** destination creation moves from the UI confirm handler into the extraction backend, which removes the newly-created empty directory on any non-completed outcome. Pre-existing destinations and any directory that received content are left intact.

## Visual evidence

N/A — the change is behavioural (a folder not being left behind and navigation not jumping into it). The observable result is covered by the tests below.

## How to test

1. In a folder, place an encrypted archive (e.g. `content-encrypted.7z`, password `secret`) and a plain `later.zip` containing `later.txt`.
2. Right-click the encrypted archive → **Extract to…** → name it `leftover` → confirm.
3. When the password dialog appears, press **Escape** (or Cancel).
4. Right-click `later.zip` → **Extract here**.

**Expected result:** The window stays on the current folder and reveals `later.txt` there. No empty `leftover/` directory remains on disk.

## Related issue

Closes #808


l0gicgate:
Reviewed and pushed `135919a`, including merge commit `aeee28d` for verified `lgse/strata:main` at `4b894fe`.

Fixed a retry regression: password confirmation now restores the original navigation intent rather than making **Extract here** navigate like **Extract to…**. Added GTK coverage for both modes and an E2E regression for cancelling the password prompt followed by another extraction. Audited every added comment and removed all four redundant ones.

The isolated GUI exercise confirmed no leftover destination, no navigation jump, and `later.txt` selected at the origin. No separate manual desktop run was performed.

After cancelling Extract to and extracting another archive

Local pinned quality suite passed (1,464 Rust tests; 20 existing ignores), as did all 812 E2E tests. An earlier breadcrumb E2E attempt failed with off-screen pointer coordinates; the base-commit check and final complete rerun passed without changing that unrelated code. CI is green on the pushed head. The PR remains unmerged.

PR #846, opened 2026-09-12
fix(dnd): defer autofs volume probes off the GTK thread

### LINKED ISSUE #732: fix(dnd): keep autofs volume lookups off the GTK thread
### Strata version

0.14.0 source build at `aacda2b5c8cdf8e526bd237fde0ff39d394cc506`; affected code verified unchanged in merged main `f8ab4fedb10fb82ce3e8bb892678ddbba03eb6c0` (#502).

### Installation method

Built from source

Omarchy / Arch Linux. The regression probe ran through the private Xvfb/X11 and D-Bus test runner with disposable HOME/XDG. No real automount or desktop session was modified.

### What happened?

Drop-volume lookup treats an `autofs` path as eligible for synchronous probing on the GTK thread. `Directory::classify` can walk path prefixes with `symlink_metadata`, followed by synchronous GIO filesystem-ID queries, before any asynchronous timeout applies.

For a path through an untriggered automount backed by an unavailable server, these operations can block drag-hover and the UI. Autofs is already excluded from Trash fallback scans, but not from drop lookup's potentially-blocking classification.

**Confirmed:** a synthetic autofs mount-table regression selects the synchronous path. **Not independently reproduced:** an actual UI hang against an unavailable autofs server. The issue is the unsafe scheduling decision; no real server outage is required for deterministic regression coverage.

### Steps to reproduce

Deterministic code-level reproduction in an adjacent volume test:

1. Create a temporary directory and a synthetic `MountTable` with an ordinary `/` entry and an `autofs` entry at that directory.
2. Construct a native `Location` for a descendant, such as `<fixture>/share`.
3. Call `Directory::classify(&location, &mounts)`.
4. Assert that `resolves_synchronously()` is false. This assertion fails on the reviewed code.

A separate manual check can use an expendable automounted network share in an isolated session: hover a drag over a destination beneath an untriggered mount while its server is unavailable. Never deliberately stall the user's live mount for this test.

### Expected behavior

Detect potentially blocking automount paths from mountinfo **before** doing filesystem probes on the GTK thread. Route their lookup through the existing asynchronous/cancellable path, preserving timeout feedback and UI responsiveness.

This scheduling decision must not, by itself, declare two filesystem identities different.

### Screenshots or video

N/A — deterministic scheduling regression; no real-server UI hang recording was made.

### Relevant logs

Sanitized review-probe result:

```text
private_review_autofs_must_not_use_synchronous_lookup ... FAILED
assertion failed: !directory.resolves_synchronously()
```

### Additional context

Follow-up to merged #502. Keep this as a small scheduling fix, not a general volume subsystem rewrite.

Relevant code at the merge:
- [Directory classification and prefix probing](https://github.com/lgse/strata/blob/f8ab4fedb10fb82ce3e8bb892678ddbba03eb6c0/src/adapters/volume.rs#L147-L185)
- [Mount filesystem classifications](https://github.com/lgse/strata/blob/f8ab4fedb10fb82ce3e8bb892678ddbba03eb6c0/src/adapters/volume/mounts.rs#L77-L137)

Acceptance:
- [ ] Autofs descendants are classified potentially blocking before caller-thread filesystem probes.
- [ ] Add deterministic regression coverage for autofs, while retaining ordinary local lookup behavior and existing timeout/cancellation behavior.
- [ ] Preserve existing Copy/Move/Ask policy and restore security checks; do not redefine filesystem equality as part of this fix.

l0gicgate:
## Description

Detect autofs ancestors from mountinfo before drag/drop volume classification can probe path prefixes on the GTK thread. Use the existing asynchronous, cancellable lookup and timeout, including for nested local mounts beneath autofs.

Keep this scheduling decision separate from remote classification and filesystem identity; Copy/Move/Ask policy and Trash restore safeguards are unchanged. Add a no-probe regression and extend existing async/identity fixtures with synthetic autofs mounts.

## Visual evidence

N/A — scheduling-only fix with no visual changes. No live-server outage or UI hang was induced; reproduction uses synthetic mount tables.

## How to test

1. In a disposable isolated session only, prepare an expendable autofs share backed by an unavailable test server. Do not interrupt a live/user mount.
2. Hover a file drag over a destination beneath the untriggered automount, then leave the target or cancel the drag.
3. Repeat ordinary local and cross-volume drops in Columns, List, and Icons, using the Copy/Move/Ask preferences.

**Expected result:** Drag-hover remains responsive and unresolved identity falls back through the existing timeout behavior. Leaving cancels feedback. Ordinary local drops and cross-volume policy remain unchanged; autofs scheduling alone does not mean the filesystems differ.

## Related issue

Closes #732

### DIFF
--- a/src/adapters/volume.rs
+++ b/src/adapters/volume.rs
@@ -144,14 +144,22 @@ struct Directory<'a> {
 
 impl<'a> Directory<'a> {
     fn classify(location: &'a Location, mounts: &MountTable) -> Self {
-        let is_remote = match location.native_path() {
-            Some(path) => mounts.is_remote_path(path),
+        Self::classify_with_probe(location, mounts, native_query_may_leave_mount)
+    }
+
+    fn classify_with_probe(
+        location: &'a Location,
+        mounts: &MountTable,
+        native_probe: impl FnOnce(&Path) -> bool,
+    ) -> Self {
+        let may_block = match location.native_path() {
+            Some(path) => mounts.query_may_block(path),
             None => location_is_remote(location),
         };
-        let synchronous = !is_remote
+        let synchronous = !may_block
             && location
                 .native_path()
-                .is_some_and(|path| !native_query_may_leave_mount(path));
+                .is_some_and(|path| !native_probe(path));
         Self {
             location,
             synchronous,
--- a/src/adapters/volume/mounts.rs
+++ b/src/adapters/volume/mounts.rs
@@ -48,6 +48,15 @@ impl MountTable {
         self.fs_type_for(path).is_some_and(is_remote_fs_type)
     }
 
+    pub(super) fn query_may_block(&self, path: &Path) -> bool {
+        // Prefix probes can trigger autofs even when a nested mount is local.
+        self.is_remote_path(path)
+            || self
+                .entries
+                .iter()
+                .any(|(mount_point, fs_type)| fs_type == "autofs" && path.starts_with(mount_point))
+    }
+
     pub(super) fn is_mount_point(&self, path: &Path) -> bool {
         self.entries
             .iter()
--- a/src/adapters/volume/tests.rs
+++ b/src/adapters/volume/tests.rs
@@ -294,34 +294,40 @@ fn native_path_and_file_uri_of_the_same_directory_share_an_identity() {
     );
 }
 
-fn mounts_treating_as_nfs(path: &Path) -> MountTable {
+fn mounts_treating_as(path: &Path, fs_type: &str) -> MountTable {
+    let mount_point = path
+        .to_str()
+        .expect("UTF-8 fixture path")
+        .replace('\\', "\\134")
+        .replace(' ', "\\040")
+        .replace('\t', "\\011")
+        .replace('\n', "\\012");
     MountTable::parse(format!(
-        "1 0 0:1 / / rw - ext4 /dev/root rw\n2 1 0:2 / {} rw - nfs4 server:/export rw\n",
-        path.display()
+        "1 0 0:1 / / rw - ext4 /dev/root rw\n\
+         2 1 0:2 / {mount_point} rw - {fs_type} fixture rw\n",
     ))
 }
 
 #[test]
-fn native_path_on_a_remote_mount_is_looked_up_asynchronously() {
+fn native_path_on_a_blocking_mount_is_looked_up_asynchronously() {
     let root = tempfile::tempdir().expect("tempdir");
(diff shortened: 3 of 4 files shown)


l0gicgate:
## Guided manual testing — results and methodology

**Build:** `9e6901ff5d91f4b03f50b81f442f14c80ec276c2` (this PR's head). **Result: all six checks passed.**

### Methodology

- Built and launched the source worktree on Omarchy/Arch with the native host toolkit. Strata ran on a private Xvfb display with private D-Bus/AT-SPI buses and disposable HOME/XDG settings. An interactive VNC viewer exposed only this test display through a private Unix socket; clipboard sharing was disabled.
- Used **Columns view**, with the cross-volume preference set to **Always ask**.
- Created six distinct disposable text files and a **Local destination** folder on the same filesystem. **Other filesystem** was a symlink to a dedicated temporary destination on `/dev/shm`; differing device IDs were verified before testing. This also exercises a destination requiring asynchronous symlink-following lookup, without a real network mount.
- The owner performed each drag and confirmed the prompt/modifier behavior. After each step, the agent checked the resulting files on disk. A final sweep verified all six source/destination states and exact contents of every remaining file.

### Results

| Check | Action | Confirmed outcome |
| --- | --- | --- |
| Local move | Drag `01-local-move.txt` onto Local destination | Moved without prompting; original removed. |
| Cross-filesystem copy | Drag `02-cross-copy.txt` onto Other filesystem; choose **Copy** | Choice prompt appeared; identical contents retained in both locations. |
| Cross-filesystem move | Drag `03-cross-move.txt` onto Other filesystem; choose **Move** | Choice prompt appeared; destination retained contents and original disappeared. |
| Cancel | Drag `04-cross-cancel.txt` onto Other filesystem; choose **Cancel** | Original unchanged; no destination file created. |
| Force copy | Hold **Ctrl** through dropping `05-force-copy.txt` onto Other filesystem | Copied without prompting; original preserved. |
| Force move | Hold **Shift** through dropping `06-force-move.txt` onto Other filesystem | Moved without prompting; original removed. |

### Scope and limitations

These manual checks validate unchanged local/cross-filesystem drop policy and modifier behavior in **Columns only**. List and Icons were not repeated manually in this session. This native-toolkit session is not a substitute for the canonical pinned-container gate.

**No unavailable autofs server or live UI hang was reproduced, and no live automount was modified or deliberately stalled.** The autofs-specific scheduling guarantee remains supported by the previously passing deterministic mount-table regressions, including proof that classification skips caller-thread prefix probes. The guided results should not be interpreted as a real autofs outage test.

PR #876, opened 2026-09-12
feat(columns): reveal adjacent column peek margin when navigating

### LINKED ISSUE #874: feat(columns): reveal the next column when clicking column background
### Problem or use case

In Miller Columns mode, when clicking a column's empty background to focus it, the viewport scrolls to show only that column. If there are more columns to the right (off-screen), there is no indication they exist. Users have to manually scroll or click each column one at a time to discover or reach the last column.

### Proposed solution

When clicking a column's background, the viewport should reveal the next column to the right (the child column), not just the clicked column. Each click advances rightward (click a column, see the next one), until there are no more child columns.

### Alternatives considered

N/A

### Visual examples

N/A

JoeJoeflyn:
## Description

In Miller Columns mode, when focusing or navigating columns, adjacent offscreen columns (such as the next child column on the right or parent column on the left) were not visible unless manually scrolled into view.

This change adds a 48px peek margin to \`horizontal_reveal_target\` when scrolling forward or backward, allowing users to see a sliver of the adjacent column and discover/click into it directly, while keeping the root and leaf columns cleanly aligned to boundaries.

## Visual evidence

https://github.com/user-attachments/assets/512cb619-7e4d-4b51-9b81-aad4f438ae46

## How to test

1. Open Strata in Miller Columns mode.
2. Navigate deep into nested folders so multiple columns exist across the viewport width.
3. Click or navigate between columns.

**Expected result:**
When revealing a column forward, a 48px peek of the next child column is visible on the right edge (if not at the boundary). When revealing backward, a 48px peek of the parent column is visible on the left edge.

## Related issue

Closes #874

### DIFF
--- a/src/ui/browser/columns.rs
+++ b/src/ui/browser/columns.rs
@@ -359,6 +359,8 @@ pub(in crate::ui) fn max_child_natural_width(widget: &gtk::Widget) -> i32 {
     max_natural
 }
 
+pub(super) const COLUMN_PEEK_WIDTH: f64 = 48.0;
+
 fn horizontal_reveal_target(
     current: f64,
     page_size: f64,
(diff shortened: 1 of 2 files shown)


ogarza:
let's implement something similar in the rework here https://github.com/lgse/strata/pull/888

we use peeking but giving priority to the focused column because exactly 1:1 like we do it here won't work with the new flow

JoeJoeflyn:
you or me do it ?

ogarza:
I'm already working on it!

PR #892, opened 2026-09-12
fix(chooser): allow selecting a folder as save destination without navigating into it

### LINKED ISSUE #890: fix(chooser): allow selecting a folder as save destination without navigating into it
### Description
In a SaveFile or SaveFiles chooser, clicking a folder navigates into it instead of selecting it as the save destination. When a user clicks a folder and presses Save, the file should be saved inside that selected folder without navigating into it first.

### Expected Behavior
Single-clicking a folder in the SaveFile/SaveFiles chooser selects it as the destination folder for Save.

JoeJoeflyn:
## Description

In a `SaveFile` or `SaveFiles` chooser, clicking a folder in List view or Icons view previously navigated in-place into the folder instead of allowing the user to select it as the save destination.

This change:
1. Updates `ChooserState::active_folder()` to prioritize `selected_folder()` (from `selected_search_results()` or `browser().selected_entries()`) before falling back to `active_location()`.
2. Prevents single clicks on folders in chooser mode from triggering in-place navigation in List and Icons views.
3. Adds comprehensive acceptance tests covering single folder selection as the save destination across List view, Icons view, SaveFiles batch mode, and search results.

## Visual evidence

https://github.com/user-attachments/assets/713d9a65-e8a5-419d-842b-7314b482a663

## How to test

1. Run `/usr/bin/python3 scripts/portal-test.py save --binary target/debug/strata --view list`
2. Single-click any folder in the list (e.g. `Documents`).
3. Click **Save** (or press Enter in the filename field).

**Expected result:**
The file is saved inside the selected folder without requiring navigation into it first.

## Related issue

Closes #890


wmfeht:
**Blocker:** `selected_folder()` treats the post-navigate load cursor as a user pick, so Save without clicking the list writes into the first listed folder.

Chooser init calls `browser.navigate(initial_directory)`, and `navigate()` always uses `select_first: true`. The first visible row is a load cursor, not a committed selection (same invariant paste already honors in `paste_destination`: a single selected folder counts only when `!load_cursor`). GTK echoing that row does not clear the cursor; only `commit_selection()` (pointer press on the list) does. SaveFile then focuses the name field, so a typical “open Save, keep the name, click Save” never commits the list.

If that first row is a directory, `active_folder()` returns it and the file lands inside the child instead of the opened folder.

**Fix:** when not using search results, ignore `selected_folder()` while `browser.selection_is_load_cursor()` is true. Keep search-result folders (those are an explicit row select).

**Missing test:** directory whose first row is a folder; click Save without `select_item`; assert the URI is in the parent. The new tests all `select_item(0)` on a one-entry listing, so they cannot catch this.

`selected_folder()` for a committed single directory is the right Save rule; this is the load-cursor hole, not the feature itself.

Issue #766, opened 2026-09-10
feat(list vew): Select the source folder on go up/back


tordex:
In the list view mode when I enter to any folder and then go up/back, the first list item is selected. It is very uncomfortable with keyboard navigating. I think the source folder should be selected on go up/back, like this works in the columns view.

JoeJoeflyn:
@ogarza what is ur behavior? i forgot

ogarza:
Right now it goes back to top of list as designed, we can discuss of it needs to be changed.

@tordex could you explain in what workflow or scenario you would want to go back to the old entry ? What happens if that entry no longer exists?

tordex:
This useful during the navigation using the keyboard:
* Navigate over the folders with arrows
* `Enter` to enter the selected folder
* `Alt+Left` or `Alt+Up` returns back to the parent folder
* The folder I leave is selected and I can continue navigation from the same point

If it goes to the top of list, I have to "hold down Down key" to select the next folder or use mouse.
If entry no longer exists it is ok to go to the top of list. This will happen rarely, I guess.

BTW. The same issue in the Columns view. On `Alt+Left` or `Alt+Up` resets selected item to the top of column.

ogarza:
Are you navigating like this to search for a specific file, going one
subfolder at a time?

tordex:
Not just searching a file. Most likely I should check the folders content.

Selecting the source folder also helps with navigation with mouse inside the long list of directories. Now I have to scroll down and remember what directory I visited last. Selecting the source folder will make this folder or screen and keep eyes focus.

l0gicgate:
This will be fixed in next release:

https://github.com/user-attachments/assets/b1f8fe2c-e3cd-4222-884c-064d8ba41aca

tordex:
Thank you very much!

PR #893, opened 2026-09-12
feat(list): restore navigation selection and scroll position

### LINKED ISSUE #868: feat(list): restore scroll position and selection when returning from a folder
### Problem or use case

When browsing in List mode, entering a folder and then returning to its parent loses the previous browsing position and selection/highlight. In a long directory, this means scrolling back to find the folder again and interrupts keyboard navigation.

### Proposed solution

Preserve the parent directory’s List-mode scroll position and selection/highlight when entering a child folder. Restore both when returning through Back or Up navigation, so the user can continue from where they left off rather than starting at the top or highlighting the first row.

### Alternatives considered

Manually scroll back and reselect the folder after returning. Restoring only the selection is insufficient if the previous viewport is lost.

### Visual examples

N/A — navigation-state restoration; no new interface is proposed.

### Strata version

Not specified.

### Additional context

Related to #766, which requests selecting the source folder on Up/Back. This enhancement explicitly covers restoring the scroll position together with selection/highlight.

Acceptance criteria:
- In List mode, open a directory with enough items to scroll, scroll partway down, select a folder, and enter it.
- Return using Back: the previous scroll position and selected/highlighted folder are restored.
- Repeat using Up: the previous scroll position and selected/highlighted folder are restored.
- Repeat through nested folders: each previously visited parent restores its own state.
- Keyboard navigation can continue from the restored highlighted row rather than the first item.

l0gicgate:
## Description

Remember List-mode selection, keyboard cursor, and viewport per directory, restoring them after navigation finishes loading. Nested parents retain their own state; missing entries are matched safely by location. Restored selection does not redirect paste until explicitly selected.

## Visual evidence

| Before | After |
| --- | --- |
| Before | After |

Synthetic fixture; the before capture disables the new state cache to reproduce the previous behavior.

## How to test

1. In List mode, open a directory with enough folders to scroll. Scroll partway down, select a folder, and enter it.
2. Return with Back, then repeat with Up and through nested folders. Press Down after returning.

**Expected result:** Each parent restores its previous viewport and highlighted folder; Down continues from that folder rather than the first row.

## Related issue

Closes #868. Related to #766.


l0gicgate:
Owner-guided testing passed on a630250f:
- Back restores scroll position and the selected/keyboard-focused folder; Down continues to the next row.
- Up restores the same state and keyboard continuation.
- Nested Back/Up restores each parent’s own viewport and highlighted folder.

These are owner-reported results from the PR binary with host GTK 4.22.4, disposable fixtures/settings, and private D-Bus. The desktop window was explicitly requested for guided testing; this supplements, rather than replaces, the pinned-container validation.

Issue #613, opened 2026-09-08
test(gtk): investigate the offscreen new-entry fixture failure on GTK 4.14


l0gicgate:
### Strata version
0.13.0, source commit 916095ac27f097d72863883ca5c3b1aae25540e3.

### Installation method
Built from source inside the existing pinned E2E Ubuntu container (Rust 1.98.1).

### Environment
Ubuntu 24.04, GTK 4.14, rootless Podman, private Xvfb, GDK_BACKEND=x11, GTK_A11Y=none, NO_AT_BRIDGE=1, STRATA_REQUIRE_GTK_TESTS=1. Reproduces with the default renderer and cairo. The same full Rust suite passes with the newer native Arch GTK.

### Steps to reproduce
1. Build the Rust test binary inside the pinned E2E toolchain image.
2. Under private Xvfb, run `cargo test --all-targets --all-features ui::browser::inline_edit::tests::entries::new_entries_scroll_into_view_when_the_default_name_sorts_past_the_initial_viewport -- --exact --nocapture` with `STRATA_REQUIRE_GTK_TESTS=1`.
3. The first Columns/file iteration (2,000 synthetic entries) times out waiting for its inline editor. Waiting for a mapped, positive-height collection and giving it focus before creation did not resolve it.

### Expected behavior
The new entry scrolls into the visible viewport and exposes its rename editor; the fixture passes on supported GTK versions.

### Actual behavior
The file is created but the fixture does not obtain a rename field before its deadline. This is a confirmed Rust fixture failure; whether an equivalent real-input application bug exists still needs investigation.

### Sanitized logs
```text
thread ...new_entries_scroll_into_view_when_the_default_name_sorts_past_the_initial_viewport panicked at src/ui/browser/inline_edit/tests.rs:36:9:
rename fixture did not settle
```

### Additional context
Found while auditing CI coverage for #607. The current quality job has no display and does not enforce GTK initialization, so it does not expose this existing failure. Do not delete, skip, or weaken this regression test to enable a full GTK quality gate.

l0gicgate:
The migration in #634 reproduced this on the public GTK 4.14 image with GTK initialization required: 1,106 passed and this fixture failed. Inspection showed row 2000 was bound but unmapped with zero allocation; row 1999 was visible. The allocation guard prevented the next reveal request.

[...]

PR #898, opened 2026-09-12
perf(preview): bound PDF page rendering

### LINKED ISSUE #815: fix(preview): PDF preview hangs and crashes when scrolling
### Strata version

0.15.0 (`319d799`)

### Installation method

Built from source

Arch Linux, Hyprland 0.56.2

### What happened?

Opening a PDF in quick preview and scrolling down causes Strata to hang and then crash. The PDF viewer is a `GtkListView` that lazy-loads each page on `factory.connect_bind`, firing a preview load per page (`src/ui/preview.rs`, `render_pdf_viewer`). Scrolling rapidly binds/unbinds many pages and starts many concurrent preview loads, after which the window freezes and exits.

### Steps to reproduce

1. Select a multi-page PDF in a browser pane.
2. Open quick preview (space).
3. Scroll down through several pages.
4. Observe the window hangs and then crashes.

### Expected behavior

Scrolling a PDF preview should page in rendered pages smoothly without freezing or crashing, cancelling/skipping loads for pages that scroll out of view.

### Screenshots or video

N/A

### Relevant logs

No sanitized logs captured yet. If reproducible, collect with:

```shell
coredumpctl list strata
coredumpctl info strata
journalctl --user _COMM=strata --since "30 minutes ago"
```

### Additional context

Reporter environment: Arch Linux, Hyprland 0.56.2, built from source. Likely related to unbounded concurrent page loads and load cancellation on unbind (`factory.connect_unbind` removes the load handle but may not cancel in-flight rendering). Worth checking whether page loads are throttled and whether in-flight loads are cancelled when a page unbinds.

JoeJoeflyn:
## Description

Render PDF pages at the preview viewport width instead of a fixed oversized surface, and serialize sandboxed Poppler page renders across previews. This reduces first-content work and prevents rapid scrolling from multiplying renderer CPU and memory pressure. Rendered-page caching remains bounded and size-aware.

## Benchmark

Local engineering sample using the reported 179-page, 17.9 MB optimized PDF (531 × 666 pt pages), a debug build, Poppler 26.08.0, and a warm filesystem cache. The alternating first-page samples ran in one process to reduce startup variance; timings remain machine-specific.

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Median first-page render | 1,583 ms at 1400 × 1755 | 909 ms at 640 × 802 | 43% faster |
| First-page PNG payload | 938,382 bytes | 246,071 bytes | 74% smaller |
| Peak PDF-helper RSS while paging four visible pages | ~411 MiB across concurrent workers | ~112 MiB with serialized work | 73% lower |

The scheduler intentionally prioritizes first visible content and bounded memory over completing several offscreen page renders simultaneously; later visible pages are queued FIFO and cancelled if they leave the viewport.

Validation note: the owner directed this push before local full checks completed. Targeted PDF, sandbox, provider, and preview tests passed. Formatting and Clippy passed. The full Rust suite reached 1,517 passing tests but reported the pre-existing #613 inline-entry failure and one search test that passed on isolated rerun. The canonical E2E run was interrupted twice; the second attempt displayed early failures before interruption, so local E2E is not reported as passed and GitHub CI remains required.

## Visual evidence

https://github.com/user-attachments/assets/770e5eb9-807d-410e-892b-da098d186bf4

## How to test

1. Open quick preview for a large, complex, multi-page PDF.
2. Confirm the first page appears, then scroll rapidly through many pages and reopen the preview.

**Expected result:**

The first page appears sooner; page rendering remains responsive and bounded without a freeze or crash, and revisited rendered pages use the cache.

## Related issue

Closes #815


wmfeht:
**Blocker:** cancelled in-flight PDF page renders now emit `PreviewEvent::Failed`, which can blank a page that is still (or again) visible.

`abort_safe` correctly refuses to `task.abort()` while `gio::spawn_blocking` holds the PDF permit, so the next helper cannot start until Poppler is gone. Unbind still cancels the token. `wait_for_renderer` then returns `"Preview cancelled"`, and `LocalPreviewProvider::load` emits `Failed`.

The PDF list emit in `src/ui/preview.rs` always `pdf_loads.remove(&page_index)` before checking the overlay. Recycle reuses `pdf-page-{n}`. If page N unbinds mid-render and binds again, the cancelled helper’s `Failed` drops the new `LoadHandle` and sets “Unable to render this PDF page” — the #815 blank-preview symptom, now on the normal unbind path. Previously `task.abort()` dropped the glib future so that emit did not run.

After `spawn_blocking` returns (and `abort_safe` is restored), return without emit when `cancellation_for_task.is_cancelled()`, matching the existing post-permit check. Optionally still cache a successful PNG. A test that an acquired permit whose token is cancelled does not emit Ready or Failed would lock it. Swallowing only the `"Preview cancelled"` string is not enough: a late real error after unbind has the same `loads.remove` side effect.

Do not fix this by aborting during `spawn_blocking`; that would release the permit while the helper is still running.

l0gicgate:
Merged main at `6e5bdf0f8487d8ed6c6566a88672f7fd321cee91`, preserving its SVG-thumbnail fallback alongside the PDF sizing changes.

Fixed the [cancellation blocker](https://github.com/lgse/strata/pull/898#issuecomment-5647785432): cancelled in-flight renders now discard both late success and late failure, without releasing the PDF permit before the blocking worker exits. Added a deterministic provider regression covering both outcomes. The final comment/doc-comment audit found no added or modified comments requiring removal.

Manual verification on private Xvfb/D-Bus: opened a synthetic 40-page PDF, rapidly scrolled in both directions, and reopened it. Later pages remained rendered and reopening showed page 1; no crash or persistent blank preview was observed. This does not reproduce the original reporter’s exact document or Wayland environment.

Current screenshots were captured locally. GitHub attachment-upload tooling is unavailable in this session, so the screenshots are not attached; local paths are provided to the owner.

l0gicgate:
Follow-up in `8cf42edf`: fixed the first-open PDF size jump reported during owner testing. Cold loads previously displayed a shared thumbnail as a one-page document, then rebuilt the viewer when full pixels and the real page count arrived. PDFs now wait for that complete first-page result; image placeholders and the PDF rendered-page cache are unchanged.

Added a regression with a populated shared thumbnail cache: cold opening and reopening must both deliver the full-size page and the correct document page count. It fails when the old PDF placeholder path is restored.

Manual check on private Xvfb/D-Bus: a synthetic 40-page PDF opens directly at its fitted page size, remains rendered after rapid scrolling, and reopens correctly. A current cold-open recording is captured locally; attachment-upload tooling remains unavailable, so it is not posted here.

PR #899, opened 2026-09-12
fix(delete): play dissolve after progress dismissal

### LINKED ISSUE #895: fix(delete): dissolve animation collides with progress popup on multi-file deletes
## Description

When deleting multiple files, the dissolve animation (560ms) and the progress popup appear at the same time, which looks awkward. The animation should play after the progress popup completes (or be skipped for large batches).

Currently:
- The dissolve animation plays immediately when the user confirms the delete dialog.
- The progress popup appears after `DeletionStarted`, which fires inside the animation's `on_done` callback (so after the animation for large batches ≥16).
- For small batches (<16), the popup's 350ms delay can land mid-animation, causing visual overlap.

## Expected behavior

The dissolve animation should play after the progress popup completes, or be gated to only play for small batches where no progress popup is shown.

## Constraint

Files are removed from the model incrementally during `DeletionProgress`, so by the time the animation plays, fragment positions may be stale. For full-batch deletes this is fine; for partial views, fragments may briefly overlay surviving items.

## How to test

1. Select 20+ files
2. Press Delete
3. Confirm the dialog
4. Observe the dissolve animation and progress popup overlapping

## Related context

Discord discussion (l0gicgate, Gio, fadilasif):
- l0gicgate suggested showing the animation only for smaller batches with no progress
- Files disappear as they get deleted, so delaying row removal is not straightforward

JoeJoeflyn:
## Description

Capture the visible rows selected for permanent deletion before the operation mutates the browser model, then play their dissolve animation only after the deletion progress dialog has fully dismissed. Failed, partial, and cancelled deletions discard the prepared animation.

This prevents the dissolve effect from competing visually with deletion progress while keeping off-screen rows out of the animation.

## Visual evidence

https://github.com/user-attachments/assets/f5e17b6b-89ee-461b-8969-b39356f35fee

## How to test

1. Open a directory containing at least 16 files, select all visible entries, and permanently delete them.
2. Confirm that deletion progress appears without a simultaneous dissolve effect.
3. Confirm that the visible deleted rows dissolve only after the progress dialog finishes closing.
4. Repeat with reduced motion enabled and confirm the dissolve is skipped.

**Expected result:**

Deletion progress and the dissolve animation never overlap. After successful deletion, only rows visible when deletion began dissolve after progress closes; failed or cancelled deletions do not play the dissolve.

## Related issue

Closes #895

### DIFF
--- a/src/app/browser.rs
+++ b/src/app/browser.rs
@@ -185,7 +185,9 @@ pub enum BrowserEvent {
         completed: usize,
         total: usize,
     },
-    DeletionFinished,
+    DeletionFinished {
+        succeeded: bool,
+    },
     RestorationStarted {
         total: usize,
     },
@@ -2005,7 +2007,9 @@ impl Browser {
                 });
             }
             if deleting {
-                browser.emit(BrowserEvent::DeletionFinished);
+                browser.emit(BrowserEvent::DeletionFinished {
+                    succeeded: matches!(&event, OperationEvent::Deleted { .. }),
+                });
             }
             if restoring {
                 browser.emit(BrowserEvent::RestorationFinished);
--- a/src/app/browser/tests.rs
+++ b/src/app/browser/tests.rs
@@ -93,6 +93,12 @@ fn deletion_monitor_changes_publish_once_after_the_terminal_event() {
             .count(),
         1
     );
+    assert!(
+        events
+            .borrow()
+            .iter()
+            .any(|event| matches!(event, BrowserEvent::DeletionFinished { succeeded: true }))
+    );
 }
 
 #[test]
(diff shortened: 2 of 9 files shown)


l0gicgate:
Merged upstream main at 29ac1dd3a71d765ade46495ffaaaa2eb812d4dfe and reviewed the complete resulting diff, including every added comment. Removed two redundant dissolve doc comments; no functional blockers found.

Validation: pinned-container formatting and Clippy passed; Rust suite passed (1,525 tests, 18 explicitly ignored); all 821 canonical E2E tests passed. The progress-dismissal/dissolve regression passed. No separate manual visual verification was performed; review edits do not change UI behavior. All current CI checks passed (optional codesmith skipped). The PR remains unmerged.

PR #904, opened 2026-09-12
feat(media): isolate media workers and ship matching executable bundles

### LINKED ISSUE #850: feat(media): isolate media workers in a helper and support atomic two-binary updates
### Problem or use case

PR #839 (merged main `e598625aeea971473f303cf6ad623d0d29321265`, closing #824) improves preview startup by streaming sandbox-decoded media, but adds direct native GStreamer core/app/base dependencies to the main executable. If a required shared library is missing, the dynamic loader prevents Strata from starting: the application cannot show its existing “Preview unavailable” dependency guidance.

The owner has selected a **separate media helper process** rather than bundling GStreamer or dynamically loading it in the UI process. Ship two executable artifacts: `strata` and `strata-media-helper`. Support multiple instances/workers of the helper for media work (thumbnails, image/RAW/PDF previews, video, audio and animated images), not a single global audio-only daemon.

This requires a coordinated release/installer/updater migration. The current updater installs only `strata`; putting another executable in the archive is insufficient. Aim to ship seamlessly in the next release, but do not release a broken migration to meet that target.

#### 1. Executable and dependency boundary

- Split Cargo targets/crates so the UI binary no longer directly or transitively links GStreamer through Strata's media implementation. A second `[[bin]]` in the same dependency-heavy target graph is not proof of isolation: inspect the final ELF dependency closure.
- Name the helper `strata-media-helper`; provide explicit versioned operation modes for current thumbnail/preview work and raw PCM audio output. Keep the design extensible to multiple worker processes without requiring another executable for every media type.
- Preserve current fast incremental decoding. Move `src/ui/media/audio.rs` GStreamer output out of the UI process, with asynchronous control, bounded transport and playback-clock/error reporting. Keep validated texture presentation in GTK.
- Reuse/refactor existing `src/sandbox.rs`, `src/sandbox/media.rs`, `src/sandbox_helper.rs`, `src/sandbox_helper/media.rs`, `src/media.rs`, and preview/thumbnail service interfaces instead of introducing a parallel media stack.
- Resolve the helper from the running application's trusted installation/version, not arbitrary PATH or cwd. Preserve installed-executable snapshot/rename handling currently used by sandbox helper discovery.
- Explicitly distinguish Strata-added dependencies from the platform toolkit's own dependencies. Some distro GTK builds already depend on GStreamer: promise removal of Strata's additional startup dependency, not that a GTK installation can run with its own required libraries removed.
- Decide and document capability granularity: if the common helper links GStreamer at startup, missing GStreamer will disable **all** of its modes, including thumbnails. Do not promise unaffected image/PDF previews without implementing and testing a dependency-isolated strategy within the chosen two-artifact design.

#### 2. Preserve sandbox boundaries

Current main parses the original only inside bubblewrap/FFmpeg, returning validated owned RGBA frames and fixed-format PCM. GStreamer receives application-constructed PCM caps, not compressed input or decoder-provided pipeline strings.

- Keep original-file parsing sandboxed, single-input/read-only, with no network, desktop/session bus, or audio-server access for parsers. No unsandboxed fallback when dependencies or sandbox startup fail.
- Audio output needs access to the host audio service. Use a distinct helper process/role with narrowly scoped capabilities that consumes validated PCM only; **do not grant audio/desktop sockets to the parser sandbox** just because both roles use one executable.
- Version the protocol and validate operation, job/generation identity, framing, dimensions/strides, payload sizes, timestamps, duration and message ordering on both sides of each boundary. No arbitrary command/pipeline execution supplied through IPC.
- Retain limits, backpressure and cancellation across extra process hops. Never block the GTK thread on a pipe, process wait or shutdown join. Child crash, timeout, malformed output and EPIPE must fail the job, not the application.
- Threat-model binary discovery, writable paths, inherited descriptors/environment, subprocess privileges and IPC. An independent security review is warranted for the changed trust boundaries.

#### 3. Multiple workers and lifecycle

- Support several concurrent helper instances with explicit job ownership and bounded scheduling for thumbnails, video and audio. Avoid a single process becoming a global bottleneck or crash domain.
- Preserve the current four-session-per-application policy across windows/chooser, fifth-player busy feedback, bounded queues, 30-second preview interval, 30 fps, 1280-pixel axis limit, GIF behavior, and 30-second paused resource cleanup with retained frame/position unless a policy change is explicitly approved.
- Model a playback session separately from its process count: a session can own parser/FFmpeg/audio children; do not accidentally consume multiple session slots for one preview. Define bounded thumbnail concurrency, fairness, cancellation priority and total resource budgets rather than an unbounded worker per request.
- Stop descendants and reap children on seek, selection change, close, timeout, EOF or window/application destruction; reject stale responses. No leftover helpers after jobs finish or orphaning if the UI exits unexpectedly.
- Keep fresh parser sandbox/decoder generations across untrusted inputs. Process reuse is not automatically authorized by worker pooling. Coordinate with #841 (parent-side coordinator reuse/RSS investigation) and #516 (thumbnail helper reuse); do not conflate their scopes or silently weaken per-decode isolation.
- Preserve startup/live mute and volume preferences across windows and rebuilds. Define behavior when audio becomes unavailable: actionable unavailable state or explicitly supported muted/video-only degradation, never silent failure or a false loading state.

#### 4. Two-binary packaging and atomic updates

- Produce `strata` and `strata-media-helper` for x86_64 and aarch64 in the same release archive, with matching version/protocol metadata, executable permissions, checksums/attestation, licenses and separate useful debug symbols.
- Update `.github/workflows/release.yml`, `install.sh`, `packaging/aur/PKGBUILD.in`, generated stable/RC PKGBUILD and SRCINFO files, installation documentation, and relevant container/test inputs together. Audit all supported installation channels, including Omarchy-managed installs, not only the standalone installer.
- Audit helper runtime packages, particularly `autoaudiosink`/actual audio output backends. Main currently uses `appsrc ! audioconvert ! audioresample ! volume ! autoaudiosink`; AUR lists `gst-plugins-good` as optional while the installer includes it. Define required-for-launch versus required-for-capability packages accurately. Do not assume another Omarchy package supplies them.
- Replace the single-file updater operation in `src/services/update_install.rs` with a transactional bundle install. Existing `find_binaries()` already accepts multiple names but currently receives only `strata`; finding two files is not an atomic installer.
- Suggested standalone layout: versioned directories containing both binaries and associated metadata, plus an atomically switched `current` pointer/launcher. An equivalent atomic layout is acceptable if it meets the same requirements.
- Download/verify and fully stage the matching pair on the destination filesystem before activation. Validate archive paths, symlinks, duplicates, architecture, required members and manifest/protocol compatibility. Do not execute downloaded binaries merely to discover their version or validate an archive.
- Missing optional runtime libraries must not make helper installation fail; installation integrity and runtime capability are separate checks.
- Preserve the old bundle until activation succeeds. On interruption, checksum failure, disk-full, permissions error, concurrent update or missing/corrupt helper, leave a working old version rather than a mixed pair. Serialize competing update/rollback entry points across processes as needed.
- Rollback restores the **whole matching bundle**, including associated integration metadata where applicable. Preserve desktop entries, icons, portal/file-chooser and D-Bus launch paths, symlink resolution, custom installation locations and uninstall behavior.
- Old running application instances must keep locating their own matching helper after `current` switches. Do not garbage-collect an in-use version directory; define a safe bounded retention/cleanup policy.
- Continue respecting package-managed install ownership: AUR/Omarchy/system-managed updates stay with their package manager; do not replace package-owned files through the standalone updater.

#### 5. Migration from already-published single-binary releases — release gate

Older updaters extract and replace only `strata`, discarding any helper in the downloaded archive. Updating those installations directly to the new bundle format would leave them without the helper.

- Design and test the legacy-to-bundle migration before publication. Choose an explicit compatibility mechanism: a verified first-run installer/recovery flow, update-feed routing to a compatible transition, or another approach preserving trust and rollback.
- A transition release alone is **not sufficient** if an older client can skip it and download the newest release directly. Cover direct upgrades from all supported published versions as well as users who installed the transition.
- Do not fetch/execute an arbitrary “latest helper” at runtime. Any helper acquisition must match the running release and go through authenticated/checksummed release installation, with an offline/retry/error path.
- The main application must remain usable if helper installation is missing/incomplete, and offer clear repair/reinstall instructions rather than crash. Do not label a missing bundled helper as merely a missing GStreamer package.
- Define rollout ordering for stable and RC channels, old release links, package metadata and release notes. Prefer a prerelease exercising the real installed upgrade path before stable promotion.

#### 6. Capability reporting and UX

- Distinguish: helper absent, non-executable/wrong architecture, protocol mismatch, missing GStreamer shared libraries, missing plugins/audio backend, missing FFmpeg/thumbnail tools, unavailable sandbox, parser failure, and worker saturation.
- Keep “Preview unavailable” and provide accurate distro-aware package/repair guidance. Capture bounded/sanitized helper startup stderr; loader failures happen before a protocol handshake.
- File browsing and unrelated UI remain usable. Allow a capability recheck after installation or explain when restart is needed. Avoid exposing private file contents/paths in diagnostics or blindly recommending package commands from untrusted stderr.

#### Acceptance and release validation

- [ ] Final main ELF inspection shows no Strata-added GStreamer startup linkage; helper dependency closure and toolkit-provided dependencies are documented.
- [ ] Main launches on a supported minimal installation with media-only dependencies absent; exercising affected features yields actionable errors. Restore dependencies and verify recovery.
- [ ] Missing plugins versus missing shared libraries, helper missing/corrupt/mismatched, no FFmpeg, no audio server and no sandbox are separately tested. No unsandboxed fallback.
- [ ] One job/helper crash does not crash the UI or unrelated workers. Test malformed/truncated IPC, stale responses, cancellation, process trees and parent death.
- [ ] Multiple windows/chooser and mixed thumbnail/audio/video workloads retain limits, controls, preferences, fairness and no leaked descendants/FDs/threads. Include 100-cycle ownership/RSS/PSS and startup/seek comparisons without claiming a universal memory plateau.
- [ ] Existing media/protocol/sandbox/preferences regressions plus full canonical pinned Rust and E2E validation pass. GUI automation uses private Xvfb/D-Bus; hardware/speaker tests use explicit operator consent and separate recorded evidence.
- [ ] Real installed x86_64/aarch64 bundles exercise install, launch, helper discovery, direct legacy upgrade, interrupted/concurrent update, restart, running-old-instance behavior, rollback, custom paths and package-managed ownership. Test release artifacts, not just development binaries.
- [ ] Verify actual audio on supported audio stacks, video-only/audio-only/A/V/GIF playback, seeking, EOS, failure paths and 31-second pause/resume. #839 used fakesink for unit tests; real audio and ARM64 were not established by that evidence.
- [ ] Confirm Ubuntu compatibility, including the unresolved alternatives-library sandbox problem in #806; do not claim the helper split fixes it automatically or widen mounts without review.
- [ ] Release notes disclose new layout/capabilities, dependency guidance, migration/recovery/rollback, remaining platform limitations and what is not bundled. Required CI checks pass before merge/release.

### Alternatives considered

- Keep GStreamer mandatory in the main executable: simpler packaging, but a missing library prevents displaying dependency guidance, and old binary-only updates cannot install OS packages.
- Bundle a private GStreamer runtime: larger per-architecture dependency/plugin and licensing/security-maintenance burden; still requires coordinated bundle updates. Not the selected approach for this issue.
- Dynamically load GStreamer in the UI: avoids direct startup linkage but adds loader/API complexity and keeps audio failures in the UI process. Not selected.
- One singleton media daemon: unnecessary shared failure/bottleneck and isolation concerns. Prefer bounded separately supervised workers using one versioned helper executable.

### Visual examples

N/A — preserve the existing preview-unavailable presentation, refining dependency and installation-repair guidance as needed.

### Strata version

Main `e598625aeea971473f303cf6ad623d0d29321265` (merged PR #839), inspected when filing. Recheck installation/update code against current main before implementation.

### Additional context

- #839 / #824: incremental sandboxed decoding and its evidence/limits (`docs/evidence/824/`, `docs/preview-sandbox.md`).
- #841 and #516: related but distinct worker-reuse work; coordinate rather than duplicate.
- #806: unresolved Ubuntu sandbox dependency resolution. #773: historical hardware/cover-art failure; do not infer that this architecture validates all hardware paths.
- #779 / #830 and [GTK MR !10367](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/10367): legacy GtkMediaFile/GtkGstSink leak work. Main's new player bypasses GtkGstSink/GstPlay for playback, using GStreamer only for PCM output. This issue does not bundle a private runtime, retire the remaining GstPlay patch, or claim all upstream memory/crash problems are fixed.
- Main's current installer/AUR already declare GStreamer core/base dependencies; the release archive still contains only the main executable. Avoid treating the existence of the legacy `--preview-helper` mode (the same executable re-executed) as a separate dependency-isolated binary.
- This ticket records the owner-selected direction; no production code or system configuration changed while filing it.

l0gicgate:
## Description

Move native media parsing and PCM playback into short-lived, role-specific `strata-media-helper` processes, with versioned bounded IPC, cancellation, trusted helper discovery and actionable capability failures. The browser and file chooser keep their shared preview path.

Ship matching UI/helper bundles for x86_64 and AArch64, including finalized-helper embedding for offline recovery after published single-binary updaters discard the sidecar. Installers and the in-app updater validate the complete archive without executing downloaded binaries, activate modern bundles transactionally, preserve running instances and support verified rollback. Audited legacy releases run from a separate flat launcher so their old updaters cannot overwrite cached rollback archives. Release production and package metadata cover both executables. Per owner direction, Release is build/package/publish only: no test, E2E, Quality/Lint or runtime smoke jobs; normal PR CI stays separate.

This does not ship the experimental private media runtime or claim to resolve every GTK/driver memory or VRAM leak. System GTK and its mandatory GStreamer libraries remain host requirements on Arch/Omarchy. The optional artifact diagnostics are manual only. The added release-test job was removed and its remaining ARM64 runtime validation cancelled at the owner's request; that runtime coverage is not claimed. Portal/D-Bus activation keeps the stable launcher while validating its canonical target, and update/startup refresh repairs older version-specific service entries. Upstream PDF rendering limits are retained inside the isolated helper. Scope and reproducible evidence: [#850 evidence](https://github.com/lgse/strata/blob/feat/850-isolated-media-workers/docs/evidence/850/README.md).

## Visual evidence

N/A — internal process, installation and release changes; the existing preview layout is unchanged. Generated capability-error/playback screenshots are retained with the native release-gate evidence.

## How to test

Use locally built matching release bundles in a disposable installation prefix; see [bundle installation and recovery](https://github.com/lgse/strata/blob/feat/850-isolated-media-workers/docs/media-helper-bundles.md).

1. Preview video, audio, A/V, GIF, image and PDF files from the browser and Strata file chooser. Seek, toggle mute/volume, close/reopen, and resume after a pause longer than 30 seconds.
2. Upgrade a standalone installation while an old window remains open, then restart through its normal launcher. Roll back to an audited older release, upgrade using that old client's updater, return to a complete modern bundle, then repeat the rollback. Repeat with an installation updated by a legacy single-binary updater; previewing should recover the exact matching helper offline.
3. In a disposable container only, remove a media-only dependency that GTK itself does not require. Confirm the UI starts, reports repair guidance and still previews text; restore the dependency and retry.
4. Enable the Strata chooser in the disposable installation, activate another bundle, then start a fresh chooser request. The user D-Bus service must keep `Exec=<bin>/strata --portal`, not a path under `versions/`.

**Expected result:** previews use matching isolated workers, workers are released when no longer needed, failed staging preserves the working installation, old windows retain their matching helper, and restart/portal launch paths follow the activated bundle.

## Related issue

Closes #850


l0gicgate:
Fixed the Arch packaging bootstrap in `5d38c090` and `fb89205f`:
- Install Git **before** checkout, avoiding the REST-archive fallback without commit metadata.
- Run Python tests as the existing `builder` user **after** transferring checkout ownership. Their private `HOME` no longer needs checkout's `safe.directory` exception. Git ownership checks remain enabled.

Validation: reproduced the exact `detected dubious ownership` failure with the selected local-install test in a disposable rootless Podman `archlinux:base-devel` container. Applied the workflow's user/ownership order and ran `su builder -c "python3 -m unittest discover -s scripts -p 'test_*.py'"`: **203 tests passed**. `git diff --check` passed. Evidence: `packaging-ci-ownership-fix.log` in the task-local scratch.

The fixes change only packaging-job bootstrap, not release workflow or test implementation. Rust/GUI/lint suites were not rerun locally. Follow-up `8c5e9b93` incorporates current `main` to unblock PR CI and retains the upstream SVG thumbnail fallback in the relocated parser; Python tests are unchanged. Remote [Packaging CI run 34714993194](https://github.com/lgse/strata/actions/runs/34714993194) passed on `fb89205f`, including Python tests, both package builds, namcap, installed-path checks and executable provenance.

wmfeht:
**Request changes** (private review of `fb89205f`; one blocker).

Portal/D-Bus activation does not follow bundle `current` after an in-app update.

`install()` / `refresh_stale_portal()` now call `installation::launch_path` (the stable `<bin>/strata` → `.strata-bundles/current/strata` launcher). `install_at` then runs that path through `secure_executable`, which `canonicalize`s the symlink chain to `versions/<archive-sha>/strata` and writes **that inode** into `org.freedesktop.impl.portal.desktop.strata.service`.

After a later transactional activate:
- Desktop entries are refreshed with `exe_dir.join("strata")` and follow `current`.
- The portal service file still names the old version directory.
- `refresh_after_in_place_update` only restarts xdg-desktop-portal; it does not rewrite Exec, so the next activation is still the old pair.
- `refresh_stale_portal_at` looks for `Exec={launcher} --portal` and will not match a canonical versioned path, so it cannot repair this.

Running old UI windows correctly keep their matching helper via pinned inode. The *next* portal start is not a running window; #850 and `docs/media-helper-bundles.md` require portal/D-Bus launch paths to use the activated launcher.

Fix: keep the ownership/permission walk on the canonical target, but persist `launch_path`'s launcher path in the service file, and rewrite that Exec on in-app update the same way desktop entries are refreshed. Add a test that a versioned `current_exe` plus `launch_path` yields `Exec=<bin>/strata --portal`, and that after `current` switches a portal refresh still points at the launcher.

l0gicgate:
@wmfeht Addressed the portal blocker in `0d51d03e`.

- The canonical target still passes the existing ownership/permission walk; the service stores the verified stable bundle launcher, never an arbitrary replaceable alias.
- In-app updates and startup repair old version-specific `Exec` entries before restarting the portal, preserving chooser preferences and other service fields. Startup also recognizes an old version-specific process after a failed restart, so repair remains retryable.
- Regressions cover install → update → rollback, both repair routes, failed-restart retry, another build's activation, and replaceable aliases.

Conflicts with `main` are resolved through `a018d0b4`. Upstream PDF bounds and native rendering tests remain in the helper/shared protocol; the old in-process parser is not reintroduced. The later CLI-version and installer changes are retained.

Validation: full pinned Rust suite **1,578 passed** and canonical E2E **850 passed** on the reconciled PDF/portal baseline. After the retry refinement and later CLI merge, reran all **21 portal tests**, **2 launch-mode tests**, **1 version test**, **208 Python tests**, pinned formatting/Clippy, and `git diff --check`; all passed. No release-artifact gate or desktop-session testing was run. Normal PR CI remains required before merge.

<details>
<summary>Local commands and scope</summary>

Task-owned rootless Podman wrapper on `PATH`, `STRATA_CONTAINER_ENGINE=podman`:

- `./scripts/quality.sh test`
- `./scripts/e2e.sh`
- `./scripts/quality.sh fmt`
- `./scripts/quality.sh clippy`

In the private-display/D-Bus pinned container wrapper:

- `/opt/e2e-venv/bin/python scripts/test-headless.py portal_setup::tests`
- `/opt/e2e-venv/bin/python scripts/test-headless.py tests::launch_mode`
- `/opt/e2e-venv/bin/python scripts/test-headless.py tests::version_line`

Also: `python3 -m unittest discover -s scripts -p 'test_*.py'` and `git diff --check`.

Full checks were selected for the parser/protocol/harness reconciliation. Only affected checks were repeated for the later portal-retry/CLI changes. The failed-restart regression was confirmed failing before the argv-to-launcher fix, then passing. Existing ignored tests, including archive-dependent release diagnostics, were not claimed as passed.
</details>

l0gicgate:
Resolved the conflicts in `2a54ccdc` and pushed through `7dc78722`. GitHub now reports **MERGEABLE**. The newer word-wrap commit on main is also conflict-free; the branch is behind that nonconflicting commit, not blocked by conflicts.

Upstream preview sizing/session behavior and full-length playback are preserved, with parsing and PCM output still isolated in the helper. Long-timeline transport and clock bounds now agree across both processes.

The initial full Rust run passed 1,560 tests but exposed a recovery-lock reuse race. `7dc78722` explicitly releases the lock even while a fork retains a descriptor duplicate, with a regression test. All 7 affected recovery tests and all 25 helper/9 protocol tests subsequently passed; the otherwise-passing UI suite was not replayed after that narrow correction. Canonical E2E passed **859 tests**; Python passed **208**; final formatting and Clippy passed.

Normal PR CI and review remain pending. Release behavior is unchanged; no release diagnostics or publication were run.

l0gicgate:
Closing this, as we don't need it right now.

PR #909, opened 2026-09-12
fix(ui): reserve space for modal dialog shadows

### LINKED ISSUE #906: fix(ui): Prevent dialog shadows from being clipped
### Strata version

Not confirmed for the screenshot; current repository and latest release are 0.16.0.

### Installation method

Other: locally installed executable at ~/.local/bin/strata; installation source not confirmed.

Arch Linux, Hyprland (reporting machine).

### What happened?

The outer shadow on the item Properties dialog appears clipped at the dialog bounds rather than extending smoothly into the dimmed background. This gives the dialog a hard rectangular edge instead of the intended soft elevation.

### Steps to reproduce

1. Open Strata and select a file or folder.
2. Open its Properties dialog (Alt+Enter).
3. Inspect the outer edges and corners against the dimmed browser background.

### Expected behavior

The shadow should extend outside the rounded dialog surface and fade smoothly into the backdrop on all sides, without being cut off by the dialog or its parent container.

### Screenshots or video

Two screenshots were supplied with the report; attachment is pending:
- Expected: the Pictures folder Properties dialog has a soft shadow extending beyond its rounded border.
- Actual: the autostart.sh Properties dialog has an abrupt outer edge with the shadow apparently clipped, especially around the bottom edge and corners.

### Relevant logs

No logs supplied; this is a visual rendering defect, not a reported crash.

### Additional context

The reference and actual screenshots use different themes and different item types. The reported defect is the shadow clipping, not the color or content differences. Root cause and whether other dialogs are affected remain unverified.

l0gicgate:
## Description

Reserve transparent space inside the shared modal scroller so dialog shadows can extend beyond their rounded surfaces without being clipped. Keep shadow-space clicks dismissing the dialog, and constrain hit testing to the visible viewport.

Implemented in the existing checkout at the owner’s request. The owner requested opening the PR without an added test or further local validation; Rust/E2E suites, formatting/Clippy gates, and visual capture were not completed. `git diff --check` passed. Rendering remains unverified locally.

## Visual evidence
image

## How to test

1. Select a file or folder and open Properties with Alt+Enter. Inspect all edges and corners; repeat with Settings and another action dialog.
2. Click inside the dialog, then in the surrounding shadow. Also resize the window until the dialog needs scrolling and verify its controls remain reachable.

**Expected result:** Shadows fade outside rounded surfaces without a hard cutoff. Inside clicks preserve the dialog; backdrop/shadow clicks dismiss it. Oversized dialogs remain scrollable.

## Related issue

Closes #906

### DIFF
--- a/src/ui/modal.rs
+++ b/src/ui/modal.rs
@@ -62,7 +62,8 @@ pub(super) fn modal_layer(
 
     let click = gtk::GestureClick::new();
     let weak_layer = layer.downgrade();
-    let weak_content = viewport.downgrade();
+    let weak_viewport = viewport.downgrade();
+    let weak_content = content.as_ref().downgrade();
     let overlay = overlay.clone();
     let root = root.clone();
     let block = block_dismiss.clone();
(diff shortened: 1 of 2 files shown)


l0gicgate:
Updated the modal scroller to fill the window while centering the shadow-padded dialog inside it. Vertical overflow now places the scrollbar at the right window edge rather than beside the dialog. This also applies to search, which shares the layout. Please verify centered short dialogs and scrollable tall dialogs when reviewing. Continued the requested no-test/local-validation handoff; `git diff --check` passed, but rendering has not been verified locally.

PR #913, opened 2026-09-12
fix(trash): refuse restore through a same-volume symlink parent

### LINKED ISSUE #862: fix(trash): refuse restore through a same-volume symlink parent
### Strata version

### Installation method

Built from source

### What happened?

Trash Restore follows a same-volume symlink parent into the symlink target. After trashing a nested file and replacing its parent directory with a symlink to another folder on the same volume, Restore offered and wrote the file inside the symlink target. The unit test only covers a symlink parent that leaves the volume.

### Steps to reproduce

1. Trash `documents/nested/inner.txt`.
2. Replace `nested` with a symlink to `pictures/` on the same volume.
3. Restore the trashed file from the confirmation that shows `/…/pictures/inner.txt`.
4. Observe `inner.txt` leaves Trash and appears as `pictures/inner.txt`.

### Expected behavior

Lookup should fail (parent is no longer a real directory / escaped the original location), or Restore should refuse to follow the symlink. Confirmation of the original location should not offer the symlink target.

### Screenshots or video

### Relevant logs

### Additional context

Found in the v0.17.0-nightly.20260912 QA sweep. `symlink_parent_that_leaves_the_volume_is_rejected` only covers leaving the volume. `docs/trash-restore-testing.md` requires confirmation of the original location and rejects bind-mount/subvolume escapes. Distinct from #735 (home-trash restore to another filesystem) and #736 (destinations whose parents do not exist).

pdwaldrop:
## Description

Trash Restore resolves a destination by canonicalizing the closest existing ancestor of the original path, which silently follows a symlink if that ancestor has been replaced by one. The existing volume-identity and mount-point checks only catch an escape that *leaves* the volume (`symlink_parent_that_leaves_the_volume_is_rejected`), so a same-volume symlink parent -- e.g. a directory swapped for a symlink into a sibling folder on the same disk -- passed through untouched and wrote the restored file into the symlink's target instead of its real original location.

`canonical_restore_destination` now checks the destination's immediate parent with `symlink_metadata` (which does not follow the final component) before any canonicalization happens, and refuses outright if it's a symlink rather than a real directory.

## Visual evidence

N/A -- this is a backend safety check in the trash-restore path with no UI change.

## How to test

1. Trash `documents/nested/inner.txt`.
2. Replace `nested` with a symlink to another folder on the *same* volume (e.g. `pictures/`).
3. Try to restore the trashed file.
4. Before this fix, the restore silently succeeds and writes the file into `pictures/inner.txt`. After, it's refused with "The original location's parent is a symlink and cannot be restored."

Added `symlink_parent_that_stays_on_the_volume_is_still_rejected` alongside the existing leaves-the-volume test, confirmed it fails against the pre-fix code and passes after.

## Related issue

Closes #862

### DIFF
--- a/src/adapters/trash_restore.rs
+++ b/src/adapters/trash_restore.rs
@@ -641,6 +641,9 @@ fn canonical_restore_destination(path: &Path) -> Result<PathBuf, RestoreTargetEr
     let parent = path
         .parent()
         .ok_or_else(|| RestoreTargetError::new("The original location is invalid"))?;
+    if std::fs::symlink_metadata(parent).is_ok_and(|metadata| metadata.file_type().is_symlink()) {
+        return Err(symlinked_parent_error());
+    }
     let mut existing = parent.to_path_buf();
     let mut missing = Vec::new();
     while !existing.as_os_str().is_empty() && !existing.exists() {
@@ -678,4 +681,12 @@ fn escaped_restore_error() -> RestoreTargetError {
     RestoreTargetError::new(
         "The original location is outside the trash volume and cannot be restored.",
     )
+}
+
+/// The item's own parent no longer names the directory it was deleted from --
+/// something replaced it with a symlink, possibly to redirect the restore
+/// elsewhere on the same volume where the plain volume/mount checks below
+/// would not otherwise notice.
+fn symlinked_parent_error() -> RestoreTargetError {
+    RestoreTargetError::new("The original location's parent is a symlink and cannot be restored.")
 }
--- a/src/adapters/trash_restore/tests.rs
+++ b/src/adapters/trash_restore/tests.rs
@@ -279,6 +279,33 @@ fn symlink_parent_that_leaves_the_volume_is_rejected() -> std::io::Result<()> {
 }
 
 #[test]
+fn symlink_parent_that_stays_on_the_volume_is_still_rejected() {
+    let fixture = tempfile::tempdir().expect("fixture");
+    let uid = 1000;
+    let trash = volume_trash(fixture.path(), uid);
+    let source = trash.join("files/inner.txt");
+    fs::write(&source, b"secret").expect("source");
+    fs::create_dir_all(fixture.path().join("documents")).expect("documents dir");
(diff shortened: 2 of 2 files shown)


l0gicgate:
Merge preparation is blocked locally; nothing has been pushed. Merged verified upstream main `87d67bd4d97210222935ece0b57ea0f27084e580` without conflicts and removed the redundant doc comment on the symlink error helper. No other PR feedback was present. All 31 trash-restore tests pass; native and pinned formatting/Clippy pass. The full isolated Rust run finished with 1,654 passed, 1 failed, 19 ignored: `ui::browser::tests::focus::filter_dismisses_only_when_focus_leaves_its_own_column` times out at `src/ui/browser/tests/focus.rs:391`. That exact failure reproduces both natively and in the pinned container. No GUI behavior was changed or manually verified. Holding the push because the required local suite is not green.

l0gicgate:
Resolved the earlier local validation blocker and pushed `b45f76bd`. Latest upstream main `1d0f3298cc5953643b5db901156593de73df0af5` was merged first, without conflicts.

The focus test used a query matching no rows, then accepted focus on a row detached by the asynchronous filter update. It now retains matching results, waits for them, and verifies focus actually reaches each column list. Its original dismissal/query-preservation assertions remain; a temporary negative control confirmed it fails when dismissal is disabled. No application/UI behavior was changed during this repair, so no manual desktop exercise or screenshots were needed. The redundant restore-error doc comment was also removed; no new code comments remain.

Full pinned validation passed (`STRATA_CONTAINER_ENGINE=<isolated-podman-wrapper> STRATA_REQUIRE_DEVICE_TESTS=1 ./scripts/quality.sh all`): formatting, Clippy, and 1,656 Rust tests passed, with 19 intentionally ignored. Native formatting/Clippy and the repaired exact test also pass. Full local E2E was not rerun for this bounded backend/test-only change; GitHub’s full E2E suite passed. All four required GitHub checks are now green. No outstanding review-thread concerns were present. Reviewer approval is still required; this PR has not been merged.

PR #915, opened 2026-09-12
fix(trash): move selected filter results to trash with Delete

### LINKED ISSUE #855: fix(trash): move selected filter results to trash with Delete
### Strata version

0.16.0 (`964d262`)

### Installation method

Built from source

### What happened?

Delete does not Move to Trash from a selected Ctrl+F filter result. Keyboard dispatch returns `Proceed` when `selected_search_results()` is some, before trash handling. The file stays in place. Distinct from #720 (refresh the filter after a successful delete from the result menu).

### Steps to reproduce

1. In Columns, press Ctrl+F and type a unique filename.
2. Click the matching result row so it is selected and focused (filter field is not focused).
3. Press Delete.
4. Observe the file remains. Same if Delete is pressed while the query field still has focus.

### Expected behavior

Delete should Move to Trash for the selected filter result, matching the footer `Del  Trash` hint for selected items.

### Screenshots or video

### Relevant logs

### Additional context

Found in the v0.16.0 QA sweep. `src/ui/window/keyboard.rs` returns `Propagation::Proceed` when `selected_search_results()` is some, before `dismissal()` handles Delete.

pdwaldrop:
## Description

`Dispatcher::handle_key` has an early fallthrough that returns `Propagation::Proceed` for any key whenever a search/filter result is selected and no text field has focus, so that search rows keep ownership of their own navigation instead of directory commands acting on a hidden (filtered-out) selection. That fallthrough runs before `dismissal`, which is what handles Delete -> Move to Trash, so it swallowed Delete too: pressing Delete on a selected Ctrl+F filter result left the file in place.

Excluded `Key::Delete` from that fallthrough so it still reaches `dismissal`, which already resolves the current selection correctly regardless of whether it came from a filter result or a plain directory listing.

## Visual evidence

N/A -- this is a keyboard-dispatch fix with no visual change.

## How to test

1. In Columns mode, press Ctrl+F and type a query matching exactly one file.
2. Click the matching result row so it's selected (filter field not focused).
3. Press Delete. Before this fix, the file remains. After, it's moved to Trash, matching the footer's "Del  Trash" hint.

Added `delete_trashes_a_selected_filter_result`, confirmed it fails against the pre-fix code and passes after.

## Related issue

Closes #855

### DIFF
--- a/src/ui/window/keyboard.rs
+++ b/src/ui/window/keyboard.rs
@@ -132,8 +132,12 @@ impl Dispatcher {
             .or_else(|| self.sidebar_commands(browser, &event))
             .or_else(|| self.context_menu_command(&event))
             .or_else(|| {
-                // Search rows own navigation; directory commands must not act on hidden selections.
-                (self.view.selected_search_results().is_some() && !event.text_has_focus())
+                // Search rows own navigation; directory commands must not act on hidden
+                // selections. Delete still falls through to `dismissal` so trashing a
+                // selected filter result works, matching the footer's "Del  Trash" hint.
+                (self.view.selected_search_results().is_some()
+                    && !event.text_has_focus()
+                    && event.key != Key::Delete)
                     .then_some(Propagation::Proceed)
             })
             .or_else(|| self.text_input(&event))
--- a/src/ui/window/tests/keyboard_dispatch.rs
+++ b/src/ui/window/tests/keyboard_dispatch.rs
@@ -146,6 +146,33 @@ fn text_view_in(widget: &gtk::Widget) -> Option<gtk::TextView> {
     None
 }
 
+fn widget_with_class(widget: &gtk::Widget, class: &str) -> Option<gtk::Widget> {
+    if widget.has_css_class(class) {
+        return Some(widget.clone());
+    }
+    let mut child = widget.first_child();
+    while let Some(widget) = child {
+        if let Some(found) = widget_with_class(&widget, class) {
+            return Some(found);
+        }
+        child = widget.next_sibling();
+    }
+    None
+}
+
+fn press_on(widget: &gtk::Widget, key: Key, modifiers: ModifierType) -> bool {
+    let controllers = widget.observe_controllers();
+    let controller = (0..controllers.n_items())
+        .filter_map(|index| {
+            controllers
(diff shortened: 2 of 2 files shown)


pdwaldrop:
Scoping note: the issue also mentions the file remains "if Delete is pressed while the query field still has focus." I left that variant as-is -- `dismissal`'s own Delete handler already refuses to act while any text field (including the filter entry) has focus, which looks like an intentional guard against Delete-to-remove-a-character in the query also trashing files. This PR only fixes the case the code pointer identifies: Delete reaching `dismissal` at all when a filter result is selected and the field itself isn't focused. Happy to revisit if that guard should be loosened too.

l0gicgate:
Review follow-up pushed as 4d8b762e: fixed a wrong-target deletion risk. The original regression filtered the same file as the hidden directory selection; filtering a different file exposed that `confirm_delete` still used that hidden selection. It now resolves selected search results explicitly, with survivor assertions and Columns/List/Icons E2E coverage. Updated rename tests that incorrectly expected Delete to do nothing, and removed redundant comment prose.

The author’s focus-guard scope note is correct: Delete must remain text editing while an entry has focus. No outstanding review threads were present. The fetched main SHA dd8727708694ef78102214c3f9868e692ff838f7 was already the second parent of the existing merge commit; no redundant merge was needed. Main advanced during validation, so this is not an attestation against subsequent commits.

Pinned rootless-Podman validation: `./scripts/quality.sh all` passed (fmt, Clippy, 1,596 Rust tests; 20 existing ignored), and `STRATA_CONTAINER_ENGINE=podman STRATA_E2E_WORKERS=4 ./scripts/e2e.sh` passed all 874 cases. Private Xvfb/D-Bus only. Captured and inspected the selected nested result before/after Delete; other matches survive. No desktop manual interaction was used.

Visual evidence is pending owner upload: no authenticated attachment-upload tool is available here. Local files are under `/home/l0gicgate/Documents/github/strata-worktrees/855-review-915/target/review-runtime/`: `delete-before-Columns.png` (nested result selected), `delete-after-Columns.png` (nested result removed, other matches remain), with matching `List` and `Icons` pairs. Please upload these through the GitHub editor; they are not attached to this comment.

PR #932, opened 2026-09-13
fix(desktop): reveal hidden files requested through ShowItems

### LINKED ISSUE #858: fix(desktop): ShowItems should reveal hidden files
### Strata version

0.16.0 (`964d262`)

### Installation method

Built from source

### What happened?

`org.freedesktop.FileManager1` `ShowItems` of a hidden file opens the parent and returns success but does not select or show the file when Hidden files is off. README promises those items are selected. A browser “Open file location” on a dotfile therefore does not reveal it.

### Steps to reproduce

1. With Hidden files off (default), call `ShowItems` on `file://…/.secret.txt` (or another dotfile in an otherwise visible directory).
2. Observe the parent directory opens and the call succeeds.
3. Observe the listing still hides the file and the selection is empty.

### Expected behavior

ShowItems should select the named items. For a hidden file that means showing it (temporarily enabling hidden files, or revealing that row) so the caller’s “show in folder” actually shows the file.

### Screenshots or video

### Relevant logs

### Additional context

Found in the v0.16.0 QA sweep. Contrast: ShowItems of a visible file selects it. Not a reopen of #290 (implement FileManager1) or #722 (CLI file arguments).

pdwaldrop:
## Description

Revealing a specific entry by name or location (`select_entries_by_name_at` / `select_entries_by_location_at`) finds the matching entry's position in the column's raw entry list and applies it through the UI's raw-position-to-view-position translation (`ColumnView`'s `ViewMap`). A hidden entry has no view position while hidden files are off -- the filtered view simply doesn't contain it -- so the translation silently drops it and nothing gets selected. `ShowItems` on a hidden file therefore opened its parent directory but left the selection empty, contrary to the README's promise that named items are selected.

`select_entries_matching_at` now checks whether any matching entry is hidden and, if hidden files are currently off, turns them on first (via the same path the Toggle Hidden action uses) before resolving positions -- so the row actually exists in the view by the time selection is applied.

This is shared by `ShowItems` and the CLI file-argument reveal path (`open_argument.rs`), which route through the same selection helper, so both get the fix.

## Visual evidence

N/A -- this is desktop-integration/selection logic with no visual design change (the visible effect is hidden files becoming shown, which is the existing Toggle Hidden behavior, not new UI).

## How to test

1. With Hidden files off (default), in a directory with a dotfile and a visible file, call `org.freedesktop.FileManager1.ShowItems` on the dotfile's URI (e.g. via `gdbus call --session --dest org.freedesktop.FileManager1 --object-path /org/freedesktop/FileManager1 --method org.freedesktop.FileManager1.ShowItems "['file:///path/to/.secret.txt']" ""`).
2. Before this fix, the parent opens but nothing is selected and the dotfile stays hidden. After, hidden files turn on and the dotfile is selected.

Added `selecting_a_hidden_entry_by_name_shows_it`, confirmed it fails against the pre-fix code and passes after.

## Related issue

Closes #858

### DIFF
--- a/src/app/browser.rs
+++ b/src/app/browser.rs
@@ -3303,6 +3303,23 @@ impl Browser {
         depth: usize,
         matches: impl Fn(&FileEntry) -> bool,
     ) -> bool {
+        // A caller selecting a specific entry (e.g. FileManager1's ShowItems) means
+        // to reveal it; a hidden match must actually become visible, not just
+        // "selected" in a list the current filter still hides it from.
+        let has_hidden_match = self
+            .state
+            .borrow()
+            .columns
+            .get(depth)
+            .is_some_and(|column| {
+                column
+                    .entries
+                    .iter()
+                    .any(|entry| matches(entry) && entry.is_hidden)
+            });
+        if has_hidden_match && !self.preferences.get().show_hidden {
+            self.toggle_hidden();
+        }
         let state = self.state.borrow();
         let Some(column) = state.columns.get(depth) else {
             return false;
--- a/src/app/browser/tests.rs
+++ b/src/app/browser/tests.rs
@@ -1955,6 +1955,22 @@ fn selecting_entries_by_name_preserves_the_full_matching_selection() {
 }
 
 #[test]
+fn selecting_a_hidden_entry_by_name_shows_it() {
+    let source = ScriptedSource::scripted(vec!["visible.txt", ".secret.txt"], Vec::new());
+    let browser = Browser::new(Rc::new(source));
+    browser.navigate(Location::local("/fixture"));
+    assert!(!browser.preferences().show_hidden);
+
+    browser.select_entries_by_name(&[".secret.txt".to_owned()]);
+
+    assert!(browser.preferences().show_hidden);
+    let selected = browser.selected_positions(0);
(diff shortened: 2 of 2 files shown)


l0gicgate:
Reviewed the complete diff against verified main `22e9c30ba1190e63147e9fcd5982c907650deae4`, already included by the existing merge commit. No production-code blockers found. There were no conversation comments, submitted reviews, or inline review threads to reconcile.

Expanded the regression to cover name/location selection, mixed hidden/visible requests, visible-only and missing/empty requests, and hidden-visibility notification preceding selection. Removed the redundant added comment; the final diff adds no comments.

Validation: pinned `scripts/quality.sh all` passed (format, Clippy with warnings denied, full Rust tests). The preceding full Rust run reported 1,584 passed and 19 existing ignored tests. Full canonical E2E passed all 654 tests with four workers. The initial auto/16-worker run had delete-animation and peek-focus failures; these did not recur in the complete four-worker run. `git diff --check` passed.

No UI behavior was changed during review. Manual ShowItems exercise and visual evidence remain pending; no capture or GitHub attachment is claimed. The PR’s existing visual-evidence N/A does not satisfy the repository requirement for this user-visible behavior. No authenticated attachment-upload tool is available in this session.

PR #942, opened 2026-09-13
feat(keyboard): keep arrow navigation scoped to the file list

### LINKED ISSUE #939: feat(keyboard): toggle to keep arrow navigation scoped to the file list
### Problem or use case

When navigating files fast with arrow keys (or h/j/k/l) in List or Icons mode, focus can escape the file list and land in the UI layer. Once selection reaches the first row, the next Up moves focus into the pane header (sort/view-mode buttons); from there Left/Right traverse header controls, and further Up can leak into the sidebar and the top navigation bar. The same happens at the left edge in Icons, where Left enters the sidebar.

This is the documented behavior from #291, but when moving quickly it stumbles: the user expects to keep arrowing through files and instead ends up tabbing around chrome, losing their place in the content. There is no way to opt out of arrows reaching the UI layer.

### Proposed solution

Add a saved preference (off by default, to preserve current keyboard-only/accessibility behavior) that keeps arrow navigation scoped to the file list. When enabled:

- Up from the first row stays on the first row (or stops) instead of focusing the header.
- Left at the edge stays in the file view instead of entering the sidebar.
- The header, sidebar, and top-bar controls remain reachable by Tab and mouse click, just not by arrows.

Surface the toggle in Settings → Keybindings, next to "Show keybinding hints", since both govern keyboard behavior. Follow `docs/preferences.md` for live updates across open windows.

Additionally, bind a keyboard shortcut to toggle this preference live, so power users can flip the mode mid-session without opening Settings (e.g. when they do want to arrow through the toolbar for a moment).

### Alternatives considered

- Hard-disable arrow escape into chrome: rejected, breaks keyboard-only and accessibility workflows that rely on full arrow traversal.
- Settings toggle only (no shortcut): simpler, but slower to flip contextually.

JoeJoeflyn:
## Description

Add a saved **Keep arrows in file list** switch under Settings → General → Browsing, off by default. When enabled, arrow keys no longer move focus out of the file list into the pane header, sidebar, or top navigation bar — fast arrowing stays scoped to content. Tab and the mouse still reach those controls. A live **Ctrl+\\** shortcut toggles the mode mid-session, and the file chooser respects the same preference.

This is an opt-out for the arrow-escape behavior introduced in #291, defaulting off to preserve existing keyboard-only and accessibility workflows.

## Visual evidence

Review-update screenshots (Columns focus and settings search) are captured locally; GitHub attachment upload is pending owner action. See the review follow-up comment.

https://github.com/user-attachments/assets/39eb4417-5fef-4fa3-bceb-709790743eb5

## How to test

1. In List or Icons, focus the file list and hold Up to the first row. With the toggle off, focus escapes into the pane header (unchanged).
2. Turn **Keep arrows in file list** on (or press **Ctrl+\\**). Hold Up from the first row again; focus stays in the file list.
3. With it on, press Left at the left edge; focus stays in the file view instead of entering the sidebar.
4. Press **Ctrl+\\** again to toggle off; confirm the Settings switch reflects the change, and that Tab and mouse still reach the header, sidebar, and top-bar controls.
5. Open the file chooser and repeat; it respects the same preference.

**Expected result:** With the toggle on, arrow keys never leave the file list. Tab and mouse remain the way to reach the chrome. The choice persists across restarts and applies to the file chooser.

## Related issue

Closes #939


wmfeht:
**Request changes** (private review of `b1c6904d`; one blocker).

The file chooser still sends Left at the sidebar edge into the sidebar when **Keep arrows in file list** is on.

`install_shortcuts` in `src/ui/chooser.rs` still handles this first:

```
if key == Left && sidebar visible && item_view_has_focus && item_at_sidebar_edge()
    → focus_active_place()
```

`item_at_sidebar_edge()` is true for List, Icons at the left tile, and Columns depth 0. That path returns `Stop`, so the later `h | Left` arm the PR gated with `!arrow_navigation_scoped()` never runs for Left. With type-to-search off, remapped `h` is already `Left` and takes the same path.

Main-window List/Icons correctly remap `SinglePaneArrow::Sidebar` to Stay. The chooser Up-to-header gates look right. This is the PR's own chooser Left-at-edge case (how-to-test steps 3 and 5, and #939).

Fix: gate the early `item_at_sidebar_edge` Left handler with the preference (and keep the Columns `h`/`Left` arm consistent). Add a chooser keyboard test that seeds the preference on and asserts Left from a left-edge item stays in the file view in Icons and List.

l0gicgate:
Review follow-up: merged verified `lgse/strata` main `9891206cb2bd017ac3de899a56e944ae91e64141` without rebasing, preserving main's keyboard-test consolidation and the new scope regression.

- Fixed Columns Left escaping into the sidebar while scope is enabled; registered the switch in Settings search and corrected its documented location to General → Browsing.
- Expanded chooser and two-window coverage for live on/off changes across all views, plus real arrows/hjkl and Ctrl+\ input. Removed both redundant test comments added by this PR.
- [Chooser feedback](https://github.com/lgse/strata/pull/942#issuecomment-5653248476): the early handler was already gated at the reviewed head; strengthened tests now verify actual left-edge readiness and both preference directions in List, Icons, and Columns.
- [Settings text-size failure](https://github.com/lgse/strata/pull/942#issuecomment-5654694218): the existing regression passes on the updated branch.
- Full local validation exposed an archive-test race: its content assertion now waits for the existing extraction-progress dismissal condition, rather than reading a newly created but unfinished file.

CI also exposed a thumbnail sampling race. Coordinates are now measured before capture after the row settles; the latest main additionally waits for repaint. The pixel-color check remains intact.

Headless visual exercise confirmed Columns Left stays on the item when enabled and reaches Home in the sidebar when disabled; Settings search reveals the switch. No desktop session was used.

**Updated visual evidence is pending owner upload:** no supported authenticated attachment-upload tool is available here. Please upload these captures through the GitHub editor (local paths below are not attachments):

- `/home/l0gicgate/Documents/github/strata-worktrees/939-review-942/target/review-942/media/columns-scoped-left.png` — enabled: focus stays on the item.
- `/home/l0gicgate/Documents/github/strata-worktrees/939-review-942/target/review-942/media/columns-unscoped-left.png` — disabled: focus reaches the sidebar.
- `/home/l0gicgate/Documents/github/strata-worktrees/939-review-942/target/review-942/media/settings-arrow-scope-search.png` — switch discoverable under General → Browsing.

PR #944, opened 2026-09-13
fix(updates): show the update notice in every open window

### LINKED ISSUE #866: fix(updates): show the update notice in every open window
### Strata version

### Installation method

Built from source

### What happened?

An update notice that appears in the window that ran Check now does not appear in other windows of the same process. `LAST_COMPLETED_CHECK` is process-wide, so a later window skips the 8s startup check. Preference binds only *clear* notices. GTK tests cover clear, not appear.

### Steps to reproduce

1. Enable automatic checks or run Check now in window A so `vX.Y.Z available` shows in that sidebar.
2. Open window B in the same process with auto-check already on.
3. Wait past the 8s startup due check.
4. Observe window A still shows the notice and window B does not.

### Expected behavior

A completed offer should show the in-window notice in every open window, without opening Settings. Clearing across windows already works.

### Screenshots or video

### Relevant logs

### Additional context

Found in the v0.17.0-nightly.20260912 QA sweep. `maybe_run_due_update_check` uses process-wide `LAST_COMPLETED_CHECK` (`src/ui/settings.rs`). `bind_update_notice_preferences` only clears. Distinct from closed #123 (notify views when the *release channel* changes) and #212 (general preference sync). About identity is correct when built with `STRATA_RELEASE_TAG` / `STRATA_BUILD_KIND=nightly`.

JoeJoeflyn:
## Description

Broadcast completed update checks process-wide so every live window receives the same sidebar notice, while caching the latest valid offer for windows opened later. Automatic results are discarded when update checks are disabled or the release channel changes while a request is in flight.

Start the automatic check on an idle callback after the first themed frame instead of waiting eight seconds, and start a due check immediately when Settings → Updates opens. The check and package-manager detection remain off the GTK thread.

## Visual evidence

Update notice in the sidebar

## How to test

1. Launch Strata with automatic update checks enabled and an update available.
2. Open a second window before the first check completes.

**Expected result:** The check is queued after the first themed frame, without a fixed startup delay, and both open windows show the sidebar notice when it completes.

3. Open another window after the check completes.

**Expected result:** The cached notice appears immediately without another network check.

4. Disable automatic checks or switch release channels while an automatic check is in flight.

**Expected result:** The stale result is not shown or replayed to a later window.

5. Open Settings → Updates when another automatic check is due.

**Expected result:** The due check starts immediately.

## Related issue

Closes #866


wmfeht:
Private review of `de6e1cd` (blocker only).

**Cache is written on Available before the existing preference guard, and replay does not re-check those preferences.** In `maybe_run_due_update_check`, `LAST_UPDATE_RESULT` is assigned, then `notice` runs only if `checks_for_updates()` is still on and `release_channel()` still matches the in-flight channel. `replay_cached_update_notice` delivers the cache unconditionally.

If the user disables checks or switches channel while a due check is in flight, `bind_update_notice_preferences` clears the cache and notices, then the completion path refills the cache and correctly skips `notice` on the triggering window. A later window's `schedule_due_update_check` then shows the offer anyway.

Write the cache only inside that same guard (and/or skip replay when checks are off / the channel no longer matches). `cached_update_notice_replays_to_later_windows` injects the cache via `set_cached_update_notice_for_test` and does not exercise this completion path.

JoeJoeflyn:
still working on

JoeJoeflyn:
Addressed the stale-cache review feedback in `488f609`: automatic results are now cached and broadcast only after confirming update checks remain enabled and the release channel still matches. The regression test covers a preference change while the due result is in flight. The same update also broadcasts to windows that were already open.

wmfeht:
Private review of `488f609` (blocker only).

**Due `Failed` / `UpToDate` now broadcast-clear and can hide a Check now offer.** `complete_due_update_check` calls `publish_update_notice(None)` for those results. On `main`, the due path only notified on `Available`, so a failed or up-to-date automatic check left an existing sidebar offer alone. Check now still starts a second fetch without `CHECK_IN_FLIGHT` or a shared generation.

This head also starts the due check at two seconds and again when Settings → Updates first builds, so the overlap is the normal Check now path. Sequence: due check in flight → Check now returns `Available` and publishes → due `Failed` (or `UpToDate`) lands last → every window loses the notice while Settings still shows the Check now result.

Do not publish `None` from the due completion path for `Failed`, and do not let a concurrent due result override Check now without a shared in-flight owner. The stale-cache guard from `de6e1cd` looks correct.

wmfeht:
Private review of `c534fc5c` (blocker only).

**Sharing `CHECK_GENERATION` with Check now can freeze another window's Updates row.** `run_check` used a row-local generation so a newer check on that same row owned `checking` and the button. This head moves that cell to a process-wide `CHECK_GENERATION` also bumped by `maybe_run_due_update_check`. The stale Check now poll still returns without restoring sensitivity:

```text
if is_stale_check(my_generation, CHECK_GENERATION.get()) {
    return glib::ControlFlow::Break;
}
```

Two windows with Settings → Updates: Check now in A, or a channel toggle / re-enable (each open Updates page's `run_check` fires), increments the global generation from B. A's in-flight poll exits as stale and leaves "Checking for updates…" with Check now disabled. Due-vs-Check now sharing is the right fix for the `488f609` clobber; the Check now UI still assumes the newer check is the same row.

Keep the due path on the shared generation / in-flight flag. Do not let a different window's `run_check` abandon another row's button. Due `Failed`/`UpToDate` no longer broadcast-clear, and the preference guard from `de6e1cd` still looks correct.

l0gicgate:
Merged verified main `41c9bf69303924b9d4dd339b528f0e3c24227bb4` and replaced the fixed startup delay with an idle callback queued after the first themed frame. The first automatic check still bypasses the persisted cache; network/package-manager work stays off the GTK thread, disabled checks remain respected, and concurrent windows share the in-flight guard. Removed redundant added comments.

Reviewed all conversation feedback (no submitted reviews or inline threads): the preference guard prevents stale automatic offers from being cached/replayed; due Failed/UpToDate results preserve notices and superseded generations cannot overwrite a newer check; row-local generations restore another window’s Check now button without interfering with a newer same-row check.

Manual verification: live-release startup latency has not been measured; no timing recording was captured or uploaded. The existing screenshot shows appearance, not latency. To exercise: start with automatic checks enabled and a cached result; after first paint, expect a fresh background request without the old fixed delay. Opening another window must not duplicate an in-flight request.

l0gicgate:
CI follow-up: all local pinned checks passed (`./scripts/quality.sh fmt`, `clippy`, `test`: 1,667 passed / 19 ignored; canonical Podman E2E: 674 passed). Initial CI failed one test before its body because GTK could not initialize the private display. Retrying failed jobs then hit an artifact infrastructure blocker: `quality-bundle-2` does not exist (Rust shard 1, https://github.com/lgse/strata/actions/runs/34909898674/job/104196415566). The quality aggregate remains red; this is not a green/merge-ready attestation. No test assertions were weakened. Review work is pushed at `32b0952d`; worktree retained pending CI resolution.

l0gicgate:
Resolved the red CI gate without changing code: a complete workflow rerun regenerated matching attempt-scoped artifacts. Attempt 3 passed both Rust shards, formatting/lint, exact test coverage aggregation, and all E2E shards. The original GTK initialization failure did not recur. Head remains the already-pushed `32b0952d3175c3089af4fbb5719c8ad28a704420`; no empty commit or test weakening was needed. https://github.com/lgse/strata/actions/runs/34909898674/attempts/3

PR #946, opened 2026-09-13
feat(browser): sort by Type using MIME descriptions in all views

### LINKED ISSUE #508: feat(browser): sort by Type using MIME descriptions in all views
### Problem or use case

**Type** sort is wired up in Columns, Icons, and List, but it only compares coarse filesystem kind (`Folder` / `File` / link / other). Every ordinary file is the same key, so with Folders first on (the default) Type ascending and Type descending look like Name ascending. Toggling Type does nothing useful in a mixed folder.

A file manager should cluster by what the file *is* — JSON with JSON, PDFs with PDFs — not by inode kind. List **Group by type** already shows shared-MIME descriptions (`JSON document`, `Python script`, …). Type *sort* should use that same notion.

### Proposed solution

When the pane sort key is **Type**, order by filename MIME **description**. Guess the type from the **name only** (`content_type_guess(name, None)`), the same path Group by type already uses. Do not sort by `EntryKind`. Do not sniff file bytes, stat for emptiness, or query `standard::content-type` during enumerate.

**Extensions only, not Dolphin-style sniffing.** Dolphin (and GIO `standard::content-type` on the inode) inspects contents. A 0-byte `z.txt` becomes `application/x-zerosize` / empty and shows as **Other**; a `.txt` with bytes shows as **Text**. Strata will not do that. `.txt` is always the OS description for `text/plain` (e.g. Plain text document), empty or not. Same ceiling as Group by type: wrong extensions and extensionless files follow the name, not the bytes. Matching Dolphin would mean I/O on every file and fights streaming / incremental sort.

**Ascending**
1. Known MIME descriptions A–Z
2. **Other** last (unrecognized / `application/octet-stream` / empty description)
3. Within each type, names always A–Z

**Descending**
1. **Other** first
2. Known MIME descriptions Z–A
3. Within each type, names still A–Z (do not reverse names)

**Folders first** (when enabled)
- Folders come before known types and before Other
- Folders themselves are always A–Z, including when Type is descending
- Then apply the Type order above to everything else

When Folders first is off, folders take part in the Type order like any other known type (e.g. “Folder” among A–Z / Z–A). Other is a special bucket: last in ascending, first in descending — not sorted as the word “Other” in the alphabet.

Example with Folders first on:

- Ascending: `alpha/` `zeta/` → Markdown… → JSON… → PDF… → Other (`archive-index`, `blob.qqqqq`)
- Descending: `alpha/` `zeta/` → Other → PDF… → JSON… → Markdown…

Names inside each type stay A–Z in both directions.

This is the shared pane sort, so Columns, Icons, and List all follow it. If List **Group by type** is on while sorting by Type, section order should match so headings and rows agree. The List Type column should show the same MIME description (and Other) the sort uses, not `File`/`Folder` kind labels.

Keep it on the cheap path: suffix-cached labels, no extra `String` on every `FileEntry`, no Size/Modified-style metadata fill or second sort.

### Alternatives considered

- Keep sorting by `EntryKind`: already implemented, and it is a no-op for ordinary files.
- Treat Other as a normal A–Z label: would bury unknowns in the O’s instead of parking them at the end (or start, when descending).
- Reverse names when Type is descending: worse for scanning a type alphabetically.
- Sniff contents like Dolphin (empty `.txt` → Other, non-empty `.txt` → Text): more accurate for zerosize and misnamed files, but real I/O on every entry. Out of scope; filename MIME is the Type sort.

### Visual examples

Image
Image
Image

### Strata version

`main`

### Additional context

Lift the existing MIME description helper and suffix cache out of List Group by type so `compare_entries` can use it. Type sort already runs once per listing; Columns, Icons, and List all render that order. Object kind (`EntryKind`) stays an internal tag for directories, links, and capabilities; it is not the Type sort key.

Labels still come from the OS shared MIME database (e.g. `text/plain` → “Plain text document”). Strata does not invent those strings. It only maps suffix → MIME id → description, with Other as the unknown bucket.

JoeJoeflyn:
## Description

Replaces coarse `EntryKind` sorting with filename-based MIME description ordering across Columns, Icons, and List views.

- Resolves MIME content types strictly from filename via `gio::content_type_guess(name, None)` with bounded thread-local filename caching to ensure zero disk I/O and no heap overhead per `FileEntry`.
- In Ascending order: Known MIME types are sorted A–Z, with unrecognized ("Other") entries placed last, and identical types tie-broken A–Z by filename.
- In Descending order: "Other" entries are placed first, known MIME types are sorted Z–A, and identical types remain tie-broken A–Z by filename.
- Respects the *Folders first* preference: folders lead the listing A–Z when enabled, or participate alphabetically as `"Folder"` among known MIME types when disabled.
- Updates List view's Type column and *Group by Type* sections to display and group by MIME descriptions instead of generic file tags.

## Visual evidence

https://github.com/user-attachments/assets/86496cd4-c73e-45d9-a31c-11664bd673f2

## How to test

1. Open a directory containing different file types (e.g. `.pdf`, `.py`, `.json`, `.txt`, folders, and unrecognized files).
2. Switch sort key to **Type** and toggle between **Ascending** and **Descending**.
3. In List view (`Ctrl+2`), observe the **Type** column displaying human-readable MIME descriptions (e.g. "JSON document", "PDF document", "Folder", "Other").
4. Enable **Group by Type** in List view and verify the section headers.

**Expected result:**
Entries sort by their human-readable MIME descriptions in A–Z / Z–A order, "Other" files remain at the boundary (last in ascending, first in descending), filenames inside each type tie-break A–Z, and List view displays MIME descriptions.

## Related issue

Closes #508


l0gicgate:
Reviewed and pushed the `main` merge (`51f561b2f331cd0364a6771e0fc30672e742d512`) plus two fixes:

- Grouped Icons/List now follow live Type direction and Folders first, with stable filename ties; special filesystem entries consistently use Other.
- MIME guesses are cached by complete filename so `.tar.gz`/`.gz` and filename-specific MIME globs cannot contaminate one another. The cache remains bounded and filename-only, with no file-content reads or added `FileEntry` storage. Updated the description to reflect this correctness tradeoff.
- Updated fixtures for main's metadata fields, added regressions, and removed redundant comments. No existing review comments or threads required reconciliation.

Private-display interaction checks exercised ascending/descending in all three views, grouped Icons/List, both folder preferences, JSON filename ties, unknown files, and distinct compressed archive types. Local pinned formatting/Clippy and the full Rust suite passed (1,699 passed; 18 existing ignores); canonical E2E passed 728 tests, followed by 30 sorting/search tests after the final refactor.

**Updated visual evidence pending upload:** authenticated attachment-upload tooling is unavailable in this session. Please upload these sanitized captures through GitHub's editor:
- `/home/l0gicgate/.cache/pi-tmp/pr946-review.KoYI1W/visuals/list-folders-False-ascending.png` — grouped Type ascending, Other last.
- `/home/l0gicgate/.cache/pi-tmp/pr946-review.KoYI1W/visuals/list-folders-False-descending.png` — live Type descending, Other first, folders participate in type order.
- `/home/l0gicgate/.cache/pi-tmp/pr946-review.KoYI1W/visuals/list-folders-True-descending.png` — descending with folders remaining first and A–Z.

These are local paths, not uploaded attachments. The original PR video remains in the description. `main` advanced during review; the exact fetched/merged baseline above is preserved rather than treating a subsequently moved tracking ref as the reviewed baseline.

PR #955, opened 2026-09-13
feat(browser): add optional checkbox selection for multiple items

### LINKED ISSUE #937: Add checkboxes to select multiple items
It would be nice if strata provides an optional feature, where multiple items can be selected with a checkbox.

JoeJoeflyn:
## Description

Add a saved **Checkbox selection** switch under Settings → BROWSING, off by default. When enabled, each List view row shows a checkbox, plus a "select all" checkbox in the header. Clicking a row checkbox selects/deselects that item; clicking the header checkbox selects/deselects all. The header shows an inconsistent (dash) state when some but not all items are selected.

Checkboxes drive the existing `gtk::MultiSelection` model, so marquee, Ctrl+A, keyboard, and checkbox selection all stay in sync. A guard prevents the select_all checkbox's programmatic state updates from re-triggering selection changes (re-entrancy loop). Marquee origin surfaces now treat `CheckButton` as interactive chrome so clicks reach the checkbox instead of starting a drag. Selection changes (including Esc clearing) sync checkbox state even when the `selection_changed` handler is suppressed by the syncing guard.

This is off by default to preserve the existing modifier-based selection flow for power users.

## Visual evidence

https://github.com/user-attachments/assets/fd3384e7-9bfb-4a31-ad11-2059dee6f85f

## How to test

1. Open Settings → BROWSING, turn **Checkbox selection** on.
2. Switch to List view (Ctrl+3). Each row should show a checkbox on the left, plus a "select all" checkbox in the header.
3. Click individual row checkboxes to select/deselect items.
4. Click the header "select all" checkbox to select/deselect all. Rows should show the accent highlight color.
5. Select some but not all items — the header checkbox should show an inconsistent (dash) state.
6. Press Ctrl+A — checkboxes and highlight should sync.
7. Drag marquee — checkboxes should sync.
8. Press Esc — all checkboxes and selection should clear.
9. Toggle the pref off — checkboxes should hide.

**Expected result:** With the toggle on, checkboxes appear on every List row and the header. They stay in sync with all selection methods. The choice persists across restarts.

## Related issue

Closes #937


wmfeht:
**Blocker:** the new Settings row is not registered for Settings search.

`settings_option` tags rows from `settings/search.rs` `TARGETS` by exact title. There is no TARGET for **Checkbox selection**, so `search::tag` is a no-op.

Searching Settings for "checkbox" / "checkbox selection" returns no matches. Untagged rows also stay visible when a sibling in the same BROWSING group matches, so a query like "Folder peeking" still shows this switch.

`docs/preferences.md` requires new settings to be registered in `settings/search.rs`. Please add a TARGET (id e.g. `checkbox-selection`, aliases for checkbox / multi-select / list) so the row gets a `settings-search-*` name.

JoeJoeflyn:
Addressed the Settings search blocker in b17744c: registered Checkbox selection with checkbox/multi-select/list aliases and added coverage for direct matches plus sibling filtering.

archisman-panigrahi:
This is already good. However, Windows file explorer and Dolphin (in KDE) also provide this feature in all views (the checkbox shows up after hovering with mouse in other views). It would also be nice to have.

JoeJoeflyn:
@archisman-panigrahi sure, ill take a look at how they implemented on those apps

archisman-panigrahi:
> sure, ill take a look at how they implemented on those apps

You can also look at this merge request of thunar

JoeJoeflyn:
still working on this

PR #959, opened 2026-09-13
feat(navigation): add ranked folder jump

### LINKED ISSUE #276: feat(navigation): add ranked quick folder search
### Problem or use case

Strata's pins work well for a small, manually curated set of locations, but there is no keyboard-first way to jump back to folders discovered during everyday browsing. Recursive search can find a known path, but it does not prioritize locations based on how recently or frequently they were visited.

Users want a quick folder-only search, similar to Yazi's zoxide-powered jump workflow, that learns from navigation and makes commonly used or recently visited directories easy to reopen.

### Proposed solution

Add a quick folder-jump palette that reuses Strata's existing search interaction where practical:

- Search directories only by folder name and path.
- Rank matches using navigation history, taking both frequency and recency into account.
- Record folders visited while browsing in Strata so results improve over time.
- Open the selected folder immediately and support keyboard navigation throughout the palette.
- Keep this distinct from pinned folders: pins remain explicit favorites, while quick jump is learned automatically.

The implementation could either maintain Strata's own ranking data or optionally integrate with zoxide. If zoxide is used, the feature should degrade gracefully when it is not installed and should make the data-sharing behavior clear.

### Alternatives considered

- **Pinned folders:** useful for permanent favorites, but require manual maintenance and do not adapt to browsing habits.
- **Recursive search (`Ctrl+K`):** searches the current tree rather than a learned set of previously visited folders.
- **Location entry (`Ctrl+L`):** fast when the exact path is already known.
- **Direct zoxide integration only:** provides an established frecency database, but would introduce an external dependency and may mix terminal and Strata navigation history in ways users do not expect.

### Visual examples

N/A — the requested interaction is comparable to Strata's current search palette, but restricted to folders and ordered by recency/frequency. Yazi's zoxide jump workflow is a behavioral reference.

### Strata version

N/A

### Additional context

The request specifically referenced Yazi and suggested that its behavior may be backed by zoxide. The desired outcome is a ranked, recent/frequent folder search; zoxide compatibility is an implementation option rather than a requirement.

JoeJoeflyn:
## Description

Add a keyboard-first folder jump palette backed by Strata's own navigation history. Successful native folder visits are persisted locally and ranked using fuzzy relevance, frequency, and recency without requiring zoxide or changing pinned folders.

## Visual evidence

https://github.com/user-attachments/assets/72e9ab67-3950-413b-878e-d8a99b79426b

## How to test

1. Launch Strata and browse into several folders, revisiting one of them a few times.
2. Press `Ctrl+Shift+K`, type part of a visited folder's name or path, use the arrow keys, and press Enter.

**Expected result:**

The palette initially lists visited folders by frecency, filters to folder-only matches as you type, and immediately navigates to the selected folder. `Ctrl+K` continues to open global file search.

## Related issue

Closes #276


JoeJoeflyn:
still working on

JoeJoeflyn:
good to go

PR #985, opened 2026-09-14
ci: reduce E2E and Rust test shard counts

l0gicgate:
## Description

Reduce E2E from at most eight shards to three, with a 180-second worker target, and Rust tests from four shards to two. Balance Rust tests across both runners while executing the sensitive regression in its own libtest process. Preserve GUI isolation and exact-once coverage gates. Refresh its stale 154-second timing hint to the 1.32 seconds measured in run 34811363582. Update planner fixtures and documentation.

Fix the creation-collision E2E setup to refresh and use keyboard selection after inserting fixture files: the failed run showed creation in pictures instead of the root. Also correct the spelling-policy warning in its comment.

Owner explicitly authorized no ticket and pushing without local tests. Local test/build/lint suites were not run; diff review and `git diff --check` completed. Latest changes await CI validation.

## Visual evidence

N/A — CI scheduling and test-only changes; no application UI changes.

## How to test

1. Inspect the generated plans and shard job timings on this PR run.
2. Inspect the creation-collision scenario result and aggregate coverage reports.

**Expected result:** At most three E2E shards and two Rust shards cover every runnable test exactly once. The isolated Rust regression runs alone in a separate process, then its runner executes other assigned tests. Creation selects the first free numbered filename without overwriting existing files, a broken symlink, or a directory. E2E shard jobs around 3–4 minutes are acceptable; actual Rust consolidation timing remains unverified.

## Related issue

None — owner explicitly waived the ticket requirement.


l0gicgate:
Updated the sharding regression fixtures missed in the initial change: tighter-budget inputs now stay below the cap before growing, unknown-duration and serial-group inputs cross the new budget, and cap coverage exercises the new 168/169-test boundary. The default-plan regression checks observable two-shard scheduling instead of repeating constants. No local tests were run, per owner instruction; diff review and `git diff --check` completed. CI validation remains pending.

PR #1000, opened 2026-09-14
fix(trash): refuse restore through a same-volume symlink parent

### LINKED ISSUE #862: fix(trash): refuse restore through a same-volume symlink parent

pdwaldrop:
## Description

Trash Restore's destination planning resolved the confirmed original location's parent chain with `Path::canonicalize`, which silently follows any symlink regardless of where it points. The existing safety check only rejected a symlink that left the trash volume entirely (caught incidentally by the mount/device comparison), so a parent directory replaced with a symlink to another real directory *on the same volume* passed every check, and the confirmation dialog itself already showed the wrong (symlink-resolved) destination before any move happened.

Restore's execution step had a matching gap: its `openat2` confinement for the destination parent already used `RESOLVE_BENEATH` (blocks escaping the volume) and `RESOLVE_NO_XDEV`, but not `RESOLVE_NO_SYMLINKS` — so even if planning were bypassed, the same same-volume redirect would slip through there too.

Both are fixed the same way: walk the parent chain component by component from the already-canonical volume root, and fail closed on the first component that turns out to be a symlink, rather than resolving through it.

## Visual evidence

N/A — this only changes destination-path validation and kernel-level path resolution flags for Trash Restore; there is no UI-visible change (the existing confirmation dialog and error-message plumbing are unchanged).

## How to test

1. Trash a nested file, e.g. `documents/nested/inner.txt`.
2. Replace `nested` with a symlink to another real directory on the *same* volume, e.g. `ln -s ../pictures documents/nested`.
3. Restore the file from Trash.

**Expected result:** Restore refuses the item with a clear error (its parent folder is a symlink) instead of writing `inner.txt` into the symlink's target directory. The existing cross-volume symlink-escape case (`symlink_parent_that_leaves_the_volume_is_rejected`) and all other existing restore-safety tests continue to pass.

## Related issue

Closes #862


l0gicgate:
Closing as this is a duplicate of https://github.com/lgse/strata/pull/913

PR #995, opened 2026-09-14
fix(columns): enable in-memory filtering for non-native locations (#994)

Rizmi:
Resolves #994

### Problem

1. In Miller Columns mode, typing into the pane filter (`Ctrl+F`) did not filter entries when viewing non-native locations such as Trash (`trash:///`), network places (`network:///`), or remote mounts (`smb://`, `sftp://`, `mtp://`, etc.).
2. When filtering was active in Trash, single-clicking an item attempted to open it with the non-native URI in an external web browser instead of selecting/previewing it.
3. In filtered columns, keyboard arrow navigation stopped after 1–2 jumps because directory commands stepped through hidden underlying directory indices instead of letting GTK's `ListView` step across visible filtered rows.

### Root Cause

1. **Non-native filter early exit**: In `src/ui/browser/columns.rs`, `bind_filter_query` assumed every column location had a native filesystem path:
   ```rust
   let Some(path) = state
       .browser
       .location_at(depth_for_search)
       .and_then(|loc| loc.native_path().map(Path::to_path_buf))
   else {
       return;
   };
   ```
   For `LocationKind::Uri` locations, `loc.native_path()` returns `None`. When this occurred, `search_active` was set to `true`, but execution returned early without notifying or applying GTK's `CustomFilter` (`apply_filter_query`). The column remained unfiltered while leaving `search_active` set.

2. **Premature activation of Trash files on single click**: `column_rows` queued `PendingActivationKind::Mapped` for any filtered item click, which called `activate_in_place` on pointer release. In Trash, this emitted `BrowserEvent::OpenRequested` for non-directory items, causing GIO to launch a web browser on `trash:///...`.

3. **Filtered keyboard navigation lockup**: In `window/keyboard.rs`, `selected_search_results().is_some()` previously checked only `column.search_handle` (recursive search). For in-memory filters, this returned `None`, allowing `item_navigation` to intercept arrow keys and call `browser.move_selection(1)`. `move_selection` advanced through the unfiltered directory entries; if subsequent entries were filtered out, the visible selection became stuck.

#### 1. In-Memory Column Filtering for Non-Native Locations
- In `columns.rs` (`bind_filter_query`), when `loc.native_path()` is `None`:
  - Cancel any background search handles/generations.
  - Call `deactivate_recursive_search` so `search_active` is `false`, search results are emptied, and the filtered model is backed by the column's directory model.
  - Apply `apply_filter_query` directly on the column's `FilterListModel` to filter loaded entries in-memory via `entry_matches`.
[...]

PR #1010, opened 2026-09-14
fix(columns): apply the pane filter to non-native locations

### LINKED ISSUE #994: fix(columns): pane filter does not filter entries in non-native locations (Trash, network, etc.)
### Summary
In Miller Columns mode, opening the pane filter (Ctrl+F) and typing a query does not filter directory entries when viewing non-native locations (such as Trash `trash:///`, network shares `smb://`, `network:///`, etc.).

### Steps to Reproduce
1. Open Strata in Columns mode.
2. Navigate to Trash (`trash:///`) or any non-native location.
3. Press Ctrl+F to open the pane filter.
4. Type any search term matching an item in the directory.

### Expected Behavior
The column list model should filter in real-time, showing only matching directory entries.

### Actual Behavior
All directory entries remain visible completely unfiltered. Additionally, selection synchronization and keyboard navigation stop functioning as expected because `recursive_search_active` is left set to `true`.

### Root Cause
In `src/ui/browser/columns.rs`, `bind_filter_query` assumes every active column has a native local filesystem path:
```rust
let Some(path) = state
    .browser
    .location_at(depth_for_search)
    .and_then(|loc| loc.native_path().map(Path::to_path_buf))
else {
    return;
};
```
For `LocationKind::Uri` locations (`trash:///`, `network:///`, `smb://`, etc.), `loc.native_path()` returns `None`. When this happens, `search_active` has already been set to `true`, but execution returns early before notifying or updating GTK's `CustomFilter` (`apply_filter_query`). Because the filter is never notified, the model remains unfiltered.

Preview:

Image

itsmunzir:
## Description

In Columns mode, typing in the pane filter (Ctrl+F) while viewing a non-native
location such as Trash or a network share left every entry visible and left the
recursive-search flag set, which also broke selection sync and keyboard
navigation for the column. `bind_filter_query` set the flag before checking for
a native path, and the early return never notified the local `CustomFilter`.

The native-path check now runs first: locations without a filesystem path take
the local name-filter path (`apply_filter_query`) with the recursive flag kept
clear, while native locations keep the existing recursive-search behavior. This
matches how the pane filter already behaves in Icons and List modes, and the
docs already describe pane filtering as plain name matching.

## Visual evidence

Trash column with the `apple` pane filter showing only `filter-demo-apple.txt` in the
same session. The broken state is already captured in #994.
strata-994-after-trash-filter

## How to test

1. Trash two files with distinct names (for example `filter-demo-apple.txt` and
   `filter-demo-zebra.txt`).
2. Open Trash from the sidebar in Columns mode and press Ctrl+F in the Trash
   column.
3. Type `apple`.

**Expected result:** only the matching entry stays visible, no
`filter-result` styling appears, selection and arrow-key navigation keep
working, and clearing the input restores the full listing.

## Related issue

Closes #994


Rizmi:
duplicate ?
https://github.com/lgse/strata/pull/995

l0gicgate:
Closing this as a duplicate, I'm working on @Rizmi's PR right now, it'll be fixed in next release.

PR #1044, opened 2026-09-15
feat(browser): add Finder-style batch rename dialog

JoeJoeflyn:
## Description

Adds Finder-style batch rename: select 2+ files, right-click, and choose **Rename N items…**. The dialog offers the same three modes as Finder:
- **Replace Text**: literal find/replace with live preview of the first changed entry.
- **Add Text**: prepend before name or insert after name before extension, with target-name preview ensuring prepended prefixes and appended suffixes remain clearly visible.
- **Format**: Counter (zero-padded to 5 digits, starting at 0 or 1), Index (starting at 0 or 1), or Date (with timestamp and unique position index, hiding unused start number input).

Key improvements and polish:
- **Live Rename Sensitivity**: The Rename button starts disabled and becomes active only when all required inputs for the active mode are present.
- **Preview Accuracy**: Evaluates the selection for the first entry whose name actually changes, preventing false "Names are unchanged" previews when non-matching files appear earlier in the selection.
- **Dialog Geometry & Ellipsizing**: Constrains modal width to 480px and middle-ellipsizes long preview names without widening the modal; full transformation mapping (`old → new`) is available on hover tooltip.
- **0-Indexed Numbering**: Allows `start_number >= 0` for both Counter (`00000`) and Index (`0`) formats.
- **Error State Management**: Automatically clears input `.error` highlights and tooltips as soon as valid text is typed or modes are changed.

## Visual evidence

Pending — please attach a short clip or screenshot via the GitHub editor (multi-select → Rename N items… → each mode with the live example, plus Ctrl+Z after a batch).

## How to test

1. Select multiple files in Columns or Grid view, right-click, and select **Rename N items…**.
2. **Replace Text**:
   - Verify the Rename button is disabled when Find is empty and enables as soon as text is typed.
   - If the first file in the selection does not match Find, verify the preview still reflects the first file that *does* change rather than "Names are unchanged".
3. **Add Text**:
   - Add text "After name" and verify extension is preserved (e.g. `doc_final.pdf`).
   - Add text "Before name" on long filenames and verify the prepended text is visible at the beginning of the `Example: <name>` line.
   - Hover over the preview label to verify the full `old → new` tooltip.
4. **Format**:
   - Switch to **Counter** or **Index**, set custom name, and set "Start numbers at" to `0`. Verify numbering starts at `00000` / `0`.
   - Switch to **Date**: verify "Start numbers at" is automatically hidden and the date timestamp with position index is previewed.
5. Apply rename and press <kbd>Ctrl</kbd>+<kbd>Z</kbd> to verify all files cleanly restore to their original names.

## Related issue

N/A


wmfeht:
**request-changes** on `c897db9` — two blockers.

1. **Sequential apply is not permutation-safe; duplicate planned names are dropped silently.** `rename_batch` calls `set_display_name_async` in selection order. GIO refuses a target that already exists, including a file in this same batch that will itself be renamed. `rename_many` also skips later items whose planned name collides with an earlier planned name, with no error (the “Nothing to rename” dialog only appears when every item was skipped).

   Format: files `photo.jpg` and `vacation 00001.jpg`; select `photo.jpg` first, then both; Format custom name `vacation`, Counter, start at 1. Plan is `photo.jpg` → `vacation 00001.jpg`, `vacation 00001.jpg` → `vacation 00002.jpg`. First rename fails on the still-present `vacation 00001.jpg`; second succeeds. `photo.jpg` is unchanged and the already-formatted file is now `vacation 00002.jpg`. Replace Text has the same chain (`1.txt` / `2.txt`, find `1` replace `2`). Two files mapping to one name rename one and leave the other with no summary.

   A final name set that only collides with sources in this batch is a valid permutation. Treat those sources as vacated (temp names or a safe order). Duplicate planned names should fail closed with a message, not skip.

2. **`plan_preview_name_handles_empty_entries`** (`src/ui/browser/batch_rename/tests.rs`) — the dialog returns for fewer than two entries. This test feeds an empty `Vec` and asserts the `unwrap_or_default` fallback. Drop it. Do not add a replacement.

GIO does not overwrite, so this is a wrong partial apply, not clobbering the target’s bytes.

wmfeht:
**request-changes** on `82e5b0db45f52bcf461dd29b1a9456a498d531be`.

The previous round (`c897db9`) is addressed. In-batch permutations stage occupied sources first, two items claiming one target fail closed with nothing renamed, and `plan_preview_name_handles_empty_entries` is gone. Clean swaps and chains apply, and Ctrl+Z replays through the same provider path.

CI is green and was not re-run.

## Blockers

1. **A failed chain leaves the staged file under `.strata-batch-*`, and undo does not bring it back.** `rename_batch` only rejects targets claimed twice inside the batch. A target held by a file that is not moving is discovered later, at `set_display_name`. Sources whose current names are planned targets are renamed to `.strata-batch-<id>-<attempt>` first. If a later final rename fails, `restore_staged` tries to put that source back on its original name, which an earlier success in the same batch may already occupy. Restore fails, the error says the item was left as the staging name, and the successful rename is still recorded for undo.

   Reproduction: `photo.jpg`, `vacation 00001.jpg`, and `vacation 00002.jpg` in one directory. Select only the first two, in that order. Format, custom name `vacation`, Counter, start at 1. Plan: `photo.jpg` → `vacation 00001.jpg`, `vacation 00001.jpg` → `vacation 00002.jpg`. `vacation 00002.jpg` is not in the batch.

   - `vacation 00001.jpg` is staged aside.
   - `photo.jpg` takes `vacation 00001.jpg`.
   - The staged file cannot become `vacation 00002.jpg` (outsider).
   - Restoring it to `vacation 00001.jpg` fails because `photo.jpg` is there.
   - `photo.jpg` is gone, `vacation 00001.jpg` holds photo's bytes, and the original `vacation 00001.jpg` bytes sit in a hidden `.strata-batch-*` name.

   Ctrl+Z undoes only the recorded success, so `vacation 00001.jpg` moves back to `photo.jpg` and the original `vacation 00001.jpg` name is empty. The bytes remain in the staging name. The same hole exists for Replace Text (`file.txt` → `file2.txt`, `file2.txt` → `file22.txt`, with `file22.txt` already present and not selected).

   Reject targets that exist and are not sources in this batch before any rename. If a later rename still fails after staging, unwind applied renames in this batch before restoring staged sources, so a staging name is not the surviving location. GIO still does not overwrite the outsider.

2. **Superseding the batch drops applied renames on the floor.** `delete`, `transfer`, and another rename call `begin_operation`, which drops the in-flight `LoadHandle` and clears `deferred_file_operation_changes`. It does not clear `rename_batch_operation`. The provider then emits `Cancelled`, but `operation_callback` ignores it because the request is no longer current, so there is no cancel dialog and no view update from the events that were just discarded. `record_rename_undo` only handles `RenamedBatch`, and `emit_batch_cancelled` forwards destination locations without the `RenameBatchRecord`s, so a cancel that is still current also cannot undo the files already renamed. `cancelled_batch_rename_leaves_sources_in_place` drops the handle before the task runs, so it never sees a partial apply. There is no progress dialog and no operation lock, so Delete or Paste during a longer batch is enough.

## Nits

- `selected_entries_in_selection_order` moves the anchor to the front and leaves the rest in view order. An upward range of alpha, bravo, charlie anchored on charlie numbers charlie, alpha, bravo, not charlie, bravo, alpha. The comment says a bottom-to-top range numbers from the bottom.
- The "Nothing to rename" dialog says every name is unchanged, invalid, or duplicated. `rename_many` returns `None` only when every item was unchanged or an identical `(location, new_name)` repeat. Invalid names and conflicting targets still go to the provider and show the summary.
- `FormatStyle::Date` is documented as `{name} {timestamp}` for every item. The planner also appends a position index.
- The staging-name search is an unbounded `loop` on the main context. `query_exists` is synchronous, the loop does not check the cancellable, and `attempt` is a `u32` that wraps.

## Notes

- Test scan: not a reason for this verdict. The empty-preview fallback test from the last round is gone. Adapter tests cover a clean chain, a swap, duplicate targets, a single outsider, an invalid name, and cancel-before-work. Planner and dialog tests each hit a distinct branch (empty find, extension split, counter width, index rollover, zero start, date position, first-changed preview). They are not tautological and not excessive. This review does not ask for more tests.
- No `working-docs/` plan on this PR. Scope is the dialog, planner, batch provider, undo/redo, and selection order described in the PR body.
- Residual: preview and apply each call `batch_rename_timestamp()`, so a Date rename can differ from the example by a second. A crash mid-batch can also leave a `.strata-batch-*` file; that is separate from blocker 1, which happens on a normal failed apply.

PR #1049, opened 2026-09-15
fix(columns): keep drag-and-drop working in narrow windows

JoeJoeflyn:
## Description

When the window is narrow enough that columns overflow horizontally, clipped columns show a reveal-button overlay covering their visible strip. That overlay was a targetable `gtk::Button`, so `pick()` returned it instead of the row beneath — the row's `DragSource` never received the press and drag-and-drop silently failed to start.

- Set `reveal_button.set_can_target(false)` so the overlay no longer wins `pick()` over the row beneath it.
- Make the scroller's peek `GestureClick` refuse to claim presses that land on actual file rows, not just clipped-column background.

## Visual evidence

Updated visual evidence is pending owner upload through the GitHub editor; the fix is a pointer-routing change with no visible layout difference at normal window sizes.

## How to test

1. Switch to Columns view and open a folder so two columns are showing.
2. Shrink the window narrow enough that the columns overflow horizontally (below ~600px).
3. Drag a file from the clipped column to the sidebar or another folder.

**Expected result:** The drag starts and completes normally, same as at full window width.

## Related issue

N/A — no existing issue was filed for this bug.


l0gicgate:
Reviewed the complete diff after merging verified upstream main `ac0f48662706d48c10184bd5343c7b3931651c00` (merge commit `b0714877`). Pushed `bb05948f` to remove redundant routing/test narration and retain only the non-obvious overlay hit-testing rationale. No behavioral blocker found; no UI behavior was changed during this review.

Validation: isolated rootless Podman `./scripts/quality.sh all` passed formatting, Clippy, and 1,730 Rust tests (18 ignored); `STRATA_CONTAINER_ENGINE=podman ./scripts/e2e.sh` passed all 729 tests. This includes the clipped-column drag regression and both older failures reported in https://github.com/lgse/strata/pull/1049#issuecomment-5685073533. All current-head CI gates passed. All conversation/review/thread endpoints were checked; there were no submitted reviews or inline threads.

No separate manual GUI session or new media capture was performed; verification used real pointer-driven E2E tests. Remaining repository-policy items: the PR has no linked bug issue, and its original visual evidence remains pending owner upload. Authenticated attachment-upload tooling is unavailable here; no media was posted. These items need owner resolution or explicit waiver before merge.

PR #1066, opened 2026-09-16
feat(search): reveal results in their containing folder

### LINKED ISSUE #1039: feat(search): reveal selected result in containing folder
From global search results it would be nice to have a “Reveal in Folder” action that closes search, navigates to the result’s parent directory, and selects/focuses the result. 

A bit more context on the use case: global search is great because nested results are flattened into a single list, but sometimes I want to jump from a result back to where it actually lives.

For the first version, my view of the requirements would be:

- Right-click a search result and choose Reveal in Folder
- Close search and navigate to the result’s containing directory
- Leave the result selected/focused and visible
- For directory results, reveal the parent and select the directory rather than opening it
- Ideally this can have a keyboard shortcut later

adamcbrewer:
## Description

Adds “Open containing folder” to global-search results through a context-menu action and Alt+Enter. The action closes search, opens the result’s parent directory, and selects, focuses, and scrolls to the original file or directory using existing browser reveal behavior. Add shortcut hints to the search footer and shortcuts reference. Existing in-pane filtering behavior is unchanged.

## Visual evidence

image

## How to test

1. Ctrl+K
2. find a file/directory
3. Right-click "Open containing folder" or press Alt+Enter. Expect the parent folder to open with the result selected, focused, and visible.

**Expected result:**

You should be in in the containing folder for the file or directory selected.

## Related issue

Closes #1039.

Note: I have noticed the browser column filter shows something similar in that files are shown "Open containing folder", but not dirs. I wonder if the conventions between the fitlers and global search should be aligned?


l0gicgate:
Nice one @adamcbrewer, will be in tonight's release.

l0gicgate:
Replaced the long partial-search footer message with a theme-aware Lucide warning triangle and “Partial results”; hovering shows the actual coverage reasons. Updated the existing coverage regression to verify the tooltip and clearing the warning when results are complete.

Verified the warning and hover tooltip on a private display with an unreadable-folder fixture. Reveal behavior remains covered for files/directories, context-menu/Alt+Enter, and all three browser modes. No blocking issues found in the remaining diff; there were no actionable review threads. Main advanced during validation, so merged a97f8c1c3a636ccf8ec089624cf28a0158851cb2 without conflicts and revalidated the merged result.

Visual evidence for this follow-up is pending owner upload: this session has no authenticated GitHub attachment-upload tool. Current screenshots are retained locally and will be supplied in the handoff.

PR #1080, opened 2026-09-17
fix(thumbnail): render thumbnails via a GVfs mirror path when available

### LINKED ISSUE #1079: Thumbnails never generated for files on non-local (smb://) backends
**Summary**

Strata never generates thumbnails for files on a network share mounted via GVFS (tested with `smb://`), even though the files are readable, the directory listing works, and metadata (size/mtime) loads correctly. Thumbnails work fine for local files.

**Environment**
- Strata version: 0.18.0
- Install: Arch Linux (Omarchy), `~/.local/bin/strata`
- Backend tested: SMB share mounted via GVFS (`gvfsd-smb`), with `gvfsd-fuse` running so a real local path exists at `/run/user/$UID/gvfs/smb-share:server=...,share=...`

**Steps to reproduce**
1. Mount an SMB share via GVFS (e.g. through the Files/Network sidebar entry, or `gio mount smb://<host>/<share>`).
2. Open that share in Strata, browse to a folder containing PNG/JPEG files.
3. Observe: no thumbnails render for any image, even after waiting, scrolling, or reopening the folder.
4. For comparison, open a local folder with the same image files — thumbnails render immediately.

**Evidence**

Ran with `RUST_LOG=trace`. For a local folder, thumbnailing fires immediately per file:
```
DEBUG strata::metrics: thumbnail requested total=1
DEBUG strata::metrics: thumbnail started total=1
DEBUG strata::metrics: thumbnails eligible count=1 total=1
```

For the exact same kind of files on the `smb://` share, directory load/metadata succeed, but there is **no `thumbnail requested`/`thumbnail started`/`thumbnails eligible` log line at all** — the thumbnail pipeline is never invoked for this backend:
```
INFO strata::adapters::local_files: directory load finished request_id=1 entries=5637 elapsed_ms=687
DEBUG strata::app::browser: selection changed depth=0 selected=1
DEBUG strata::app::browser::loading::metadata: metadata fill applied request_id=5 depth=0 filled=98 stale=0
```
(zero occurrences of "thumbnail" anywhere in the trace log for the SMB session, on a share mounted at `smb://nas1.funny-name.ts.net/home/Screenshots`)

**Why this seems fixable**

This isn't a sandboxing/permissions failure — it looks like an upfront eligibility check that gates thumbnailing to the local backend before the (bwrap-sandboxed) preview renderer is ever reached. On a system where `gvfsd-fuse` is running, GVFS-backed locations like `smb://` do have a real, bind-mountable local path (`gio info` reports a `local path:` for these mounts), so the existing sandboxed renderer should in principle be able to handle them the same way it handles local files — the eligibility check just appears to exclude them categorically.

**Suggested fix**

Relax the local-only eligibility gate to also allow thumbnailing when GIO reports a resolvable local path for the file (e.g. via `g_file_get_path()` after ensuring the GVFS FUSE bridge is available), rather than excluding all non-`file://` backends outright.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

spuder:
## Description

Render thumbnails for GVfs-backed locations through their local FUSE mirror when available, using the existing sandboxed renderer. Navigation retains its URI identity; without a mirror, files retain their generic icons. Metadata fills release thumbnails parked under the mirror path, avoiding duplicate network reads.

The URI-routing regressions use file:// fixtures; they do not exercise GVfs/FUSE integration. The author and prior reviewer separately reported live SMB/archive-mount verification in the discussion.

## Visual evidence

Author-provided sanitized captures from [the PR discussion](https://github.com/lgse/strata/pull/1080#issuecomment-5707145598):

Before:
SMB directory before: generic icons

After:
SMB directory after: image thumbnails

## How to test

1. Mount a GVfs network share containing images, for example with `gio mount smb://<host>/<share>`, and ensure its `gvfsd-fuse` bridge is available.
2. Open that share in Strata with the default name sort. Wait without scrolling or changing the sort.
3. Browse a backend without a local mirror to check the fallback.

**Expected result:** Images on the mirrored share acquire thumbnails after metadata loads, without requiring a rebind. Navigation continues to display the remote URI. Locations without a mirror retain generic icons.

## Related issue

Closes #1079


spuder:
3dprints directory is on smb:// mount

## Before

screenshot-2026-09-16_18-57-26-edited (1)

## After
screenshot-2026-09-16_19-38-52-edited

spandan11106:
Thanks for digging into this — the diagnosis and the approach are both right, and the reasoning about keeping the mirror path out of `Location` is exactly the distinction that matters here. I pushed one commit to your branch (`a15822f`) because there was a defect that stopped the fix from actually landing in the common case.

## What was wrong

The new branch passes `wait_for_metadata: true`. When mtime is unknown at bind time, the job parks in `METADATA_WAITERS` keyed on the mirror path, and the only thing that releases it is `note_metadata_entry` — which resolves its path through `local_thumbnail_path()`, and that is `None` for every URI location. So nothing ever released the park.

That is the default case, not an edge case: listings only request `time::modified` when `include_metadata` is set, which happens only when sorting by Size or Modified. With the default name sort, `entry_from_info` leaves mtime `Unknown`, so every remote thumbnail parked with nothing to wake it.

Measured on a real GVfs mount (an `archive://` mount — same FUSE mirror mechanics as `smb://`, no network needed), browsing a folder of 8 PNGs and not touching the window:

| build | requested | started | completed | applied |
|---|---|---|---|---|
| `main` | 0 | 0 | 0 | 0 |
| this PR as submitted | 8 | 0 | 0 | 0 |
| this PR + `a15822f` | 8 | 8 | 8 | 8 |

The middle row is the bug: entries get requested and counted eligible, then the jobs never start. I suspect your live check passed because scrolling recycles rows, and a re-bind after the metadata fill sees the filled mtime and proceeds — so the thumbnails do appear, but only once you scroll.

## What I changed

`note_metadata_entry` now resolves the mirror path too, so the fill releases the same path the bind parked under:

```rust
if let Some(path) = entry.local_thumbnail_path() {
    note_metadata(path, modified, file_size);
} else if let Some((mirror_path, _)) = remote_mirror_thumbnail(entry) {
    note_metadata(&mirror_path, modified, file_size);
}
```

I kept `wait_for_metadata: true` rather than flipping it to `false`. Flipping it also fixes the symptom, but it renders once with a `None` mtime (memory-only, unpersistable) and then again after the fill supplies a real key — two reads per image over the network. Parking and releasing properly gives one render with a cacheable key.

Also added `a_metadata_fill_releases_a_mirror_rendered_uri_entry`, which is the case the two existing tests cannot reach: both set `size`/`modified` to `Known(1)`, which skips the park entirely. The new test builds the entry with `Unknown` metadata, asserts it parks, then asserts the fill releases it. It fails on your commit and passes on mine.

## Things I checked that are fine

- The sandbox reads FUSE-backed files without changes — verified a `--ro-bind` of a file on a GVfs mirror under the same flags `sandbox.rs` uses, byte-identical output.
- Credential handling is not a problem: `sanitize_uri_credentials` rebuilds the URI with the username intact and strips only the password and auth params, and mount specs key on `user=`.
- Trash is unaffected: `gio info` reports no local path for `trash://`, so `remote_mirror_thumbnail` returns `None` there and the existing `thumbnail_path` route still owns it.
- Per-bind `gio::File::path()` cost: 0.004 ms once the mount is warm, 4.5 ms for the first call per mount. Resolving lazily per bind is the right call — doing it per entry at listing time would cost 0.48 ms each when no mirror exists, which is seconds of added load time on a large remote folder.

## Two small things left for you

1. The comments on the two existing tests read as though the `file://` URI stands in for a GVfs mirror. It stands in for the URI-routing mechanism only — nothing in either test touches gvfs, and `location_for_file` never produces a `file://` URI location in practice. Worth rewording so the next reader does not over-trust them.
2. Please drop the `## Tests` section listing `cargo check`/`fmt`/`clippy` — the PR template asks us not to list checks CI already runs. Keeping the note about the GTK suite is useful; I ran it here and `ui::thumbnail` is green at 48 tests, along with the trash, model, and metadata scopes.

Screenshots are still the one thing outstanding on the template. Whenever you have sanitized captures, drop them into the Visual evidence section.

Separately, and not your problem: the e2e harness pins `GIO_USE_VFS=local`, so no e2e scenario can exercise a GVfs path at all. I had to override it to run the live check above. I will open an issue for that.

spandan11106 (dismissed):

l0gicgate:
Reviewing this now. Will be in tonight's release. Thanks @spuder

l0gicgate:
Reviewed the full diff against upstream `main` at `fb282723f8c86b59ab13343b698ac0ab7778c508`, already included by the existing merge commit. Pushed `73300e33` to fix the CI compilation blocker (three fixtures missing `recent_unix_seconds`), construct escaped file URIs through GIO, and remove/tighten redundant comments. No production behavior changed during this review.

Reconciled [the earlier feedback](https://github.com/lgse/strata/pull/1080#issuecomment-5708222525): the metadata-waiter fix is present and its regression passes; fixture comments now explicitly distinguish URI routing from GVfs integration. Updated the description to the repository template, removed stale check claims, and embedded the author's existing before/after attachments. The separate GVfs E2E-harness limitation remains outside this PR.

Full local validation passed in a task-isolated rootless Podman store:
- `STRATA_CONTAINER_ENGINE=podman ./scripts/quality.sh all`: `cargo fmt --all --check`, Clippy with `--locked --all-targets --all-features -- -D warnings`, and tests with `--locked --all-targets --all-features`; **1,924 passed, 18 explicitly ignored**, with private Xvfb/D-Bus and required GTK execution.
- `STRATA_CONTAINER_ENGINE=podman ./scripts/e2e.sh`: **745 passed**.
- `git diff --check`: passed. All added comments audited.

The full runs were requested explicitly for this review. Live SMB/archive-mount manual verification was not repeated here; the existing author/reviewer evidence is linked above, and file:// unit fixtures do not establish live GVfs coverage. No new UI changes or captures were introduced.

Latest [CI run](https://github.com/lgse/strata/actions/runs/35284061643) is green. An initial unrelated archive-cancellation timeout passed its exact local rerun (1 test); failed-jobs-only CI retry could not find attempt-scoped artifacts, so the complete workflow was rerun successfully without code changes.

Approval submitted, but GitHub still reports `REVIEW_REQUIRED`: the ruleset requires approval from someone other than the last pusher (`require_last_push_approval`). A second maintainer must approve the current head. No merge performed.

l0gicgate (approved):
Reviewed the complete diff against fb282723, reconciled prior feedback, and audited added comments. Fixed the test-fixture compilation blocker in 73300e33. Full local quality and E2E suites pass; the latest CI run is green. Live GVfs manual verification was not repeated; existing author/reviewer evidence is retained in the description and discussion.

PR #1073, opened 2026-09-16
test(window): organize regression owners and dialog fixtures

l0gicgate:
## Description

Give window regressions explicit owners for keyboard policy, sidebar policy, bookmarks, devices and trash; move startup preview preference coverage into the existing preference owner. Keep live key-dispatch/type-to-search suites separate and localize five owner-specific helpers.

Preserve all 69 relocated cases and their assertions. Update the 11 GTK subprocess names alongside their module paths, and share the repeated empty device-dialog setup. No new production or harness changes relative to main; remaining complex scenario/assertion findings are retained explicitly. At the owner’s request, integrate current main and refresh seven reviewed visual baselines without changing comparison tolerances.

## Visual evidence

N/A — no new application UI or layout changes relative to main. Seven maintained visual-regression images are refreshed to its reviewed appearance; these executable baselines are visible in the file diff.

## How to test

1. Follow the window-test ownership map in `docs/e2e-testing.md` to locate policy, bookmark safety, device-action, trash and preference cases.
2. Inspect the relocated assertions and callback/cleanup ordering; only GTK dispatch names and the two equivalent dialog-fixture calls change within case bodies.
3. Confirm moved GTK cases execute inside their private child processes as well as reporting their outer result, rather than merely preserving the outer test count.

**Expected result:** All original window regressions remain executable (134 after preserving main’s added native-editing case) with clearer responsibility boundaries and no lost assertions or silently empty GTK children.

## Related issue

Addresses unit 27 of #763; keep the umbrella open. Completion requires merge and Cloud verification.

Follows the owner's request for the next bounded unit after #1065 merged. Main’s #187 native-editing regression is preserved unchanged in the keyboard-policy owner. Existing live keyboard suites touched by other feature PRs remain unchanged.

PR #187, opened 2026-09-03
feat(preview): render Markdown and HTML documents

kallupx:
## Description

Add native rendered previews for local Markdown and a bounded HTML subset. Documents can be switched between **Rendered** and **Source**, while **Render documents by default** controls the initial view of newly opened documents without persisting per-file choices.

Markdown is parsed with `pulldown-cmark`, while HTML is tokenized with `html5gum` and rendered using native GTK widgets. Supported content includes headings, paragraphs, formatting, nested lists, quotes, HTTP(S) links, code blocks, rules, and tables. Code blocks support theme-aware syntax highlighting and one-click copying.

Large rendered and source previews use recycled, virtualized rows so navigation, selection, and memory remain bounded. Cross-row selection, Ctrl+C, table copying, code copying, live theme changes, cancellation, and stale-request rejection are supported.

The HTML renderer does not use WebKit or Chromium, execute JavaScript, apply CSS, load subresources, or provide parser-initiated filesystem or network access. Unsafe and active content remains inert or is omitted with a visible warning.

## Visual evidence

https://github.com/user-attachments/assets/606f81b9-701f-4ef0-8f25-862719ea9f75

## How to test

1. Open local Markdown and HTML files containing headings, formatted text, nested lists, quotes, code blocks, links, rules, and tables.
2. Toggle between **Rendered** and **Source** using the preview action.
3. Change **Settings → General → Browsing → Render documents by default**, then open another document.
4. Drag across headings, paragraphs, lists, code blocks, and tables; press `Ctrl+C` and verify the copied text.
5. Use the copy button on highlighted and plain code blocks.
6. Activate HTTP(S) links and confirm unsupported schemes remain inert.
7. Browse large fixtures rapidly and scroll through their rendered and source views.
8. Open malformed, unsupported, oversized, deeply nested, and active-content fixtures.

**Expected result:** Supported local documents render with native, theme-aware styling. Selection and copying work across the preview, large files remain responsive, unsafe content stays inert, and unavailable rendering falls back to Source with a clear explanation.

## Verification

- [x] Static review of parser, renderer, GTK ownership, worker lifecycle, cancellation, links, gestures, and lazy view creation.
- [x] Security probes for NUL bytes, malformed markup, escaping, unsafe URI schemes, nested links, and unintended link activation.
- [x] Adversarial approximately 1 MiB documents exercising event, depth, markup, table, time, and cancellation limits.
- [x] Performance measurements in debug and release builds.
[...]

PR #1083, opened 2026-09-17
feat(recent): add recent files destination to sidebar

### LINKED ISSUE #1082: feat(recent): add recent files destination to sidebar
## Problem or use case

Finding a file that was recently opened or used can require navigating back through its original folder structure, even when the user remembers the file itself but not exactly where it is stored.

This is especially useful in workflows such as file uploads, where users often want to quickly find something they were working with recently.

This request originally came from a user on Discord:

> Having a Recent tab in the left vertical bar would be really useful, especially for file uploads. It makes finding recently used files much easier. I think some file managers already use a similar approach.

## Proposed solution

Add a **Recent** destination to the left sidebar that shows files from the desktop's existing recent-file history.

It should behave as a normal Strata location for browsing and selection, while actions on its entries should operate on the actual underlying files.

Recent files should be ordered by recent use by default, with the most recently used files first.

The feature should use the platform's existing recent-file infrastructure rather than maintaining a separate Strata-specific history.

## Alternatives considered

The current workaround is to navigate manually to the file's original directory or use search when the location is not immediately known.

Search can help when the filename is known, but it does not provide the same quick workflow as a dedicated list of recently used files.

## Visual examples

Other desktop file managers expose a similar **Recent** location as a built-in place in their navigation/sidebar.

Image

## Strata version

0.18.0

## Additional context

This feature request originated from the Strata Discord community.

The Recent destination should remain conceptually separate from Smart Folders or saved searches: Recent represents the desktop's existing recent-file history rather than a user-defined query.

guisilveira:
## Description

Adds a **Recent** destination to the sidebar, backed by the desktop's existing GTK/GIO recent-file history.

Recent behaves as a virtual collection while keeping each displayed entry backed by its real target file. This allows normal file actions to continue operating on the underlying files without treating `recent:///` as a writable directory.

This PR:

- adds Recent as a built-in sidebar destination when supported by the platform;
- resolves Recent entries to their real local or remote targets;
- sorts Recent by last-used time by default, without changing normal folder sort preferences;
- keeps normal item actions such as Open, Open With, Preview, Rename, Move, Trash, Delete, Properties, and Open File Location working on the real target;
- prevents directory-level operations such as New File, New Folder, Paste, and drop-into from targeting the Recent collection itself;
- records successfully opened files through GTK's existing recent-file infrastructure;
- monitors platform Recent history changes and refreshes the collection when it changes;
- adds a preference to show or hide Recent in the sidebar.

The implementation relies on the platform's existing Recent infrastructure rather than maintaining a separate Strata-specific history.

## Visual evidence

https://github.com/user-attachments/assets/98e25a4c-bdcb-48a3-bcfa-553ff0e7e3e1

## How to test

1. Launch Strata on a desktop where GTK Recent files and the `recent` GIO scheme are available.
2. Confirm **Recent** appears in the left sidebar and open it.
3. Verify recently used files are shown with the most recently used entries first.
4. Open a file from Recent, then return to Recent and verify its recency is updated.
5. Use normal item actions such as Open, Open With, Properties, Open File Location, Rename, Move, Copy, Trash, or Delete and confirm they operate on the underlying file.
6. Confirm collection-level actions such as New File, New Folder, Paste, and dropping files into Recent are unavailable or rejected.
7. Change the sort order while in Recent, navigate away and back, and confirm ordinary folder sort preferences are unaffected.
8. Add or remove an entry from the platform Recent history while Recent is open and confirm the view refreshes accordingly.
9. Disable **Show Recent** in Strata's sidebar preferences and confirm the destination disappears; re-enable it and confirm it returns.

**Expected result:** Recent behaves as a live, platform-backed collection of recently used files. Entries operate on their real targets, are ordered by recent use by default, and the virtual Recent root cannot be used as a normal writable directory.

## Related issue

Closes #1082


spandan11106:
Reviewed this and pushed six commits to the branch. Nice feature — the guards are placed at the action boundary rather than only in menu visibility, which is the right call and is why paste, drop, transfer and inline edit all stay correct. Five things needed fixing before this can go in.

**Every `recent` URI spelling now resolves to the collection.** `is_recent_root` matched the exact string `recent:///`, but this PR also adds `recent` to the location-bar scheme allowlist, and `location_from_input_with_home` does no path normalization. Typing `recent://` navigated successfully to a location where the check returned false, so enumeration fell through to the generic backend and published unresolved `recent:///<id>` children, and New File, New Folder, Paste, drop-into, Properties and Open With all lost their guards. I changed the root check to the same idiom `is_camera_photo_root` uses, and added `is_recent_location` for the operation guards so they cover the whole scheme the way `is_trash_location` already does.

**Recent registration no longer blocks the main loop.** `launch` called `query_file_type` on every opened file — a synchronous, uncancellable stat on the GTK main thread — so opening a file on a slow or unreachable mount froze the window after a successful launch. `open_location` had already resolved that type asynchronously moments earlier. The type query is queued now; registration is advisory and shouldn't delay the caller. An unreadable target still gets recorded, as before.

**One unreachable target no longer truncates the listing.** Targets resolved one awaited query at a time against a single load deadline, so a stale entry pointing at a dead mount spent the whole budget and every entry behind it in the batch was dropped. Batches resolve concurrently now, and entries beside a timed-out target are kept with the load reported as truncated.

**The Recent background menu had a trailing separator.** Customize and Properties are both withheld in Recent, but the separator above them was appended unconditionally, so the menu rendered a rule with nothing under it. Trash keeps Properties, so this was specific to Recent.

**Seven visual baselines were replaced with stale images.** This is the one worth a close look. The committed captures show the old `File` type descriptions and a different status bar — they came from a build predating #187. The application renders identically on this branch and on `main`; only the images changed. Confirmed by running the baselines in one worktree against both commits: `main` passes 9/9, this branch failed 7. Since Recent never appears in the E2E image (no recent backend there), these captures didn't need to change at all, so I restored them.

On testing, I added coverage for the paths the fixes touch, and for one gap that was already there: `RecentSourceError::Failed` was never exercised in either the open or the read branch, and it's the only path that surfaces an error message when the backend fails. Each fix has a test I checked fails against the previous code first.

There's also a new E2E scenario, with a caveat. The pinned image runs `GIO_USE_VFS=local` and ships no gvfs, so `recent:` genuinely doesn't resolve there and a populated collection can't be browsed end to end without changing pinned environment inputs. What it does cover is what that environment can actually prove: Recent stays out of the sidebar, neither preference state can add it back, and every URI spelling is refused with a clear report that leaves the session intact. The collection behaviour stays in the GTK tests.

Full local run on the merged result: `quality.sh fmt` and `clippy` clean, 1851 Rust tests passing, and the full `e2e.sh` suite at 737 passing with no failures.

Two smaller things I left alone, both your call:

- `CLOCK` sits before `CLIPBOARD_PASTE` in `src/assets.rs` and `data/strata.gresource.xml`, which breaks the alphabetical ordering in both.
- Recent has no Unpin context action, unlike Home, Trash and Network. `docs/preferences.md` already notes it, so this may well be deliberate.

Worth double-checking the baseline restore against what you intended — if those captures were regenerated on purpose, say so and I'll look again.

spandan11106:
Correction to my earlier comment, on the visual baselines — I had the cause wrong and want to set it straight.

This branch never touched those images. `git diff --name-only fba54cc 44fce8e -- tests/e2e/baselines` is empty. What actually happened is that #1073 regenerated all seven baselines on `main` after this branch was cut, so the branch still carries the pre-#1073 captures and they no longer match current rendering. Nothing was regenerated from a stale build here, and there was nothing careless about it — the branch is simply behind `main`. Apologies for the mischaracterization.

I traced it wrong by treating #1073's commit as this branch's base when comparing. The real merge base is fba54cc.

What that means for the commit I pushed: `aac63c9` sets those seven images to what `main` has, which is the same thing a merge from `main` will do, so it is an early partial merge rather than a fix. It does keep the branch green in the meantime — without it the suite fails those seven. When you merge `main` the images resolve identically and won't conflict, so you can leave the commit or drop it, whichever you prefer.

The other four fixes are unaffected by this.

Separately, the branch is now two commits behind: #1073 and #1075. Merging `main` conflicts in one file, `src/ui/window/tests.rs` — #1073 split that file into `tests/bookmarks.rs`, `devices.rs`, `keyboard_policy.rs`, `preferences.rs`, `sidebar_policy.rs` and `trash.rs`, and the `RecentAvailability` and `should_show_recent_place` coverage added here still sits in the old monolithic file. The sidebar availability tests most likely belong in `sidebar_policy.rs` now, but that is your call, so I have left the merge to you rather than relocating your tests on a guess.

Happy to do the merge and the test relocation if you would rather I take it.

spandan11106:
Merged `main` in and moved the sidebar tests to match the new layout.

The only conflict was `src/ui/window/tests.rs`. #1073 split it into per-area modules, so the five `should_show_recent_place` cases and the `recent:///` assertion in `sidebar_file_drops_accept_local_places_but_not_virtual_locations` now live in `tests/sidebar_policy.rs`, next to the existing standard-place and file-drop policy tests that own the same behaviour. Nothing was dropped in the resolution — the only additions this branch had made to that file were those five tests, the one assertion and their two imports.

While moving them I pulled the shared `RecentAvailability` construction into a small `recent_available()` helper, so each test now varies only the field it is actually about. Same coverage, less repetition.

The baselines resolved cleanly, as expected: the images this branch carried and the ones on `main` are now identical, so my earlier `aac63c9` has simply been absorbed by the merge.

Full local run on the merged result: `quality.sh fmt` and `clippy` clean, 1863 Rust tests passing, and the full `e2e.sh` suite at 738 passing with no failures.

The branch is now even with `main` at 823de6e.

l0gicgate:
Nice one @guisilveira, did you make sure you can pin/unpin recent as well in the settings (haven't pulled down the PR yet). Make sure you can.

spandan11106:
> Nice one @guisilveira, did you make sure you can pin/unpin recent as well in the settings (haven't pulled down the PR yet). Make sure you can.

Yes we can.

wmfeht:
The Recent production path matches #1082: target-backed entries, collection-level New/Paste/drop refused on the whole `recent:` scheme (not only `recent:///`), recency sort stays local, Show Recent live-binds through `sidebar_places_visibility`, and `strata-clock` is a Lucide icon rendered with `primary_icon`.

**Requesting changes for test quality**, not missing coverage.

`src/ui/window/tests/sidebar_policy.rs` adds five tests that each assert one boolean of `should_show_recent_place`, which is `show_recent && !local_only && availability.is_available()`. That is a combinatorial matrix of a three-line AND — it restates the implementation and cannot fail unless the helper is rewritten. The same file already tests the similar `should_show_standard_place` helper in one function. The real hide/show contract is already covered by `sidebar_visibility_prefs_hide_and_restore_default_places_across_windows` and `recent_place_appears_by_default_when_runtime_is_available`.

Drop those five, or fold the independent false axes into one test. Do not add more tests for this helper.

Same class: `recent_watch_uses_gio_monitoring_when_supported` in `src/adapters/local_files/tests.rs`. If GIO cannot monitor `recent:///`, it asserts nothing; if it can, it only checks `handle.is_some()`. The rescan policy test already covers the monitor contract.

l0gicgate:
Merged upstream `main` at `d39052be9cd9f85956d281eccc7c2383e02106fa` without conflicts and reviewed the resulting diff.

Addressed [the test-quality feedback](https://github.com/lgse/strata/pull/1083#issuecomment-5710087941): consolidated the five sidebar policy cases into one test retaining each independent condition, and removed the conditional monitor smoke test. Monitor event/rescan and browser reload coverage remain, as do startup and two-window live-preference tests. Audited every added comment, removed narration, and shortened retained rationale; also restored two existing comments to the code they describe.

Verified the earlier URI guards, asynchronous open registration, concurrent target resolution, and menu separator fixes in the current code. Baselines match updated main, consistent with the correction in the discussion. Settings supports hiding/restoring Recent; a sidebar Unpin action and icon ordering remain non-blocking polish, not scope added by this review.

Local verification: pinned `quality.sh fmt`, `quality.sh clippy`, and `quality.sh test` passed (1,921 tests, 18 explicitly ignored); canonical rootless-Podman `e2e.sh` passed all 745 cases. These run the requested Cargo checks with private GUI/session-bus isolation. No additional UI behavior changed, so no new visual capture is required. Manual populated-Recent browsing was not performed: the pinned environment lacks GVfs; its E2E cases verify unavailable-backend behavior, while injected Rust fixtures cover collection behavior. Existing PR visual evidence is unchanged.

Pushed merge `d083088b` and review cleanup `48870fc4`; all post-push CI checks are now green ([run](https://github.com/lgse/strata/actions/runs/35192416909)). The PR remains unmerged.

PR #1086, opened 2026-09-17
feat(preview): render RTF and Word documents

### LINKED ISSUE #730: feat(preview): render RTF and DOCX documents
### Problem or use case

Strata's quick preview shows RTF and DOCX files as raw/source text, which is not meaningful for DOCX (a ZIP-based binary format) and loses all formatting for RTF. Part of the broader tabular/office preview work tracked in #244; this issue scopes the RTF/DOCX portion of it.

### Proposed solution

Both formats are rich-text documents (paragraphs, headings, emphasis, lists, tables, images) rather than tabular data, so they don't fit the table-grid preview already shipped for CSV/TSV/Excel/ODS (#723). Instead, they share a rendering shape with each other: parse into a structured paragraph/block document model and render through a rich-text document view.

- Parse RTF with `rtf-parser` (MIT, mature, zero external dependencies).
- Parse DOCX with `docx-rs` (MIT, mature, structured read path covering paragraphs, runs/formatting, tables, images, and styles).
- Render both through the same native rich-document view — paragraphs, headings, emphasis, lists, tables, and images with Strata's theme tokens applied — reusing the paragraph/block document model and virtualized rich-text view PR #187 is building for Markdown/HTML previews (issue #89), rather than building a separate one-off renderer.

**This issue depends on #187 merging first**, since it introduces the document model and view that RTF/DOCX parsing would feed into.

### Alternatives considered

- Continue showing RTF/DOCX as plain text or no preview (status quo).
- Building a standalone renderer instead of reusing #187's document model — rejected, would duplicate a large amount of already-reviewed rendering infrastructure.

### Visual examples

N/A — follows the rendered-document styling introduced by #187 once that lands.

### Strata version

Current `main`

### Additional context

PPTX/PPT/ODP (presentations) and ODT are intentionally excluded here: presentation formats need a slide-based view rather than a document view, the PPTX/ODP Rust library (`pptx-to-md`) is under 1.5 years old and unproven, legacy PPT/DOC have no viable pure-Rust reader, and ODT has no structured-reader library at all. These remain tracked under the parent epic #244 without a concrete plan yet.

DOCX has substantially richer feature coverage than RTF (images, complex tables, styles), so even under one issue, implementation will likely land as separate commits/PRs per format.

spandan11106:
## Description

Quick preview renders RTF and Word (DOCX) documents instead of showing raw source or nothing at all. Both feed the paragraph/block document model and virtualized rich-text view already used for Markdown, HTML, CSV/TSV and workbooks, so they inherit its theming, selection, copy, sorting and bounded-parsing behaviour.

- **RTF** is converted in process to the bounded HTML subset the rendered view already parses: paragraphs, line breaks, bold, italic, underline, strikethrough, Windows-1252 `\'hh` escapes and `\uN` escapes with their `\ucN` fallback characters. Font tables, stylesheets, pictures, headers, footers and other non-body destination groups are discarded. It keeps its **Source** view like Markdown and HTML.
- **DOCX** is read with `docx-rs` inside the existing preview sandbox, under the same 20 MiB input limit as workbooks, and returns bounded JSON that the application reparses with the existing HTML parser. Headings, `Title`/`Subtitle`, quote styles, inline formatting, tabs, breaks, bulleted and numbered lists with nesting, and tables all render; the first table row becomes the sortable column titles. Being a ZIP container, it has no Source view — the same as XLSX/ODS.
- Underline (`u`, `ins`) is now part of the supported HTML subset. The Pango layer already had the span style but the parser dropped it, so underlined text lost its formatting and raised a misleading "unsupported or active HTML content was omitted" notice.
- `PreviewContent::Workbook` is renamed to `Rendered`, since spreadsheets and Word documents now share it.

The RTF converter is hand written rather than using `rtf-parser` as the issue proposed. That crate (0.4.3) drops `\par` entirely, so every paragraph runs together; decodes `\'93` to U+0093 instead of a curly quote; and turns `舒?` into an unrelated CJK character while leaving the `?` fallback in place. DOCX uses `docx-rs` as planned.

Known limits, documented in `docs/document-previews.md`: DOCX drops images, hyperlink targets, headers, footers, footnotes and comments, joins each cell's paragraphs into one value, and flattens nested tables into their containing cell; RTF renders paragraphs and inline formatting only, so list markers and table cells become plain text, code pages other than Windows-1252 can mis-decode `\'hh`, and files over the 1 MiB preview limit fall back to Source.

## Visual evidence

Pending — a short video of both previews will be attached in a follow-up comment below.

## How to test

1. Create a folder with a `.docx` containing headings, bold/italic/underlined text, a bulleted list, a numbered list and a table, plus a `.rtf` with several paragraphs, styled runs, curly quotes and an em dash. Any Word or WordPad/LibreOffice export works.
2. Browse to that folder in Strata, select the `.docx` and press <kbd>Space</kbd>.
3. Click a column title in the table to sort it, then drag across a few cells and press <kbd>Ctrl</kbd>+<kbd>C</kbd>.
4. Select the `.rtf` and press <kbd>Space</kbd>, then use **View source** in the preview header and switch back to the rendered view.
5. Switch theme while both previews are open.

**Expected result:** the DOCX preview shows a styled document — headings, inline formatting, nested bullets, numbered items and a sortable table whose first row is the column titles — with no **View source** control and no omission notice; the copied cells arrive tab separated. The RTF preview shows one block per paragraph with its bold, italic, underlined and struck runs, curly quotes and dashes intact, and toggles between **Rendered** and **Source**. Both follow the active theme live. Files that cannot be parsed fall back to the source view or "Preview unavailable" rather than failing the pane.

## Related issue

Closes #730


spandan11106:
https://github.com/user-attachments/assets/a0491401-a49b-47e0-b5fe-1d1ea7f8d9e8

wmfeht:
**request-changes** on `3317092a` (RTF/DOCX previews)

DOCX sandboxing, HTML reparse, the underline subset, and `PreviewContent::Rendered` look right, and the new tests are focused rather than padded. One production hole in the in-process RTF walker.

**Blocker: `to_html` can panic on `\bin`.**

`src/services/rtf.rs` walks `source.as_bytes()` but emits body text with `writer.text(&source[index..end], …)`. That `&str` slice requires both ends to be UTF-8 character boundaries.

`\binN` advances by `N` raw bytes (`next.saturating_add(length).min(bytes.len())`) and is not gated on destination skip. After that skip, `index` can sit in the middle of a multi-byte character. The next non-delimiter byte then panics (`byte index X is not a char boundary`). `gio::spawn_blocking` catches the unwind, so the pane shows “The preview worker stopped unexpectedly” instead of falling back to Source — still a failed preview, and it runs in the application, not the sandbox.

Minimal shape: `{\rtf1\ansi\bin1 é\par}` (`é` is `C3 A9`; `\bin1` skips `C3`). The same hole hits UTF-8 RTF after a picture `\bin`, which the comment says this control word exists to skip.

Related: the walker sees `from_utf8_lossy` output, so `\bin` lengths are decoded-string bytes, not original file bytes. Picture payloads already desync grouping even when they don’t panic.

`skip_fallback` already eats UTF-8 continuation bytes. `\bin` needs the same rule: never slice `str` off a boundary (decode chunks from `&[u8]`, or snap to `is_char_boundary` and treat binary as non-text).

spandan11106:
Confirmed and fixed in 3335744. Thanks — the reproduction was exact.

`{\rtf1\ansi\bin1 ékept\par}` against the previous walker:

```
panicked at src/services/rtf.rs:162:40:
start byte index 18 is not a char boundary; it is inside 'é' (bytes 17..19 of string)
```

Both halves of the diagnosis hold. `\binN` advanced by N source bytes with no boundary rule, and the preview source is `String::from_utf8_lossy` output from `decode_text_sample`, so those lengths are decoded-string bytes rather than original file bytes.

Two changes:

- `\bin` now skips N *characters* through a shared `next_character` helper, which `skip_fallback` already used for `\uN` fallbacks. Since a byte that failed to decode has already become exactly one replacement character, counting characters both stays on a boundary and tracks the original payload more closely than counting decoded bytes did.
- Body text is emitted with `String::from_utf8_lossy(&bytes[index..end])` instead of slicing `source`, so no stored length in any control word can panic the walk, including any arithmetic added later.

Regression test `binary_payload_lengths_skip_whole_characters` covers your minimal shape and a payload longer than the remaining document. It panics with the same message on the parent commit and passes here.

You are right that a picture payload can still desync grouping — a decoded length is an approximation of a binary one, so a stray `{` or `}` inside the payload can still be read as structure. That now costs a garbled or empty rendered view falling back to Source rather than a failed pane, and it is bounded by the same parser budgets. Reading RTF from the original bytes rather than a lossy decode would remove the approximation entirely; that means changing how the preview loader hands documents to the parser, which reaches past this issue, so I would rather do it separately than widen this one.

l0gicgate:
Merged verified upstream `main` (`62930eb7feab0f64c0cd663f736524dbbfeab227`) without rebasing. Fixed the DOCX sandbox regression test’s missing Bubblewrap-path argument, which caused the existing CI compilation failure, and audited every added comment, removing redundant descriptions and tightening safety rationale. No UI behavior changes in this review.

Rechecked the [RTF binary-skip concern](https://github.com/lgse/strata/pull/1086#issuecomment-5714297092): the existing byte-slice decoding and character-boundary regression prevent the reported panic; that regression passed. The [acknowledged lossy binary-length approximation](https://github.com/lgse/strata/pull/1086#issuecomment-5714491777) remains a non-blocking fidelity limitation, not a claim of byte-exact RTF parsing. There were no submitted reviews or inline threads to reconcile.

Local pinned-container verification: formatting and Clippy passed; full Rust suite 2,039 passed / 18 ignored; full canonical E2E suite 772 passed. Inspected fresh DOCX and RTF captures in a private display and exercised RTF Source → Rendered switching successfully. Fresh captures remain local because attachment-upload tooling is unavailable; the author’s existing video is in the earlier comment. All required GitHub checks are now green at `6a0986bc`.

Merge remains blocked by the repository rule requiring approval from someone other than the last pusher. My approval is recorded, but another eligible reviewer must approve this head; no bypass or merge was attempted. The target branch also requires signed commits; these local review commits are unsigned, so the eventual merge must use a rule-compliant signed integration (rather than assuming the current history can be merged unchanged).

l0gicgate (dismissed):
Reviewed the complete diff against merged upstream main, reconciled the RTF panic feedback, and audited added comments. The sandbox test compilation blocker is fixed; local pinned quality/E2E suites and all required PR checks pass at 6a0986bca16d1d2233a40537eceefbe73c929fa8. No remaining merge-blocking findings.

l0gicgate:
Guided testing found an unresolved DOCX compatibility issue: a valid document using `ns0:` bound to the WordprocessingML namespace instead of the conventional `w:` loses its text while retaining empty table structure. `docx-rs` 0.4.22 `reader/run.rs` dispatches on the literal `Some("w")` prefix. My generated long-report sample exposed this; changing only its namespace prefix back to `w:` restores headings and body paragraphs in the isolated GUI check. This is not a document-length failure. The demo has been corrected for continued scrolling/selection testing, but application support for alternate prefixes has NOT been fixed. The earlier no-blocking-findings review predates this discovery.

l0gicgate:
Fixed the namespace-prefix regression in 396e2b0c. Inside the existing sandbox, package XML is resolved by namespace URI and normalized before reaching docx-rs. This supports alternate/default namespaces without blindly replacing text; scoped rebinding remains distinct. Original and normalized XML totals are bounded to 64 MiB, and DTDs are rejected.

Regression coverage verifies equivalent rendered content/formatting for alternate and default prefixes, prefix rebinding, malformed namespaces/DTD rejection, and aggregate XML budgets with binary-part preservation. Full pinned quality passed (2,152 Rust tests, 20 ignored), as did all 774 E2E tests. After tightening the aggregate output budget, reran all 9 DOCX tests, the 4 office-preview E2E cases, formatting and Clippy successfully.

The ORIGINAL failing ns0 long report now renders headings, formatted text, lists, table cells and numbered body paragraphs in the rebuilt GUI; a separate isolated GUI check passed. The unchanged failing sample is now available for owner-guided testing as 05-alternate-prefix-report.docx. Screenshot captured and inspected; upload tooling is unavailable, so new visual evidence is pending upload (local path: /home/l0gicgate/Documents/github/strata-worktrees/730-review-pr1086/target/review-media/long-docx.png). This supersedes the unresolved compatibility finding in my earlier comment.

PR #1090, opened 2026-09-17
feat(preview): add archive Quick Look

### LINKED ISSUE #1151: feat(preview): add metadata-only archive Quick Look
### Problem or use case

Inspect archive contents from Quick Look without extracting files or leaving the file browser. PR #1090 implements this feature but has no linked tracking issue.

### Proposed solution

Provide a read-only, navigable member tree for local ZIP, 7z, TAR, and TAR.GZ/TGZ archives. Parse in the preview sandbox with cancellation and resource limits. Prompt for encrypted ZIPs and encrypted 7z headers, keep passwords out of paths/arguments/diagnostics, and display stable errors for malformed or unsupported input.

Readable 7z metadata may be listed without decrypting encrypted file contents. Preserve literal dot/backslash components and file/directory collisions; reject ZIPs whose duplicate names cannot be faithfully enumerated by the current dependency rather than silently omitting members.

### Alternatives considered

Extracting to a temporary directory or launching an external archive manager. Neither provides a metadata-only Quick Look experience.

### Visual examples

The proposed interface is demonstrated in the Visual evidence section of PR #1090.

### Strata version

0.19.0 source development; PR #1090.

### Additional context

Tracks https://github.com/lgse/strata/pull/1090 and its existing review discussion. No archive extraction behavior changes are requested here.

fadilasif:
## Description

Add Archive Quick Look support for ZIP, 7z, TAR, and TAR.GZ/TGZ archives. Archives can be previewed as a directory tree without extraction, with password handling, malformed-archive detection, sandboxed listing, cancellation, and resource limits for hostile archive names/nesting.

Also adds resource-limit protections to prevent pathological archives from causing excessive memory usage or UI stalls during tree construction.

Quick Look lists metadata only: 7z archives with readable headers do not require a password even when their file contents are encrypted. ZIPs with duplicate member names are rejected as unsupported because the current ZIP dependency otherwise silently collapses them.

Closes #1151.

## Visual evidence

https://github.com/user-attachments/assets/2c075331-d9f2-4f62-851e-d80b87e598bd

## How to test

1.Open a supported .zip, .7z, .tar, .tar.gz, or .tgz archive in Strata.
2.Use Quick Look to preview the archive and navigate through the displayed directory tree.
3.Test a password-protected archive with the correct and incorrect password.
4.Test malformed/truncated archives and an archive with very large/deeply nested names.

**Expected result:**

Supported archives display their contents without extraction. Encrypted ZIPs and encrypted 7z headers prompt for a password and reject incorrect passwords. Malformed/unsupported or excessively complex archives fail safely with an appropriate error instead of hanging or exhausting resources.


guisilveira:
I reproduced the archive-preview findings against the current PR head (`b49451ca09119c1dab6dc1715b80dabecc746fd6`) and confirmed all four.

The tests were run from a release build in an isolated GTK/Xvfb/D-Bus environment, using the current PR code without production changes.

### 1. [Blocking] Wide archives can make the preview practically unusable for tens of seconds

This is reproducible with completely valid, flat ZIP archives containing short filenames and staying within the current resource limits.

Measured time until the preview became usable:

| Entries | Time |
| ---: | ---: |
| 1,000 | 252 ms |
| 5,000 | 2.86 s |
| 10,000 | 9.92 s |
| 20,000 | 43.81 s |
| 20,000 — repeated run | 42.12 s |

The 20k archive was reproduced twice with essentially the same result.

I did **not** observe a total application freeze, so I would describe this more precisely as prolonged synchronous latency / practical preview unavailability rather than an absolute freeze.

The video shows the preview still on the spinner after both 1 second and 20 seconds, with the 20,000 rows only appearing at the end.

This matches the implementation path where the accepted archive can contain up to 20,000 entries, then the archive tree is built and the `GtkListBox` rows are materialized synchronously.

Relevant code:

- `src/services/preview.rs`
- `src/ui/preview/archive.rs`
- `src/ui/preview.rs`

Given that the PR explicitly adds resource limits to prevent pathological archives from causing UI stalls, I think this needs to be addressed before merge.

---

### 2. [Major] Some malformed archives expose parser/dependency errors directly in the UI

Also reproduced through the real Quick Look flow.

Observed messages:

- truncated TAR: `This file is not a valid archive or is damaged.`
- truncated TAR.GZ: `incomplete deflate stream`
- invalid TAR inside GZIP: `failed to read entire block`
- invalid GZIP: `invalid gzip header`
- truncated 7z: `Other("Cannot handle next_header_size 113")`

So plain TAR already has stable product-level handling, but TAR.GZ and 7z can still expose internal/dependency-specific wording.

Relevant code:

- `src/adapters/local_operations/archive/listing.rs`
- `src/ui/preview.rs`

I think these should be normalized at the adapter boundary into stable product errors rather than making dependency strings part of the user-facing contract.

---

### 3. [Major] Content-only encrypted 7z archives are listed without a password prompt

I created and independently verified two 7z fixtures using `sevenz-rust2 0.22.2`:

**A — encrypted header / filenames**

- `bsdtar` cannot list the archive without the password.
- Strata correctly shows `Password-protected archive`.
- No archive rows are exposed.

**B — readable header, encrypted file contents**

- `bsdtar` can list `protected.txt`.
- `bsdtar` cannot read the file contents without the password.
- Strata opens the tree and displays `protected.txt` immediately.
- No password prompt is shown.

So this is specifically a **content-only encryption** case. The metadata being readable does not mean the file contents are decrypted or accessible.

Relevant code:

- `src/adapters/local_operations/archive/listing.rs`
- `src/adapters/local_preview.rs`

If listing encrypted-content archives without a password is intentional for Quick Look, I think the PR description/contract should be narrowed accordingly. Otherwise, this needs additional detection/state handling.

---

### 4. [Major] Tree normalization loses archive member identity

I reproduced this with an archive containing these raw members:

```text
../../etc/passwd
./normal.txt
dir/../hidden.txt
a\b.txt
x
x/child.txt
```

The resulting preview changes their presentation:

- `a\b.txt` becomes directory `a` → `b.txt`
- `dir/../hidden.txt` becomes `dir/hidden.txt`
- `../../etc/passwd` becomes `etc/passwd`
- `x` and `x/child.txt` can produce two visually indistinguishable `x` rows
- the `x/` + `x/child.txt` variant produces only one `x` row
- duplicate `duplicate.txt` members collapse into one row

The collision behavior also changes depending on member insertion/order.

This is **not a path-traversal or extraction vulnerability** — nothing here is being extracted. The issue is that Quick Look no longer faithfully represents the member identities actually stored in the archive, which is particularly misleading when inspecting suspicious archives.

Relevant code:

- `src/services/preview.rs`
- `src/ui/preview/archive.rs`

I think the virtual hierarchy can normalize navigation structure if needed, but the original member identity should not be silently erased or merged.

---

### Reproduction summary

| Finding | Result | Confidence |
| --- | --- | --- |
| Wide archive responsiveness | Reproduced | High |
| Internal parser errors | Reproduced | High |
| Content-only encrypted 7z | Reproduced | High |
| Archive member identity normalization | Reproduced | High |

The strongest blocker for me remains the wide-archive case: a valid 20,000-entry archive takes ~42–44 seconds to become usable on repeated runs, despite being within the feature's own accepted limits.

The other three are also reproducible, but they are more localized contract/presentation issues and should be straightforward to reason about independently.

fadilasif:
Thanks for your input. Ill surely look into the mentioned issues.

guisilveira:
**[Major] Archive passwords are persisted as plaintext temporary files and can survive abnormal termination**

The password-handling comments currently imply that the submitted password is never written to disk, but the implementation stages it as a real filesystem file through `stage_secret_file()`.

I reproduced the full lifecycle against the current PR head using an isolated `TMPDIR` and a synthetic password.

Observed behavior:

```text
directory mode: 0700
password file mode: 0600
password stored verbatim: yes
normal completion cleanup: yes
password present in argv: no
password present in environment: no
password present in logs/stdout/stderr: no
unrelated unprivileged user could read it: no
```

So the existing Unix permissions are doing their job, and this is **not** a world-readable credential leak or a sandbox escape.

However, cleanup currently relies on `PrivateOutput::Drop`. When I repeated the test and sent `SIGKILL` only to the isolated Strata parent after the password file had been staged, the process could not run its `Drop` cleanup and both the temporary directory and plaintext password file remained afterward.

Example residual path:

```text
.../strata-preview-45530-2/password
```

The password contents were still present verbatim after process termination.

That makes the current statement that the password is "never ... written to disk" objectively inaccurate, and it means secret lifetime depends on graceful process cleanup.

I think this should either use a non-filesystem secret transport such as an inherited FD/pipe/memfd, or the security contract should at minimum be updated to accurately describe the temporary-file behavior and its lifecycle.

**Reproduction:** current head `b49451ca09119c1dab6dc1715b80dabecc746fd6`; normal cleanup succeeds, `SIGKILL` leaves the mode-0600 plaintext secret behind.

guisilveira:
**[Minor / Security hardening] `PreviewRequest` exposes the archive password through its default `Debug` implementation**

`PreviewRequest` now contains:

```rust
pub archive_password: Option<String>,
```

while still deriving:

```rust
#[derive(Clone, Debug)]
```

I reproduced this against the current PR head using the real `PreviewRequest` type and a synthetic password. Its normal `Debug` representation included the password verbatim:

```text
archive_password: Some("strata-security-repro-password")
```

I also searched the current codebase and did **not** find an active logging path for `PreviewRequest` or `archive_password`, so this is not evidence that passwords are currently being leaked to logs.

The issue is that a secret-bearing type is currently safe to pass to ordinary `{:?}`, `dbg!`, tracing, or future diagnostic code without any redaction. That makes the field comment saying the password is "never logged" dependent on every future caller remembering not to debug the request.

I would avoid exposing the secret through the default `Debug` implementation. A custom `Debug` implementation that redacts `archive_password` would be enough here, e.g. showing it as `[REDACTED]` or only whether a password is present.

**Reproduction:** current head `b49451ca09119c1dab6dc1715b80dabecc746fd6`, synthetic password only, no active runtime log leak observed.

fadilasif:
Thanks for the detailed review @guisilveira . I went through all the findings and performed a deeper follow up audit of the current implementation.

### Changes / verification

1. **Wide archive performance**
   - Verified that archive listing/tree construction runs off the UI thread with `gio::spawn_blocking`.
   - Verified the archive browser uses a virtualized `ListView`.
   - Verified tree construction uses indexed maps to avoid O(n²) behavior.
   - Resource limits and cancellation remain in place.
   - Re-tested 20k-entry and adversarial archives; the previous ~40s preview delay is no longer reproducible.

2. **Malformed archive errors**
   - Verified that TAR/TGZ/ZIP/7z parser errors are normalized at the listing boundary.
   - Dependency-specific errors no longer leak into the user-facing UI.
   - Valid archives and cancellation behavior were regression-tested.

3. **Archive member identity**
   - Verified that `.` / `..` and backslash handling preserve archive member identity.
   - Duplicate members and file/directory collisions remain represented.
   - Ordering does not change the resulting tree.
   - Existing empty-component normalization remains intentional, documented, and tested.

4. **Password lifecycle**
   - Verified that passwords are no longer persisted as plaintext temporary files.
   - Passwords are handled through an anonymous file descriptor and are not exposed through arguments, environment variables, logs, or filesystem paths.
   - Verified retries, cancellation, abnormal termination, invalid FDs, special-character passwords, and encrypted archive flows.
   - No password residue remains after termination.

5. **Password Debug exposure**
   - Verified that archive passwords use `SecretString` with redacted `Debug` output.
   - Normal, pretty, and nested Debug output were checked for secret exposure.

6. **Content-only encrypted 7z behavior**
   - Investigated the case where 7z file contents are encrypted but the archive headers/filenames remain readable.
   - This is **expected behavior for the current metadata-only Quick Look implementation**, not a password-handling bug.
   - Quick Look only lists readable archive metadata; it does not decrypt or read protected file contents, so a password prompt is not required for listing in this case.

7. **Sandbox**
   - Audited the archive-preview sandbox path, including helper execution, FD handling, limits, cancellation, and output handling.
   - **No sandbox regression was found.**

8. **Final cleanup**
   - Audited stale/orphan code, redundant tests, comments, error paths, security exposure, and repository consistency.
   - Corrected one documentation comment in `listing.rs`; this was the only behavior-neutral cleanup change.

### Validation

- Targeted archive, sandbox, password, preview, helper, secret, and cancellation tests: passing
- `cargo fmt --all --check`: passing
- `cargo clippy --locked --all-targets --all-features -- -D warnings`: passing
- Final diff/conflict checks: clean
- An earlier full-suite validation completed with 2030 passed, 0 failed, and 20 ignored; the final audit re-ran the targeted suites rather than the full suite.
- Archive Quick Look functionality and security fixes remain intact after rebase and cleanup.

l0gicgate:
Reviewed against upstream `main` at `f13d9f291b58f2831b2a3cf2959a848f561e1ffc` and merged it without conflicts. Linked the feature to #1151.

Fixed during review:
- A delayed rename timer could steal password-field focus and rename the archive with part of the entered password. Editable controls now retain ownership; real ZIP/7z sandbox retry tests cover the flow.
- Replaced globally inheritable temporary descriptors with close-on-exec memfds passed only through the intended helper's stdin; bounded secret reads and tested child isolation.
- Moved tree construction off the UI thread. ZIP duplicate names were still silently lost by the dependency; these archives now report unsupported instead of presenting an incomplete listing.
- Restored unsupported-selection coverage using RAR fixtures, completed the added-comment audit, and corrected two spelling-policy failures without weakening checks.

Feedback disposition:
- [Performance, parser errors, encryption, and member identity](https://github.com/lgse/strata/pull/1090#issuecomment-5719433700): verified normalized parser errors and literal dot/backslash/file-directory handling; fixed the remaining tree-thread and ZIP duplicate issues. Readable content-encrypted 7z headers intentionally remain previewable without decrypting file contents; the PR description now states that limitation.
- [Password residue](https://github.com/lgse/strata/pull/1090#issuecomment-5720114136) and [Debug exposure](https://github.com/lgse/strata/pull/1090#issuecomment-5720119102): anonymous secret transport and `SecretString` redaction verified, with the inheritance fix above.

Manual headless verification: wrong/correct passwords for ZIP and encrypted-header 7z; folder/breadcrumb navigation; literal member names; no extraction. A valid flat 20,000-entry ZIP became visible in ~0.34 seconds on this machine (informational, not a performance guarantee).

All four required checks are green on `0bce3eb1` ([CI run](https://github.com/lgse/strata/actions/runs/35466471421)). GitHub still requires an approving review from someone other than the last pusher; this review did not bypass that rule or merge the PR.

**Updated visual evidence pending owner upload:** captured and inspected before/after password-focus, unlocked archive, literal-member, and 20k-entry screenshots. This session has no supported authenticated GitHub attachment uploader; no new media has been attached. Local capture paths are provided in the task handoff.

l0gicgate:
## Guided owner testing and follow-up fixes

Pushed `2f2d3229` with the two visual fixes found during our guided session:
- Archive content and breadcrumb scrollbars now use the same fixed, non-overlay styling as CSV and other previews.
- Text/code preview line-number gutters now use the preview separator's semantic border color on the outer divider, with the inner gutter border removed. The initial color-only attempt did not fix the visible line; the revised build was retested and explicitly confirmed correct by the owner.

### Method

Created synthetic fixtures in `~/Documents/demo/compressed-files` and launched this PR's container-built app on the owner's desktop, explicitly requested for interactive testing. Used a separate D-Bus session and isolated application settings; the existing portal service was left untouched. Selected files and pressed **Space** for Quick Look rather than double-clicking/extracting them. The owner exercised each group and reported the observed results; screenshots were used to diagnose the two visual discrepancies. Rebuilt and reopened the app for follow-up visual checks.

### Owner-confirmed results

| Cases | Result |
| --- | --- |
| ZIP, TAR, TAR.GZ | Folder navigation (`docs → nested`), member listing, and Contents breadcrumb worked without extraction. |
| Password-protected ZIP and header-encrypted 7z | Wrong password rejected; correct password listed `protected.txt`. No focus theft or archive filename changes. |
| Content-only encrypted 7z | Readable metadata listed without a password prompt, as intended. |
| Literal member names and file/directory collisions | Dot/backslash names and distinct `x` / `x/` entries behaved as described. |
| Duplicate-name ZIP | Reported unsupported rather than silently hiding a member. |
| Damaged ZIP/TAR.GZ and truncated 7z | Expected errors, without crashes or raw parser details. |
| 20,000-entry ZIP | Listing and scrolling remained responsive. The mismatched scrollbar found here was fixed and visually confirmed. |
| 20,001-entry ZIP and excessive-depth TAR | Reported “Archive too large to preview.” |
| Empty ZIP / plain gzip stream | Empty listing / no archive-tree preview, respectively. |
| Text preview gutter | Owner confirmed the final divider matches the pane borders. |

### Validation and evidence

On the final code, pinned rootless-Podman formatting and Clippy checks passed, as did spelling/diff checks and **10 targeted canonical E2E cases**:

```sh
./scripts/quality.sh fmt
./scripts/quality.sh clippy
./scripts/e2e.sh tests/e2e/scenarios/test_archive_preview.py tests/e2e/scenarios/test_quick_preview.py -k 'archive_password or space_opens_and_closes or large_table_header_sort'
typos .
git diff --check
```

Container commands used `STRATA_CONTAINER_ENGINE=podman`; automated GUI checks used private Xvfb/D-Bus, not the desktop. These bounded follow-up changes reuse existing styling and only affect archive scrollers and text-preview borders, so the full local Rust/E2E suites were not repeated; the prior full-run results remain documented above. Visual appearance was checked by the owner, not by adding layout assertions.

Updated inline visual evidence remains pending upload: screenshots are retained locally, but this session has no supported authenticated GitHub attachment uploader. Owner confirmation is recorded here; no new image attachment is claimed. PR remains unmerged.

PR #1093, opened 2026-09-17
feat(terminal): follow browser navigation in the embedded terminal

spandan11106:
## Description

Stacked on `feat/93-embedded-terminal` (lgse/strata#1089). Only the second commit belongs to this change; review the diff against that base.

Second of the three steps on #93: the embedded terminal now follows the folder the browser is showing.

- The browser stays the source of truth. Navigation is observed through the existing observer; nothing here ever moves the browser.
- A directory change is injected only when the shell itself owns the terminal **and** the user has left no half-typed line at the prompt. `tcgetpgrp` on the PTY reports the first. The second is invisible to `tcgetpgrp`, so VTE's `commit` signal tracks whether the current line was submitted or discarded.
- While neither holds, the newest requested folder is held and retried on a timer until the shell is free. Navigating past several folders while a build or an editor runs applies only the last of them, once.
- Paths are single-quote escaped, so spaces, quotes and shell metacharacters in folder names are inert.
- Locations with no local path, trash included, pause synchronisation rather than inventing a path.
- The panel now tracks the location in the breadcrumb rather than the hovered column, so it matches what the window says it is showing. `Ctrl+T` is unchanged and still acts on a selected folder.

Still one-way: nothing reads the shell's own `cd` back into the browser.

## Visual evidence

Pending — a short video showing navigation following, a busy shell being left alone, and the deferred change landing afterwards will follow in a comment.

## How to test

1. Open a folder, press `F4`, run `pwd`.
2. Click into a subfolder in the browser, then run `pwd` in the panel again — it followed.
3. Run `sleep 15; echo DONE` and navigate to a different folder while it runs. Nothing is injected; the sleep is untouched.
4. Wait for it to finish. The folder you navigated to is applied on its own.
5. Type `echo HALF-TYPED` without pressing Enter, then navigate elsewhere. The typed line is still intact and unchanged.
6. Press Enter. The pending folder change applies after your command runs.
7. Navigate to Trash. Synchronisation pauses and the terminal stays where it was.
8. Make a folder whose name contains a space and a quote, and navigate into it.

**Expected result:** the terminal tracks the browser while the shell is idle, never interrupts a running process or corrupts typed input, applies only the newest pending folder once it can, and leaves non-local locations alone.

## Related issue

Refs #93


spandan11106:
https://github.com/user-attachments/assets/49dfc278-e744-4005-9b8e-0310a306331e

What the video shows, in order:

1. Follows navigation. F4 opens the panel in strata-demo. Clicking into src/ and then documents/ injects a cd each time; pwd confirms the shell moved with the browser.
2. Leaves a running command alone. With sleep 20; echo DONE running, I navigate through images/, notes/ and src/. Nothing is injected while it runs — the sleep is untouched. When it finishes, a single cd to src/ appears, not three: only the newest pending folder is applied, once.
3. Leaves typed input alone. I type echo hello without pressing Enter and navigate to documents/. The half-typed line is still intact and unmodified. Pressing Enter runs my command, and the deferred cd lands after it. tcgetpgrp can't see a half-typed line, so this case is tracked separately through VTE's commit signal.
4. Quoting. Navigating into release notes/ injects the path single-quoted, so the space — and an apostrophe in a filename — can't be reinterpreted.
5. Non-local locations pause sync. Selecting Trash leaves the terminal where it was rather than inventing a path.

The browser stays the source of truth throughout: this is one-way, and nothing reads the shell's own cd back into the window.

guisilveira:
I reviewed the current stacked head of #1093 against both its actual base, PR #1089, and cumulatively against `main`.

The stack is clean:

```text
main: fb282723
  ↓
PR #1089: b1b2113
  ↓
PR #1093: 49fd7f7
```

The overall PR2 architecture is still aligned with the direction discussed in #93. The browser remains the source of truth, synchronization stays one-way, only the newest pending cwd is retained, `tcgetpgrp()` protects against foreground child processes, and non-local locations do not invent local paths.

I tested the safety behavior directly and found four blocking issues in the PR2 delta. All four were reproduced against the current head.

### 1. Non-UTF8 native paths are changed into a different path before being sent to the shell

`change_directory()` eventually reaches:

```rust
let text = path.to_string_lossy();
```

That is unsafe for Strata's native-path model.

I reproduced this with a directory containing the native byte `0xff`.

#### How to reproduce

Create an invalid-UTF8 directory:

```sh
mkdir -p /tmp/strata-sync-test
python3 -c "import os; os.mkdir(b'/tmp/strata-sync-test/bad-\xff')"
```

Open `/tmp/strata-sync-test` in Strata, open the embedded terminal, then navigate through the browser into that directory.

Observed terminal behavior:

```text
cd -- '/tmp/.../bad-�'
bash: cd: /tmp/.../bad-�: No such file or directory
```

The browser successfully entered the native directory, but the terminal remained in the parent directory because `0xff` had been replaced by `�`.

`pwd` still reported the previous directory.

There is a second issue in the same path: `applied` is updated before the injected `cd` has any chance to succeed, so the internal sync state can say the destination was applied even though the shell rejected it.

https://github.com/user-attachments/assets/b274d922-3978-4e50-8085-ab27e6e80204

The recording shows the browser entering the invalid-UTF8 directory, the generated `cd` containing `�`, the shell rejecting it, and `pwd` remaining in the parent directory.

#### Suggested direction

I would make this fail safe rather than lossy.

If the native path cannot be represented safely for the shell command, pause cwd synchronization for that location.

In particular:

```text
Path cannot be encoded safely
→ do not inject cd
→ do not update applied
→ leave terminal cwd unchanged
```

I would avoid `to_string_lossy()` here. Supporting arbitrary non-UTF8 paths through interactive shell commands can be handled separately if needed later. The important invariant is that path A must never silently become path B.

### 2. `tcgetpgrp() == shell pid` does not mean the shell is waiting at a command prompt

This is the main safety issue I found.

`tcgetpgrp()` correctly tells us whether another foreground process owns the PTY, but it cannot tell whether the shell itself is currently waiting at its normal command prompt.

A shell builtin can still be running inside the shell process itself.

#### How to reproduce

In the embedded terminal, run:

```sh
read -p 'waiting for input> ' value
```

Do not type anything into `read`.

While it is waiting, navigate Strata to another local directory.

The automatic `cd` is sent because the shell still owns the foreground process group.

Observed:

```text
VALUE=[cd -- '/tmp/.../archive']
/tmp/previous-directory
```

The injected `cd` was consumed as input by `read`. It never reached a normal shell prompt, and the cwd did not change.

The same semantic behavior was also confirmed directly with Fish.

https://github.com/user-attachments/assets/8176a6c0-46f5-4404-86ef-b94b6a7bfc73

The recording shows `read` waiting for input, browser navigation triggering synchronization, the injected `cd` becoming the value consumed by `read`, and `pwd` remaining in the original directory.

#### Suggested direction

I would keep `tcgetpgrp()` as one safety check. It is still useful for protecting against foreground processes such as:

```text
sleep
vim
less
cargo
npm
ssh
```

But it cannot be the signal that the shell is ready for command injection.

The synchronization needs a conservative concept of `PromptReady`, ideally using shell/VTE integration that can tell us when a fresh command prompt has actually been reached.

Conceptually:

```text
foreground child exists
→ unsafe

user input / command execution
→ unsafe

shell confirms a new prompt
→ safe

pending cwd may now be applied
```

Without confirmation of a fresh prompt, Strata should leave the cwd pending rather than inject anything.

### 3. Navigating to Trash does not invalidate an older pending local cwd

A non-local destination currently causes `follow()` to return early, but an existing `pending` cwd remains intact.

That means a previous local destination can still be applied after the browser has moved somewhere that should pause synchronization.

#### How to reproduce

Start the terminal in directory A.

Run:

```sh
sleep 8
```

While it runs:

1. navigate Strata to local directory B;
2. before `sleep` finishes, navigate to Trash;
3. remain in Trash;
4. wait for the command to finish;
5. run `pwd`.

Observed:

```text
Browser: Trash
Terminal cwd: /tmp/.../archive
```

The pending local directory B was applied after the browser had already moved to Trash.

https://github.com/user-attachments/assets/650f0899-06c2-471c-bd3b-c314515ed0de

The recording keeps Trash visibly selected while the deferred cwd change lands afterward, and `pwd` confirms that the terminal moved to the stale local directory.

#### Suggested direction

Entering a non-syncable location should invalidate the older pending local destination:

```text
Browser → B
pending = B

Browser → Trash
pending = None
```

I would also revalidate the pending cwd against the browser's current location immediately before `feed_child()`.

That gives the async path a second stale-work guard:

```text
pending exists
→ destination is still the browser's current syncable location?
→ yes: inject
→ no: discard it
```

This would make stale asynchronous state harmless instead of relying only on the location that originally populated `pending`.

### 4. VTE `commit` events cannot reliably tell whether the shell command line is empty

The current `typed` state is derived from individual VTE input events:

```rust
text.contains(['\r', '\n', '\u{3}', '\u{4}', '\u{15}'])
```

That does not represent the actual state of the shell's line editor.

I reproduced a case where the implementation considered the prompt safe while user input was still present.

#### How to reproduce

In the embedded terminal:

1. type, without pressing Enter:

```text
echo HELLO
```

2. move the cursor before the final `O`;
3. press `Ctrl+D`;
4. navigate Strata to another directory.

The line editor still contained:

```text
echo HELL
```

but `Ctrl+D` caused the PR's state to consider the prompt safe.

Observed:

```text
$ echo HELLcd -- '/tmp/.../archive'
HELLcd -- '/tmp/.../archive'
```

The injected cwd command was concatenated directly onto the user's unfinished command and executed as part of it.

The terminal remained in the original directory.

https://github.com/user-attachments/assets/8f8f7e1e-d18a-4a2c-8463-30da16d2d556

The recording shows the unfinished `echo HELL` line, the navigation event, the injected `cd` being concatenated directly to that input, and `pwd` remaining in the original directory.

#### Suggested direction

I would solve this together with the previous `tcgetpgrp()` issue rather than expanding the current keyboard heuristic.

A VTE input event is not the same thing as the shell's command-buffer state.

The safer model is:

```text
any user interaction / command execution
→ prompt not assumed safe

shell integration confirms a new prompt
→ prompt safe
```

Then `tcgetpgrp()` remains the foreground-process guard, while actual prompt readiness comes from the shell side.

This avoids trying to infer shell line-editor semantics from individual keys such as Enter, Ctrl+C, Ctrl+D, Backspace, paste, multiline input, and shell-specific editing behavior.

### Stacked dependency on PR #1089

The `TerminalPanel` lifetime leak reported on PR #1089 is still present in this stack.

PR2 also adds another strong capture:

```rust
let typing = panel.clone();

terminal.connect_commit(move |_, text, _| {
    typing.state.typed.set(...);
});
```

I confirmed the cumulative lifetime issue again with the `Rc::downgrade()` regression probe:

```text
strong_count_before_drop=4
strong_count_after_drop=3
weak.upgrade() == Some(...)
```

finding-2-terminal-panel-rc-cycle

The screenshot shows the panel retaining strong references after the external owner is dropped and the `weak.upgrade().is_none()` assertion failing.

I would not treat this as a separate PR2 architecture problem because the underlying ownership bug belongs to PR1.

But when #1093 is rebased onto the #1089 lifetime fix, this new `connect_commit` callback also needs to follow the same weak-ownership pattern.

Conceptually, all self-owned callbacks should end up following the same model:

```text
child_exited callback
close button callback
commit callback
        ↓
Weak<PanelState>
```

The regression test should remain after the fix.

### Shell syntax compatibility

I also checked whether the injected `cd -- '...'` quoting works with the shells available locally.

It worked with both Bash and Fish, including paths containing spaces and apostrophes.

Nushell was not installed, so I could not validate the broader shell-grammar concern. I don't think there is enough runtime evidence to block this PR on that point right now.

It is still worth keeping in mind that PR1 launches the user's configured `$SHELL`, while PR2 injects shell-language commands. If broader shell compatibility becomes a requirement, this may eventually need a small supported-shell capability boundary.

I would not block this PR on that without a concrete failing reproduction.

### Overall

I think the architecture of this PR is still going in the right direction, and I don't think it needs a redesign.

The main thing that needs tightening is the definition of "safe to inject a command".

Right now that safety property is inferred from:

```text
foreground pgrp
+
recent VTE input events
```

The reproductions above show that those two signals are not enough to guarantee that the shell is sitting at an empty prompt.

I would keep the current browser observer, latest-pending-cwd model and retry mechanism, but make the final injection conservative:

```text
latest local cwd is pending
        ↓
shell confirms a fresh prompt
        ↓
no external foreground process
        ↓
destination is still current
        ↓
path can be represented safely
        ↓
inject cd
```

If any of those checks fail, leave the terminal where it is and keep synchronization paused.

All four behavioral failures above were reproduced against the current PR2 head using the canonical isolated E2E environment with private Xvfb, D-Bus and AT-SPI.

The four recorded behavioral reproductions passed as expected, and the generated videos were successfully decoded with `ffmpeg`. Temporary probes and recording scripts were removed afterward, the checkout was left clean, and none of the review artifacts were uploaded automatically.

spandan11106:
You were right on all four, and working through them convinced me this PR should not land in its current form. I am closing it rather than patching it. Here is the reasoning.

### The four findings

1 and 3 are straightforward and I have fixed them where they still apply. 2 and 4 are the same problem, and that problem is not fixable at the VTE version this branch targets.

### Why 2 and 4 cannot be fixed here

Your suggested direction is right: prompt readiness has to come from the shell, not from guessing at the line editor. VTE does expose exactly that — `vte.shell.precmd` and `vte.shell.preexec` termprops, driven by OSC 133 — through `termprop-changed`.

Those need VTE 0.80. The E2E base image is Ubuntu 24.04, which ships VTE 0.76, as does Debian stable. Adopting prompt marks means raising the runtime floor and dropping current LTS users, which is too high a price for a convenience feature — especially since even at 0.80 it only works for shells that actually emit OSC 133, so most bash and zsh users would get nothing.

I checked whether OSC 7 could stand in, since VTE 0.76 does expose it and `vte.sh` emits it from the prompt hook. It deadlocks: the signal only fires when the reported directory *changes*, and the directory only changes if we inject first, so the permission to inject never arrives.

I also considered shipping our own prompt hook. Only bash picks one up from the environment; zsh and fish would need the user to source a snippet, so it covers one shell and asks everyone else to opt in.

That leaves your own rule as the operative one: without confirmation of a fresh prompt, leave the cwd pending rather than inject anything. At VTE 0.76 that means never injecting — which is this PR with its feature removed.

### What happens to the work

Once injection goes, almost nothing here is left. #1089 already reads the browsed folder at spawn time, so "a new session starts where you are browsing" is existing behaviour. This PR's real delta was the injection machinery — `tcgetpgrp` gating, the retry timer, the `typed` heuristic, `change_directory`, the rustix `termios` feature — all of which existed to serve it.

One fix here was worth keeping and has moved to #1089 as 033c088: the panel resolved its folder the way `Ctrl+T` does, which in Columns mode prefers the pointer-hovered column, so it could open in the parent the pointer happened to rest on rather than the folder the breadcrumb names. It now follows the active location.

Your findings 1 and 3 survive in that form too, as spawn-time guards on #1089: a path that cannot be represented for the spawn, and a location with no local path, must not become the next session's folder. Both are covered by 244ea01, which refuses rather than falling through to VTE's inherited cwd, with an E2E scenario for the Trash case.

The lifetime leak is fixed on #1089 in the same commit. All three strong captures now hold weak references — the `commit` handler you spotted here included, though it goes away with this PR — and the regression probe you wrote is kept as `the_panel_releases_its_state_when_dropped`.

#1094 has been retargeted onto #1089 and no longer depends on this branch.

### What happens to synchronisation

I would rather revisit it properly than ship a heuristic you have already shown to be unsound. I will open a follow-up issue for the `PromptReady` design you sketched, so it can be picked up when the VTE floor moves or shell integration is worth requiring explicitly. Until then the terminal opens where you are, and closing the session re-targets it.

Thanks for the reproductions — the `read -p` and `Ctrl+D` cases in particular are what settled this.

spandan11106:
Follow-up filed as #1103, carrying the `PromptReady` design and the VTE version constraint that blocks it.

PR #1126, opened 2026-09-18
fix(operations): sanitize FAT-invalid names instead of discarding the copy

### LINKED ISSUE #1123: fix(operations): copying to FAT/exFAT drives fails on names with ? : * etc. and discards the whole copy
### Strata version

0.18.0

### Installation method

Other

Omarchy 4.0.4 (Arch Linux), Hyprland 0.56.2

### What happened?

Copying a folder to an exFAT USB stick fails with **"Invalid filename"** when any file inside it has a name containing a character FAT/exFAT does not allow (`" * / : < > ? \ |` or a control character). Nothing is left on the stick: Strata writes the whole tree into a staged `.strata-replacement-*` directory first, then discards it all when the one bad name fails. In my case ~300 MB were written and then removed because of one file deep inside the tree, a LibreOffice icon cache entry named `sc_macroorganizer?tabid:short=1.png`.

GNOME Files copies the same folder to the same stick without errors, because it replaces the rejected characters with `_` when the destination is a FAT-family filesystem.

Pasting a single file whose name contains one of these characters also fails.

### Steps to reproduce

1. Create a folder containing a file with a FAT-invalid name, e.g. `mkdir demo && touch 'demo/a?b:c.txt' demo/ok.txt`.
2. Mount an exFAT (or vfat) formatted USB stick. No hardware needed: `truncate -s 256M fat.img && mkfs.exfat fat.img && udisksctl loop-setup -f fat.img`.
3. In Strata, copy `demo` and paste it onto the stick.
4. Observe: "Invalid filename", and `demo` does not exist on the stick.

### Expected behavior

The folder is copied, and names the destination filesystem cannot store are adjusted (e.g. `a?b:c.txt` → `a_b_c.txt`), matching GNOME Files. At minimum, one unrepresentable name should not discard everything that was already copied successfully.

### Screenshots or video

_No response_

### Relevant logs

```shell
$ cp 'a?b:c.png' /run/media/$USER/USB/
cp: cannot create regular file '/run/media/$USER/USB/a?b:c.png': Invalid argument
```

### Additional context

Root cause: in `src/adapters/local_operations.rs`, `copy_recursively_local` and `copy_recursively_with_progress` create every child with `target.child(&child_name)`, using the source name verbatim, and the paste loop does the same for top-level items. The vfat/exfat kernel drivers reject those names with `EINVAL`, and `copy_new_recursively_with_progress` then discards the staged directory.

A possible fix, mirroring GNOME Files: query `filesystem::type` of the destination once per transfer (GIO reports `msdos`/`vfat`/`exfat`). When it is FAT-family, map each child name by replacing rejected characters with `_` and dropping trailing dots and spaces, which the drivers strip anyway. A mapped name that collides with a sibling in the same source directory should get a numbered suffix, so that one entry never silently overwrites another in the staged copy. `WouldRecurse` moves go through the same copy path, so they get the fix too.

The diagnosis was done with help from an AI coding assistant.

pdwaldrop:
## Description

Copying a directory onto a FAT/exFAT/vFAT destination (a USB stick, most commonly) failed outright with "Invalid filename" as soon as any nested entry's name contained a character those drivers reject (`" * / : < > ? \ |`, or a control character) — and since the transfer stages into a temporary sibling directory first, the whole staged copy was discarded on that one failure, even after a large amount had already been written. Pasting a single file with such a name failed the same way.

Mirrors GNOME Files' behavior: when the destination reports a FAT-family filesystem type (checked once per transfer against the parsed mount table, not per file), rejected characters are replaced with `_` and trailing dots/spaces are trimmed (the drivers strip these anyway). Sibling names that collapse to the same result after sanitizing get a numbered suffix (`name (1).ext`) instead of one silently overwriting the other. This applies to the top-level paste target and every nested copy path: the recursive copy itself, move's cross-filesystem `WouldRecurse` fallback, replace, and merge.

Non-FAT destinations are completely unaffected — verified with a dedicated regression test.

## Visual evidence

N/A — this only changes filename handling during file transfers; there is no UI-visible change.

## How to test

1. Create a folder containing a file with a FAT-invalid name, e.g. `mkdir demo && touch 'demo/a?b:c.txt' demo/ok.txt`.
2. Mount an exFAT (or vfat) test volume, e.g. `truncate -s 256M fat.img && mkfs.exfat fat.img && udisksctl loop-setup -f fat.img`.
3. In Strata, copy `demo` and paste it onto the mounted volume.

**Expected result:** the folder is copied in full; `a?b:c.txt` lands as `a_b_c.txt` instead of the whole transfer failing and being discarded. Pasting a single file with an invalid name onto the same volume also succeeds, sanitized the same way.

*Local validation note: this machine has no docker/podman set up, so the pinned `./scripts/quality.sh` container runner wasn't available — used plain host `cargo fmt --all --check` / `cargo clippy --all-targets --all-features -- -D warnings` instead (both clean), plus `./scripts/test-headless.py` for the full `adapters::local_operations::` and `adapters::volume::` suites (all passing; two pre-existing, unrelated archive-compression test failures were confirmed present on unmodified `main` too, via `git stash`, before ruling them out as caused by this change).*

## Related issue

Closes #1123


l0gicgate:
Reviewed against upstream main `59ea3f6b2cf9f1867f77e6ea2ddad59a3b62cde3`, merged without rebasing, and pushed fixes in `1fe2e65a`:

- Merge planning now uses the same sanitized names and deterministic collision order as copying, so overwritten originals are backed up and undo records the actual destination paths.
- Handle case-insensitive sibling collisions, top-level batch collisions, and suffix-counter overflow. Added backup/path and collision regressions.
- Audited every added comment; removed redundant narration and kept only the driver constraint and ordering invariant.

There were no existing conversation comments, submitted reviews, or inline review threads to reconcile.

Local validation in the verified pinned image with isolated rootless Podman: `./scripts/quality.sh fmt`, `clippy`, and `test` passed (2,011 Rust tests passed; 18 ignored). These run the requested all-target/all-feature Cargo checks. Full `STRATA_CONTAINER_ENGINE=podman ./scripts/e2e.sh` initially passed all 771 tests; the final run passed 770 with one global-search failure caused by my container storage being visible through the workspace mount. Hiding that storage fixed the environment; the complete affected file, `tests/e2e/scenarios/test_search_filter_sort.py`, passed all 42 tests. No application workaround or test weakening was needed. `git diff --check` passed. All GitHub CI checks now pass for the pushed head.

No UI code changed. Manual testing on an actual mounted FAT volume was not performed; filesystem regressions exercise the FAT policy on temporary directories. No screenshots are claimed.

Remaining merge gate: repository rules require approval by someone other than the last pusher (`require_last_push_approval`). My approval is recorded, but GitHub still reports `REVIEW_REQUIRED`; another reviewer must approve this head. The PR has not been merged.

l0gicgate (approved):
Reviewed the complete updated diff and all added comments. The FAT merge backup/undo blocker and collision handling are fixed with regressions. Local checks and all GitHub CI checks pass; no remaining merge-blocking code findings. Actual mounted-FAT manual verification remains unperformed, as noted in the review summary.

PR #1128, opened 2026-09-18
feat(ui): add optional Yazi-style minimal browsing mode

### LINKED ISSUE #1152: feat(ui): add optional Yazi-style minimal browsing mode
### Problem or use case

Keyboard-driven browsing currently shares the default keymap with type-to-search and dense pane controls. Users who want Yazi/Ranger-style navigation need an explicit opt-in mode with a consistent command surface.

### Proposed solution

Add an experimental, saved minimal mode with home-row navigation, selection and file operations, footer prompts, and reduced browser chrome. Apply preference changes across open windows while preserving the default map when disabled. Keep the portal file chooser's keyboard handling and full chrome independent.

### Alternatives considered

Keep using the existing default keyboard shortcuts and type-to-search, or use a terminal file manager. The proposed mode keeps these preferences opt-in within Strata.

### Visual examples

PR #1128 has visual evidence marked pending owner upload. No captures are attached to this issue.

### Strata version

Source implementation reviewed at 907328a8f706b103eb598a4d0f89b7cb42c22bbf; local documentation cleanup at bf7196dff01df21deb1be839dd678ba4eb8b7ff9.

### Additional context

Tracks the existing draft PR https://github.com/lgse/strata/pull/1128. Created during its requested phase 0 cleanup to repair the missing issue linkage. Follow-up cleanup covers search focus/event delivery, preference transitions and object lifetimes, key semantics, responsive filesystem work, regression coverage, and review evidence before final validation.

wmfeht:
## Description

Adds **minimal mode**, an opt-in Yazi/Ranger-style browsing map. It hides dense pane chrome, disables type-to-search, and uses the footer as the typed-command surface so home-row keys can drive navigation, selection, and file operations.

Default remains **off**. The portal file chooser keeps its own dispatcher and full chrome. The mode is labeled experimental in Settings, Keybindings, F1, and the footer `MIN` tag.

**Enable / leave**

- Settings → General → Browsing → Minimal mode, or **Ctrl+Shift+M**
- **q** leaves the mode (footer flash `Left minimal mode — Ctrl+Shift+M returns`) and does not close the window
- **Q** closes the current window
- Saved as `minimal_mode` and live-updates every window, including teardown of open prompts, chords, and visual state
- **Type to search** and **Keep arrows in file list** stay saved; while the mode is on those Settings rows show **Not used in minimal mode.**

**Chrome**

- Hides window Search / Close and pane filter / refresh / sort controls in interactive browsers
- F1 / **~** and Settings → Keybindings swap to the minimal map
- Context-menu shortcut hints follow the active map (`x` cut, `y` yank, `p` paste, `d` / `D` trash / delete, `r` rename, `i` preview)
- Footer shows a `MIN` tag, chord hint lists, and prompt status

**Navigation**

- **h** / ← / Backspace / Alt+↑: leave preview, or parent folder (closes the Miller child)
- **l** / →: open a directory, or enter a file's preview when possible (does not open the file). Unpreviewable files and empty folders flash `Nothing to preview`
- **Enter** / **o**: open the focused item; **O** is Open With
- **j** / **k** / arrows: next / previous in listing order
- **g g** / Home, **G** / End, **Ctrl+U** / **D**, **Ctrl+B** / **F**, PgUp / PgDn
- **H** / **L** / Alt+← / Alt+→: history back / forward
- **i** toggles the preview drawer
- **J** / **K** (and listing motion keys) scroll the open preview once **l** / → has moved keys into it; **h** / ← returns to the listing without closing the preview

**Selection**

- **Space** toggles the focused item and moves down (does not preview)
- **v** / **V**: visual select / visual unset
- **Ctrl+A** select all, **Ctrl+R** invert
- After Space / Ctrl+A / Ctrl+R / leaving visual, browse motion moves the cursor without rewriting the fill
- **Esc** cancels a chord, then dismisses a hidden filter, then find highlights, then leaves visual (keeping the fill), then clears the selection
- Visual and multi-select in **f** filter and **s** search operate on visible matches / hits

**Files**

- **y** / **x**: yank / cut the selection, or the focused item when the fill is empty. Copied rows show the copy overlay; cut still uses scissors
- **p** paste (Keep Both focused on conflicts); **P** paste (Replace focused)
- **Y** / **X**: clear cut marks and this process's clipboard payload (does not wipe another application's clipboard)
- **d** / Delete: Move to Trash with confirmation; **D** / Shift+Delete: permanent delete with Cancel focused
- **r** / F2: rename in the footer prompt (not the row editor)
- **a**: create a file; a trailing `/` makes a folder. Conflicts error instead of uniquifying
- **c c** / **c n**: copy path / name
- **.** / Ctrl+H / Ctrl+.: hidden files
- **, a** / **, m** / **, s** / **, e**: sort by name / modified / size / type; Shift reverses. Armed `,` stays exclusive so the second key does not create, search, or find. **`,n`** and **`,t`** are not sort completions (natural sort and a type alias are not implemented)
- Empty-folder verbs flash `Nothing to yank` / `cut` / `delete` / `rename` / `preview` / `select`; empty clipboard **p** flashes `Nothing to paste`

**Places (`g` chord)**

While **g-** is armed, sidebar keycaps and a second-key hint list appear. Completions only run in that chord.

- **g** first item, **h** Home, **d** Downloads, **c** Config, **t** Trash, **n** Network, **r** Recent, **k** Documents, **p** Pictures, **v** Videos, **1–9** visible pins
- **Space**: footer `go ›` — type a path or URI; **Tab** / **Shift+Tab** cycle matching folders (`~` is home). No match keeps the typed text
- Missing special dirs / pins flash `No {name} folder` / `No pin N`; unknown second keys flash `Unknown chord`
- Sidebar rebuild cancels a pending **g** chord

**Prompts** (footer only; never the pane filter revealer or Search dialog)

- **/** / **?**: find next / previous name in this listing (does not hide rows). Enter keeps substring highlights; Esc from the listing dismisses them. **n** / **N** repeat (N reverses)
- **f**: filter this listing (hides non-matches). Stays until dismissed; reopening **f** pre-fills the query; view rebuilds keep the funnel collapsed
- **s**: recursive name search in the current folder (cap 100), independent of Include subfolders. Esc keeps hits so visual / Space / **i** / → can use them; Esc again (or **h**) dismisses. **Shift+S** is unbound
- **z** / **Z**: jump to a visited / recent folder from Strata history (not a zoxide database)
- Clicking a listing row closes the prompt and keeps that selection
- Up / Down move the listing or pick a history candidate without leaving the prompt

**Still bound vs remapped**

Ctrl+C / X / V, Delete / Shift+Delete, F2, F5, Ctrl+L, Ctrl+K, Ctrl+1/2/3, Ctrl+Shift+N, Alt+Enter, Menu / Shift+F10, and text-size chords stay available. Unbound or remapped default-map keys include Ctrl+Shift+K (use **z** / **Z**), Ctrl+T, Ctrl+\\, Ctrl+D (half page), Ctrl+F (full page; filter is **f**), Ctrl+B (full page up), Ctrl+R (invert; rename is **r**), **y** / **p** (yank / paste, not path-copy / pin), and Space (toggle selection, not preview).

Docs: [minimal mode keymap](docs/minimal-mode.md), with short summaries in keyboard navigation and preferences.

## Visual evidence

Pending owner upload. This is a user-visible chrome and keymap change; please attach sanitized screenshots or a short clip of (1) default vs minimal chrome, (2) the footer `g-` / `,` / `c` hint lists, and (3) a footer prompt (`/` find or `s` search) through GitHub's PR editor.

## How to test

1. Launch Strata with the default keymap. Confirm type-to-search and pane filter / sort chrome still behave as today.
2. Enable **Settings → General → Browsing → Minimal mode** (or press **Ctrl+Shift+M**). Confirm the experimental note, hidden Search/Close and pane filter/refresh/sort chrome, footer `MIN` tag, and F1 / Settings → Keybindings swapping to the Yazi map.
3. With type-to-search still on in Settings, press **j** / **k** / **h** / **l** and confirm they move the listing instead of typing a filter.
4. Open a folder with **l** / →; preview a file with **l** / →; scroll the preview with **j** / **k** / **J** / **K**; return with **h** without closing the drawer. Unpreviewable files and empty folders should flash `Nothing to preview`.
5. Yank with **y**, move with **h** / **l** / **g h**, paste with **p**. On a name conflict, **p** should focus Keep Both and **P** should focus Replace.
6. Open **g** and confirm sidebar keycaps plus the hint list; complete **g h**, **g d**, **g Space** with Tab completion, and an unknown second key (`Unknown chord`).
7. Exercise footer **/** find (highlights stay after Enter, Esc dismisses them), **f** filter (stays until dismissed), and **s** search (first Esc keeps hits, second Esc / **h** restores the directory). Confirm **v** / Space multi-select works on filter and search hits.
8. Sort with **, a** / **, m** / **, s** / **, e** (Shift reverses). **, n** and **, t** should cancel as unknown chords, not sort or run find/create.
9. Press **q** and confirm the leave flash and restored default chrome. **Q** should close the window. Open the portal file chooser with the preference on and confirm it still has full chrome.

**Expected result:** Minimal mode is a saved, live, opt-in map. Default browsing is unchanged when it is off. The chooser never takes the map. Chords stay exclusive until the second key, prompts stay in the footer, and leave/disable tears down prompts, hints, and visual state.

## Related issue

Closes #1152


wmfeht:
Review of `907328a8` — **request changes**. CI is skipped on the draft, so I ran fmt / clippy / the isolated Rust suite / canonical `./scripts/e2e.sh` locally (4-CPU VM). fmt and clippy pass; Rust suite 2124 passed / 7 failed; E2E 767 passed / 34 failed. Blockers only below; nits are in the private review.

1. **Default-map regression: Icons/List filter typing loses keystrokes.** `src/ui/inline_search.rs:759-770` re-`grab_focus()`s the previously focused widget after every debounced non-empty query. In default Ctrl+F / type-to-search that is the pane filter's `GtkText`; `grab_focus` on an already-focused text selects all, so the next key replaces the query. 30 of the 34 E2E failures are default-map Icons/List filter scenarios (`test_filter_results.py` ×13, `test_search_filter_sort.py` ×8, `test_inline_renaming.py` ×3, `test_clipboard.py[icons]` ×3, `test_quick_preview.py` ×2, `test_recursive_file_double_click_launches_once`) with assertions like `assert 'ote' == 'match-note'` / `'tch-note' == 'match-note'` and a11y snapshots showing the filter entry focused but empty after `type_text("photo")`. Restore only if focus actually moved, and use `grab_focus_without_selecting` for text widgets (see `collection.rs:257`).
2. **Leaving minimal mode leaks forced-recursive `s` state into the default map.** `apply_minimal_mode(false)` (`src/ui/window/keyboard.rs:270-283`) never calls `dismiss_hidden_filter()`; `force_recursive_search` / `InlineSearch.force_recursive` are only reset there (`browser.rs:1939`, `browser_modes.rs:917`) and are OR'd into every later filter (`columns.rs:853`, `inline_search.rs:748`). After `s foo` then `q`, later default Ctrl+F in that pane recurses regardless of *Include subfolders*, the hit list stays, and the footer mark is cleared. No test covers disable-while-`s`-showing.
3. **E2E tests contradict the PR's own behavior.** `test_minimal_mode.py::test_minimal_l_and_right_preview_a_file` (L314-315) and `test_minimal_search_hit_l_and_right[*]` expect a second `l`/Right to close the preview; `keyboard/minimal.rs:352` makes it a no-op while the preview owns keys and `preview_file_if_possible` uses `show`, not `toggle`. The GTK test `minimal_right_and_l_preview_a_file` pins the opposite. 4 deterministic E2E failures.
4. **Timing-dependent GTK tests fail/flake on a 4-CPU machine.** `minimal_icons_filter_and_list_search_visual_fill` failed 4/4 (alternating `wait_until` timeout and `Icons filter v then j must fill visible matches, got ["a.txt"]`; with a condition wait added it passed the Icons half and failed the List half 1/2 — a race between hit population and `v`/`j`), `minimal_g_space_tab_cycles_matching_folders` 2/3 alone (`navigate(root)` onto the shown root does not reload the new `docs`/`nested` rows), `minimal_file_view_focus_restored_after_prompt_search_and_error` once in the full run. Fixed `pump_mainloop(50ms)` after key presses instead of condition waits.
5. **Tautological tests.** `minimal_search_does_not_highlight_hit_names_live` (minimal_mode.rs:2128) and `minimal_search_keeps_hits_without_full_highlights_after_submit` (2168) assert on CSS class `search-hit-name`, which no production code adds (only `tests/minimal_mode.rs:202`) — constant-true; `minimal_filter_does_not_full_highlight_names` (2377) — nothing in the `f` flow touches find highlights; `shortcut_footer/tests.rs:387-406 minimal_reference_lists_the_yazi_map` — `contains` asserts on `const` slices; `settings/tests/reference.rs:200-206` const-vs-const `LABELED_TITLE` line.
6. **Excessive tests.** `tests/e2e/scenarios/test_minimal_mode.py`: 18 functions / 30 launches vs the design's "enter/leave, hjkl, yank-paste, footer /" scope; 15/18 re-run a GTK test in `minimal_mode.rs` (4 by identical name). Duplicate groups in `minimal_mode.rs`: `minimal_h_closes_the_miller_child` ⊂ `minimal_right_and_l_open_a_directory`; `minimal_a_selects_the_created_row` ⊂ `minimal_a_creates_file_and_folder_without_uniquifying`; four search-hit `l`/Right tests (3807/3873/4001/4055) for one dispatcher arm and one shared `InlineSearch`; `minimal_listing_l_previews_a_file_in_list` vs `minimal_right_and_l_preview_a_file`; `, s` ×3 and `,S` ×3; List block = Icons block in `minimal_search_footer_shows_result_count`; 4 fixtures for one helper in `minimal_g_special_dir_flashes_when_missing`; three disable-teardown tests. `clipboard/tests.rs:551-636 filtered_copy_feedback_follows_results_across_windows_and_rebuilds` clones the existing cut test — fold the `copied` asserts into it.
7. **`ViewState::handle` silently drops ModeViews events on re-entrancy.** `browser/events.rs:953` `borrow_mut()` → `try_borrow_mut()` (plus 12 new `try_borrow` fallbacks in `browser.rs`; base had none). Events are synchronous (`app/browser.rs:3215`), so a re-entrant `ColumnReloaded`/`EntryCreated`/`FocusChanged` now leaves Icons/List stale instead of panicking. Fix the re-entrant caller (likely `write_hidden_query`/`clear_all_filters` under a `mode_views` borrow → `set_filter_entry_query`/`show_directory_listing` → `notify_visible_listing`) or defer, rather than swallow.
8. **Reference cycles / per-window leaks.** `shortcut_footer.rs:607-615` `bind_minimal_mode` captures `self.clone()` (strong widget refs incl. `root`) in a process-wide preference listener anchored on `self.root`; the listener is only removed on `root::destroy`, which cannot fire while the listener owns `root`. Same shape at 616-624, 498-507 (`dispatch = footer.clone()`), 577-586 (`observe_browser` strong `view` + `footer` → `ViewState` self-cycle), and `keyboard.rs:62-82` (`window.clone()` / `view.clone()` in the prompt-entry focus controller). The base file uses `downgrade()` everywhere; see `browser/preferences.rs:13-18`.

Also: base `59ea3f6b` conflicts with current `main` in `src/ui/browser.rs`, `src/ui/browser/columns.rs`, `src/ui/chooser/tests.rs`; and `docs/design/minimal-mode.md` "Follow-up adjustments (not yet implemented)" items all say "Landed", including item 4 (`s` full-name highlight) which is not implemented and whose tests assert the opposite.

src/ui/window/keyboard/minimal.rs
```diff
@@ -0,0 +1,2295 @@
...
+                self.leave_visual_keep_fill();
+                self.view.select_focused_if_empty();
+                self.view.cut_selection();
+            }
+            Key::v if !event.shift() => {
+                if self.view.filter_has_focus() || event.text_has_focus() {
+                    return Some(Propagation::Proceed);
+                }
+                self.leave_visual_keep_fill();
+                self.view.paste();
```
wmfeht: **[bug]** Ctrl+V does not paste like `P`. `P` calls `minimal_paste(true)`, which uses `paste_into_listing` whenever the fill is not explicit, so a cursor-only directory stays the listing destination. Ctrl+V calls `BrowserView::paste` (`paste_with(..., honor_selected_folder: true)`). After `j`/`k`, `focus_only` clears `load_cursor`, so a focused folder is a normal one-item selection and the paste goes into that folder. Docs say Ctrl+V does the same as `P` (listing destination, Replace focused). Ctrl+V also skips the empty-clipboard "Nothing to paste" flash.

**Suggestion:** Call `self.minimal_paste(true)` from the Ctrl+V arm (still Proceed when the filter or another text entry is focused). Add a cursor-only folder case that asserts Ctrl+V and `P` land in the same directory.

src/ui/browser.rs
```diff
@@ -1715,16 +1850,779 @@ impl BrowserView {
...
+    pub fn toggle_focused_selection(&self) -> bool {
+        self.keyboard_navigation();
+        self.state.browser.toggle_focused_selection()
+    }
+
+    /// Selects the keyboard cursor when the filled selection is empty, so
+    /// yank, cut, and delete act on the cursor item instead of no-oping.
+    /// Search-result selections keep their existing behavior.
+    pub fn select_focused_if_empty(&self) {
+        if self.selected_search_results().is_some() {
```
wmfeht, line 1989: **[bug]** On an `f` or `s` overlay, an empty fill does not act on the focused hit. `select_focused_if_empty` returns immediately when `selected_search_results()` is `Some`, including `Some([])`. That is reachable: Ctrl+A then Ctrl+R on an `f` overlay inverts every match off, and visual unset can walk the selection down to empty, while `MinimalState`'s search cursor still points at a hit. `y` / `x` / `d` / `D` then flash or no-op, and Ctrl+C / Ctrl+X call `copy_selection` / `cut_selection`, which treat `Some([])` as "nothing" instead of the cursor. `c c` / `c n` go through `named_clipboard_entries` (around line 1427), which on an empty overlay uses `search_hit_index().unwrap_or(0)` — the GTK selection, not the independent cursor — so they copy hit 0. `O` goes through `open_with_entries` (around line 1458): a non-empty overlay uses the leftover fill (after Space, `o`/Enter activate the cursor and `O` does not), and an empty overlay falls back to `browser.focused_entry()`, the hidden directory row. `rename_target_entry` / `focused_search_result` already know the right hit.

**Suggestion:** When the overlay selection is empty, select or operate on the dispatcher’s current search cursor (the same entry `o`, `i`, and `l` use). Do not substitute hit 0 or the hidden listing focus. Keep using the filled overlay selection when it is non-empty, matching yank.

src/ui/window/keyboard/minimal.rs
```diff
@@ -0,0 +1,2295 @@
...
+                {
+                    let direction = if event.key == Key::Down { 1 } else { -1 };
+                    self.prompt_focus_lock.set(true);
+                    if matches!(
+                        kind,
+                        MinimalPrompt::HistoryFuzzy | MinimalPrompt::HistoryRecent
+                    ) {
+                        self.shortcuts.move_history_selection(direction);
+                    } else if !self.view.move_inline_search_results(direction) {
+                        let listing_hidden = self.view.view_mode() != BrowserMode::Columns
```
wmfeht: **[bug]** Footer prompt Up/Down does not walk Columns search or filter hits. `move_inline_search_results` always returns false in Columns, and `listing_hidden` is also false in Columns, so the prompt calls `minimal_step(..., searching: false)`, which is `browser.move_selection` on the hidden directory. The visible model is the search `StringList`. `FocusChanged` then runs `set_column_selections` with `column.map.view_position` of that directory index, so the hit list either stays put or selects an unrelated row. Icons and List move the overlay via `move_inline_search_results`. The prompt never calls `minimal_search_step` / `focus_search_hit`, and it does not update `search_cursor`.

**Suggestion:** If `selected_search_results()` is showing, call `minimal_search_step` (and remember the cursor) for every view mode, including Columns. Use directory `minimal_step` only when there is no overlay, such as `/` find.

src/ui/window/keyboard.rs
```diff
@@ -180,14 +249,142 @@ fn inside_pdf_scroll(widget: &gtk::Widget) -> bool {
...
+    lock: &Cell<bool>,
+) {
+    if lock.get() || minimal.borrow().prompt().is_none() {
+        return;
+    }
+    if footer.is_prompt_entry(&gtk::prelude::RootExt::focus(window)) {
+        return;
+    }
+    let kind = minimal.borrow().prompt();
+    minimal.borrow_mut().leave_prompt();
```
wmfeht, line 373: **[bug]** Clicking the listing out of an `f` prompt neither commits nor cancels the filter. The live `changed` handler has already written the hidden query and the `filter:` mark, but `remember_applied_filter` runs only from Enter. Focus loss calls `leave_prompt` and `hide_prompt` and then special-cases only find highlights and `s`. `applied_filter` stays `None` while the rows stay filtered. A later `s` overwrites the filter entry; dismissing that search calls `restore_applied_hidden_filter`, sees no remembered query, and clears the filter the user was still looking at. Find focus-loss cancels highlights; filter Esc cancels via `cancel_prompt`. Click-away is a third state, and `minimal_filter_stays_visible_until_dismissed` only covers Enter.

**Suggestion:** On Filter focus loss, if the hidden query is non-empty, `remember_applied_filter` it (the listing the user clicked is already filtered). If the query is empty, dismiss it the same way prompt Esc does. Do not leave a visible filter that search dismissal will drop.

wmfeht (commented):
Fixed

wmfeht:
## Review of `30244e8b4c99a5a1b88e9158a87b3065d2475435`

**Verdict: request-changes**

Agent `bc-ffc15806-506b-53a3-9851-37ceed7acc7a`. Base `c43f1eed`. The delta since `64797db3` is the modal focus restore in `src/ui/browser.rs` plus one extra case in `minimal_file_view_focus_restored_after_prompt_search_and_error`. That change is sound: it skips a mapped shortcut-footer widget, refuses to steal focus while a modal is open, and the new assertions fail if the shortcuts button keeps focus after the error dialog closes.

### Blockers

1. **Recursive `s` is not its own mode.** `docs/minimal-mode.md` says the footer shows `search: …` until dismissed, that hits survive a prompt Esc, and that Icons `h` / `l` / arrows stay on those tiles. `ShortcutFooter::set_filter_mark` only writes `filter:` (`src/ui/shortcut_footer.rs`). Search submit never sets a mark. `ActivePaneFilter` stores only `query` and `revealed` (`src/ui/browser/collection.rs`). `BrowserView::set_view_mode` restores that text and does not put `force_recursive` back (`src/ui/browser.rs`). With Include subfolders off, Ctrl+1 / 2 / 3 drops nested hits, and the view-change idle then labels the leftover query `filter:`. Pressing `f` while hits are showing is not a search command. It opens a filter seeded with `hidden_filter_query()` and, when that seed is non-empty, immediately calls `set_filter_query_without_revealer`, which writes the query with recursion forced off (`src/ui/window/keyboard/minimal.rs`). Nested hits disappear before Enter.

2. **Icons `h` / `j` / `k` / `l` stay on the preview map after a view switch.** `minimal_preview_keys` runs before icon motion and does not look at the view mode. Ctrl+1 / 2 / 3 applies the new mode and does not call `leave_preview_keys` (`src/ui/window/keyboard/minimal.rs`). After `l` on a file in List or Columns, Ctrl+2 leaves the drawer owning keys. In Icons, `j` / `k` scroll the document, `l` / Right are swallowed, and `h` only returns to the list. The keymap says Icons motion never takes preview ownership, even if a preview is already open.

3. **Chooser Esc in the filename field cancels the dialog before a filter, highlights, visual mode, or preview.** The filename control is a `gtk::Entry` (`src/ui/chooser.rs`). Any non-prompt text focus hits an early-out that closes the window, except location edit (`src/ui/window/keyboard/minimal.rs`). `docs/minimal-mode.md` says Esc cancels the dialog only after a prompt, filter, or preview is dismissed. With the file list focused, `minimal_escape` still does that. With the filename field focused, the first Esc closes the dialog.

4. **Backspace and Alt+Up do not go to the parent while the preview owns keys.** The navigation table assigns both to the parent folder, and the Icons section says they still go to the parent when a preview is open. The preview handler stops both after `leave_preview_keys` (`src/ui/window/keyboard/minimal.rs`). The parent action in `minimal_motion` never runs. A second press is required. The preview section only exempts `h` / Left from parent navigation.

### Nits

- Listing Esc snapshots the visible filter selection and puts it back after `dismiss_hidden_filter`. Esc inside the `f` prompt, and leaving the mode, dismiss the filter and do not. `docs/minimal-mode.md` says leaving the mode keeps the ordinary listing's filled selection. If that fill lives only on the filter rows, `q` drops it.
- GitHub Rust shard 0 still fails `sidebar_toggle_animated_expand_restores_visibility` (`content.position()` stayed 195, expected 0) after a fixed 600 ms settle with animations on. The same test passed in the local suite below. The wait does not observe the collapsed position, so a stalled frame clock fails the check.

### Notes

Fixed since the `907328a8` review and the `bd9f7d4` QA: default-map filter focus uses `grab_focus_without_selecting`; leaving the mode calls `dismiss_hidden_filter` and clears forced recursion; `p` opens the conflict dialog focused on Keep Both and `P` / Ctrl+V focus Replace; context-menu accelerators follow the active map; empty-folder `l` / `y` flash; `a` can open again; `try_borrow` no longer swallows mode-view events; footer and keyboard listeners hold weak refs; the contradictory E2E preview-close cases and the constant-true highlight tests are gone. No tautological or excessive-test blocker in the tests added or changed on this head. A committed non-recursive `f` filter is what view-switch restore covers; recursive `s` is the hole.

GitHub E2E shard 2 failed `test_filtered_item_menu_actions_use_the_real_location` for List, keyboard and pointer, timing out on "preview of the nested result" after Quick preview. Earlier asserts that the filter text was still `match-note` had passed. E2E was not re-run here.

### Validation

- `cargo fmt --all --check`: pass
- `cargo clippy --all-targets --all-features -- -D warnings`: pass
- Private Xvfb Rust suite (`GTK_A11Y=none`, `NO_AT_BRIDGE=1`, `STRATA_REQUIRE_GTK_TESTS=1`): 2399 passed, 41 failed, 20 ignored. Forty `portal_setup` failures are environmental (fixture path from `current_exe`, not this diff). One search-refresh test failed once with an empty query and passed on an immediate rerun; it is not in this diff. The sidebar animation test and the minimal-mode focus, tab-complete, and visual-fill tests passed locally.

wmfeht:
## Review

- **verdict:** `request-changes`
- **head SHA:** `7e8ab670eed8fe9b586cd38a39fe31953194c30d`
- **agent id:** `bc-3c7d20cf-3d2c-556d-8062-6a95ddb05cd1`
- **delta since `30244e8b`:** `Code review findings` (10 files, +809 / −24)

### Prior blockers (`30244e8b`)

1. **Recursive `s` as its own mode.** Fixed for the reported frame. `ActivePaneFilter.force_recursive` is restored across Columns / Icons / List. The footer uses `search:` while `force_recursive_search` is set. Prompt Esc on `s` keeps that mark. Opening `f` no longer rewrites the seed before the prompt is up. Icons `h` / `l` / arrows on those hits stay in the folder. Blocker 2 is what that open still does not cover.
2. **Icons motion after a view switch.** Fixed. `Ctrl+2` calls `leave_preview_keys`, and `minimal_preview_keys` drops preview ownership when the mode is already Icons.
3. **Chooser filename Esc.** Fixed. A filename `gtk::Entry` now runs `minimal_escape` instead of closing on the first Esc. Location edit still cancels in place.
4. **Backspace / Alt+Up while the preview owns keys.** Fixed. Both call `leave_preview_for_parent` on that press. `h` / Left still only return to the listing.

### Blockers

1. **Filter-prompt Esc poisons the next search prompt's Up/Down.** `cancel_prompt` for `f` now restores the overlay fill and `apply_kept_listing_fill` sets `explicit_fill` (`src/ui/window/keyboard/minimal.rs`). It does not clear the search cursor. The next `s` prompt still has that flag and the old cursor. `search_should_keep_fill` then returns true, so `focus_search_hit` is called with `replace_selection` false and the hit list selection stays empty while the internal cursor advances. `minimal_columns_prompt_up_down_walks_filter_and_search_hits` fails every run: filter Down/Up still move `index` `None → 0 → 1 → 0`, then search Down logs `keep_fill=true`, `selected_count=0`, `index=None`, `cursor` `0 → 1 → 2`. The same test passes on `30244e8b`.

2. **`f` over a live recursive `s` is only preserved until the prompt opens.** `preserve_search` skips the immediate non-recursive write, and `show_prompt` sets the seed before `connect_changed`, so the open frame keeps `search:` and the nested hit. Esc in that `f` prompt still takes the filter cancel path and calls `dismiss_hidden_filter`, which clears the search. Click-away (`dismiss_prompt_after_focus_loss` in `src/ui/window/keyboard.rs`) still `remember_applied_filter`s the search text and `set_filter_mark`s it, so the footer says `filter:` while `force_recursive` stays true. Dismissing search afterward restores that query with recursion off. The new test stops at "prompt is open".

### Nits

- The appearance menu's Columns / Icons / List buttons call `apply_browser_mode` and do not call `leave_preview_keys` (`src/ui/window.rs`). The Icons guard in `minimal_preview_keys` releases ownership on the next key, so `j` does not keep scrolling the document. `owns_keys_chrome` stays set until that key.

### Notes

The other new tests match the four previous blockers and the fill nit, and they passed: search mark across Esc / view change / open-`f`, Icons preview-key release, Backspace and Alt+Up, chooser filename Esc, filter-fill restore on prompt Esc and `q`, and the sidebar collapse wait. No tautological or duplicate-suite blocker in the tests added on this head.

`portal_setup` failures are environmental (`The Strata executable path must not be writable by other users`). That crate is not in this diff.

### Validation

GitHub checks on this draft head: Metadata policy success, `[code]smith` skipped. The fmt / clippy / Rust / E2E workflows did not run. E2E was not run here.

| Check | Result |
| --- | --- |
| `cargo fmt --all --check` | pass |
| `cargo clippy --all-targets --all-features -- -D warnings` | pass |
| private Xvfb `cargo test --all-targets --all-features` | **2415 passed, 30 failed, 20 ignored** in 921 s. 29 `portal_setup` failures are environmental. `minimal_columns_prompt_up_down_walks_filter_and_search_hits` failed (blocker 1). |
| Same test, 3× alone on this SHA | failed 3/3 |
| Same test on `30244e8b` | passed |

Build env: `CXX=g++ CC=gcc LIBRARY_PATH=/usr/lib/gcc/x86_64-linux-gnu/13`. Private Xvfb only (`GDK_BACKEND=x11`, `GTK_A11Y=none`, `NO_AT_BRIDGE=1`, `STRATA_REQUIRE_GTK_TESTS=1`).

wmfeht:
# Review: lgse/strata#1128 `feat(ui): add optional Yazi-style minimal browsing mode`

- **verdict:** `request-changes`
- **head SHA:** `7eacc9992bd573bc715160002cb1245f9c173747`
- **agent id:** `bc-5d42db0f-f7c5-5682-8e13-66d82068aa0f`
- **base:** `c43f1eedd1129f22cd735fb661e9f4d8468586b1`
- **delta since `7e8ab670`:** `fix(ui): stop filter dismiss from corrupting search and preview keys` — 5 files, +400 / −115

## Prior blockers (`7e8ab670`)

1. **Filter-prompt Esc poisons the next search prompt's Up/Down.** Fixed for that frame. Cancelling a real `f` prompt clears the search cursor and does not mark the restored rows as an explicit fill. `minimal_columns_prompt_up_down_walks_filter_and_search_hits` now requires the first search Down to select hit 0, and it passed.
2. **`f` over a live recursive `s` is only preserved until the prompt opens.** Fixed for Esc and click-away. Both leave `search:`, `force_recursive`, and the nested hit in place, and dismissing the search does not bring the query back as a filter. `minimal_recursive_search_keeps_search_mark_across_escape_view_and_filter` passed.

The previous Icons-menu nit is fixed. The appearance menu calls `release_owned_keys` on Icons, and `minimal_icons_menu_releases_preview_keys` passed.

## Blockers

1. **Filter-prompt Esc drops a Space fill on the next `j`.** `cancel_prompt` for a real filter always calls `set_explicit_fill(false)` before restoring the kept rows (`src/ui/window/keyboard/minimal.rs`). Space leaves the fill on the current row and moves the cursor to the next row. Esc then restores that one location, and `apply_filled_locations` parks the cursor on it because the cursor was outside the fill. The flag is clear and the selection is cursor-only, so the next `j` uses `step_selection` and replaces the fill. Reproduced on this SHA and then removed: after Esc, cursor `0` and fill `[0]`; after `j`, cursor `1` and fill `[1]`. Up/Down inside the filter prompt does not set `explicit_fill`, so `clear_search_nav` plus `mark_explicit: false` already covers the search-Down failure. A multi-row Ctrl+A fill still sticks because `cursor_only_fill` is false. `minimal_columns_prompt_up_down_walks_filter_and_search_hits` never Spaces first, so it stays green.

## Nits

None.

## Notes

Enter in an unedited `f` prompt over a live recursive `s` still submits `set_filter_query_without_revealer`, which forces recursion off. The new coverage is Esc and click-away. Those were the reported dismiss paths.

The added tests match those paths and the Icons menu. No tautological or duplicate-suite blocker.

Three `portal_setup` failures assert installed `Exec` paths (`file_manager_service_installs_and_reports_status`, `install_and_uninstall_restore_an_existing_user_configuration`, `startup_refreshes_a_configured_portal_running_an_old_executable`). That crate is not in this delta.

## Validation

GitHub checks on this draft head: Metadata policy success, `[code]smith` skipped. The fmt / clippy / Rust / E2E workflows did not run. E2E was not run here.

| Check | Result |
| --- | --- |
| `cargo fmt --all --check` | pass (`FMT_EXIT:0`) |
| `cargo clippy --all-targets --all-features -- -D warnings` | pass (`CLIPPY_EXIT:0`) |
| `xvfb-run -a env -u WAYLAND_DISPLAY GDK_BACKEND=x11 GTK_A11Y=none NO_AT_BRIDGE=1 STRATA_REQUIRE_GTK_TESTS=1 cargo test --all-targets --all-features` | **2443 passed, 3 failed, 20 ignored** in 838 s. The 3 failures are the `portal_setup` cases above. The three new or extended minimal-mode tests passed. |
| Throwaway Space, `f`, Esc, `j` on this SHA | failed as in blocker 1; the test was removed |

Build env: `CXX=g++ CC=gcc LIBRARY_PATH=/usr/lib/gcc/x86_64-linux-gnu/13`. Private Xvfb only.

wmfeht:
# Review: lgse/strata#1128 `feat(ui): add optional Yazi-style minimal browsing mode`

- **verdict:** `request-changes`
- **head SHA:** `1918921b923368de1359b88f1c09e9fc23309e24`
- **agent id:** `bc-303f3db1-9c66-5fa4-a976-b0f604b1bd41`
- **base:** `152299cbc2da3393815bbc1f5647323b9667e826`
- **since `7eacc999`:** `b8a6779` (CI fix), then merges of `main` through `152299c`

CI on this head is green (format, lint, Rust tests, E2E). Not re-run.

## Prior blocker (`7eacc999`)

**Filter-prompt Esc drops a Space fill on the next `j`.** Still present. `b8a6779` only changes `rename_target_entry` and search pointer focus. `cancel_prompt` for a real `f` still calls `set_explicit_fill(false)` and restores with `mark_explicit: false`.

## Blockers

1. **Filter-prompt Esc still collapses a Space fill on the next `j`.** Space sets `explicit_fill` and `move_cursor`s to the next row, leaving the fill on the previous row. Esc in an unedited `f` prompt (`cancel_prompt` in `src/ui/window/keyboard/minimal.rs`) clears that flag, then `apply_filled_locations` parks the cursor on the restored row because the cursor is outside the fill. The selection is then cursor-only with the flag clear, so the next `j` takes `step_selection` / `move_selection` and replaces the fill. A multi-row fill still sticks because `cursor_only_fill` is false. Up/Down inside the prompt still does not need the flag clear; that clear is what drops the Space fill. `minimal_columns_prompt_up_down_walks_filter_and_search_hits` never Spaces first.

2. **Permanent-delete focus moves from Cancel to Confirm when the size summary finishes.** `D` and Shift+Delete call `confirm_delete_preferring_cancel`, and `show_delete_confirmation` does start on Cancel when `prefer_cancel` is set. The merged total-size task then always runs `confirm.grab_focus()` after `aggregate_directory_summary` (`src/ui/browser/trash.rs`), and Confirm is insensitive until that moment. Once the subtitle appears, Enter activates permanent delete. `docs/minimal-mode.md` still says Cancel is focused. The updated trash tests call `show_delete_confirmation` with `prefer_cancel: false`, so they stay green.

## Nits

1. **Redo never reaches the default handler in minimal mode.** `minimal_motion` handles Ctrl+Z undo, then unmatched Ctrl chords return `Propagation::Proceed` from `minimal_commands`. `handle_key` returns that directly and does not run `file_commands`, so Ctrl+Shift+Z and Ctrl+Y do not call `redo_last_operation`.

2. **Archive member keys from `docs/keyboard-navigation.md` do not run while minimal mode is on.** `archive_navigation` is only on the default dispatcher. Minimal `j`/`k`/arrows move the listing, or scroll the preview once it owns keys.

## Notes

New tests in `b8a6779` match real failures: empty-overlay rename must not fall through to the hidden file, pointer focus retargets search commands without clearing a Ctrl-click selection, the keybindings page must drop its listeners, and a Trash subtitle must not say “permanently deleted”. No tautological or duplicate-suite blocker.

Opening `f` over a live recursive `s` and pressing Enter still submits `set_filter_query_without_revealer` (recursion off). Unchanged from the previous head; Esc and click-away are the covered dismiss paths.

Shared search rows no longer use GTK single-click-activate. Unmodified release calls `single_click` when the selection size is at most one; Ctrl/Shift presses set `PointerSequence` activation to false and do not open. That split matches the previous activate-signal split.

PR #1129, opened 2026-09-18
feat(window): make close, minimize, and maximize buttons configurable

### LINKED ISSUE #1100: feat(window): make close, minimize, and maximize button visibility configurable
### Problem or use case

Strata always hides GTK title buttons (`HeaderBar::set_show_title_buttons(false)`) and always shows a custom close control in the in-app header. There is no minimize or maximize control, and close cannot be turned off.

That mix does not fit both tiling window managers and ordinary stacking desktops:

- On a TWM (i3, Sway, Hyprland, and similar), window management is the compositor's job. In-app close/minimize/maximize are redundant or wrong (minimize/maximize especially).
- On a normal window manager (GNOME, KDE, XFCE, and similar), users expect close, minimize, and maximize to be available on the window.

Visibility of those three buttons should be a user choice, not a single hard-coded layout.

### Proposed solution

Add a saved preference that controls whether the close, minimize, and maximize buttons are shown. Apply it at window construction and live, through the existing preference-binding path; Settings should only edit the preference.

Tiling WM users should be able to hide some or all of the buttons. Stacking WM users should be able to show the conventional three. Keep today's close-only header as the default unless a better default is chosen in implementation.

This is about button visibility on Strata's header, not the outer window frame.

### Alternatives considered

- Leave the current close-only header: works on some TWMs, but stacking desktops never get minimize/maximize and cannot hide close.
- Always restore GTK title buttons: better for some stacking WMs, worse for TWMs that already decorate or keybind those actions.
- Detect the window manager and pick a layout automatically: brittle, and still needs an override for mixed setups.
- Fold this into #1098: that ticket is the outer border/chrome, and it explicitly keeps the in-app header and close control.

### Visual examples

N/A — no mockup; compare a TWM session (often no in-app window buttons) with a stacking desktop that expects close/minimize/maximize.

### Strata version

N/A — current main header construction (custom `HeaderBar` plus always-on close). Recheck against current main before implementation.

### Additional context

Searched open and closed lgse/strata issues for close button, minimize, maximize, window controls, titlebar buttons, CSD buttons, TWM, window manager decorations, and hide window buttons. No existing ticket asks to configure those three buttons' visibility.

Related but distinct:

- #1098: drop outer window borders/chrome on non-Hyprland desktops; keep Strata's header and close control.
- #156: close-button padding, not visibility.
- #469 / #497: XFCE toolbar and pane icon scale.
- #577 / #579: compact the in-app top bar height.

Filed from the feature-request form. No P-band; triage can band it later.

spandan11106:
## Description

The in-app header always showed a close button and nothing else, which suits a tiling WM but leaves stacking desktops without minimize or maximize, and no way to hide close.

Three saved preferences now control the header's window buttons. Close stays on by default and minimize/maximize stay off, so the current layout is unchanged until someone opts in. The buttons bind through the shared preference path, so they apply at window construction and live across open windows; Settings only edits the preference. The maximize button toggles maximize/restore and swaps its icon with the window state.

## Visual evidence

Pending — screenshots of the header with all three buttons enabled, and of Settings → General → Window buttons, to be attached here.

## How to test

1. Open Settings (Ctrl+,) → General → **Window buttons**.
2. Turn on **Show minimize button** and **Show maximize button**; watch the top bar of every open window update immediately.
3. Click maximize, confirm the window maximizes and the icon becomes a restore icon; click again to restore.
4. Turn off **Show close button** and confirm it disappears; restart Strata and confirm all three choices are still in effect before opening Settings.

**Expected result:** the three buttons appear and disappear live, the choice survives a restart, and a fresh profile still shows close only.

## Related issue

Closes #1100


spandan11106:
image
image

PR #1132, opened 2026-09-18
fix(a11y): name the location entry and its confirm/cancel buttons

### LINKED ISSUE #1131: fix(a11y): name the location entry and its confirm/cancel buttons
### Strata version

0.19.0 (`984706cdd123`, lgse/strata `main`)

### Installation method

Built from source

Arch Linux, Hyprland. Reproduced from the accessibility tree (AT-SPI), not a specific desktop session.

### What happened?

Ctrl+L shows the location entry and two icon-only buttons (confirm and cancel). The entry has a tooltip and placeholder, and the buttons have tooltips, but none of them expose a stable accessible name.

Screen readers therefore announce an unnamed text field and unnamed buttons. Automation that locates widgets by accessible name cannot target the location bar, confirm, or cancel without a dedicated action such as focusing via Ctrl+L and then sending Return/Escape.

Header chrome already has spoken names (`Search (Ctrl+K)`, `Toggle sidebar (Ctrl+B)`, and so on). The location bar is the missing piece of that same contract.

### Steps to reproduce

1. Open Strata.
2. Press Ctrl+L so the location entry replaces the breadcrumbs.
3. Inspect the accessibility tree for the focused text field and the two icon buttons beside it.

### Expected behavior

The location entry has an accessible name, and the icon-only confirm and cancel buttons have accessible names, so they can be found by role and name the same way as other header controls.

### Screenshots or video

N/A — this is an accessibility-tree naming defect, not a visual one.

### Relevant logs

N/A — no crash or terminal error.

### Additional context

`src/ui/browser.rs` builds `location_entry` with placeholder `Enter an absolute path` and tooltip `Location (Ctrl+L)`, and the icon-only confirm/cancel with tooltips `Navigate (Enter)` and `Cancel (Escape)`. GTK does not treat those as accessible names for an entry or for icon-only buttons whose child is a themed image.

Existing coverage owner for named chrome is `tests/e2e/scenarios/test_accessibility.py` (`test_toolbar_controls_are_named`). Location navigation today finds the field by focused editable role (`editable_field()`), not by name.

Do not request a core dump.

wmfeht:
## Description

Give the location entry and its icon-only confirm/cancel buttons stable accessible names.

Ctrl+L already shows those widgets, but GTK does not treat the entry tooltip or the icon-only button tooltips as AT-SPI names. Screen readers and automation that locate controls by name could not target them without focusing the field and sending Return or Escape.

The names match the existing tooltips: `Location (Ctrl+L)`, `Navigate (Enter)`, and `Cancel (Escape)`.

## Visual evidence

N/A — this only sets accessible names. The location bar looks the same.

## How to test

1. Open Strata and press Ctrl+L so the location entry replaces the breadcrumbs.
2. Inspect the accessibility tree (or Accerciser / the e2e dump) for the focused text field and the two icon buttons beside it.
3. Type a path and click the named Navigate button instead of pressing Enter. Cancel by clicking the named Cancel button.

**Expected result:**

The field is named `Location (Ctrl+L)`, confirm is `Navigate (Enter)`, and cancel is `Cancel (Escape)`. Clicking Navigate opens the typed path. Clicking Cancel restores the breadcrumbs.

## Related issue

Closes #1131


l0gicgate:
Merged verified upstream `main` (6a9cf98a), preserving its path/URI autocomplete while retaining the location controls’ accessible names. Added an AT-SPI end-to-end regression that finds the named field and clicks Navigate and Cancel; shortened the location-label comment. Also reduced the list-row focus-visible outline from 2px to 1px so a focused file remains thinly outlined after Escape clears selection. I inspected a fresh isolated-display screenshot after keyboard selection/Escape and exercised the relevant E2E interactions. The previous Rust shard failure was an intermittent camera-streaming stress assertion; its isolated rerun and the subsequent full local Rust suite and new CI shards passed. Visual evidence is captured locally but **pending upload**: this CLI cannot upload a screenshot as a GitHub attachment; the owner can attach the sanitized capture via the GitHub comment editor.

PR #1133, opened 2026-09-19
feat(browser): persist the Icons view thumbnail size across restarts

### LINKED ISSUE #1113: feat(icons): remember Icons thumbnail size across restarts
## Description

The **thumbnail size slider** in **Icons view** (drag the "px" control in the Icons toolbar) is not remembered across app restarts. Setting it to e.g. 95 px works for the session, but relaunching Strata resets the grid to the default 64 px.

## Expected behavior

Once set, the thumbnail size stays at the chosen value across restarts (and ideally is stored per-preference along with the other persisted browser/view settings).

## Actual behavior

Every launch begins at the hardcoded default of 64 px; the slider change is only session-local.

## Root cause (from a quick look at the source)

In `src/ui/browser_modes.rs`, the value lives purely in memory and is never written to a saved setting:

- `const DEFAULT_ICONS_THUMBNAIL_SIZE: i32 = 64;`
- `icons_thumbnail_size: Rc<Cell<i32>>` is seeded to `DEFAULT_ICONS_THUMBNAIL_SIZE` when the browser is built;
- the Icons controls write it back with `thumbnail_size_for_change.set(size)` (slider `value-changed`);
- nothing persists it — no key exists in the saved settings/settings.toml, and the thumbnail-generation path (`src/ui/thumbnail.rs`) always keys the shared cache at `CANONICAL_MAX_EDGE` (256 px), so the cache is unaffected.

Because the size is not one of the persisted browser preferences, the grid re-renders at 64 px on each new session.

## Suggested approach

Persist the Icons thumbnail size the same way other per-view settings are saved (e.g. alongside the existing persisted per-view state / settings plumbing), and restore it when the browser window is rebuilt. A small clamp on load (e.g. 16–256) would mirror the runtime `clamp(16, 256)` already used by the thumbnail request code.

## Notes

- The feature request is independent of thumbnail *cache* behavior — the shared freedesktop thumbnail cache (`~/.cache/thumbnails/large`, canonical 256 px, mtime-validated) already works correctly and is display-size-agnostic; changing the slider does not invalidate it.
- Related prior work: #888 deliberately made several Preview/sizing values "session-local" with "no new saved settings"; this request is specifically about the Icons thumbnail/zoom control crossing to persisted state.

pdwaldrop:
## Description

The Icons view's thumbnail-size slider (the "px" popover in the Icons toolbar) only lived in memory: relaunching Strata always reset it to the hardcoded 64px default, discarding whatever size was chosen last session.

Added `icons_thumbnail_size` to the saved preferences (clamped 32–256px, matching the slider's own range) and bound it the same way other browser-construction preferences work:

- A browser binds it before the browser-mode preference applies, so if the saved mode is Icons, the pane is built at the right size on the very first frame — no flash at 64px.
- The popover slider's own live drag persists the new value.
- Another window's currently visible Icons pane moves its slider (and resizes thumbnails) to match, without a full pane rebuild — the same live cross-window sync every other saved preference gets.

## Visual evidence

Slider at 203px, then after quitting and relaunching Strata, the Icons view rendering at 203px again immediately (no manual re-adjustment):

Thumbnail slider at 203px before restart
Thumbnail slider still at 203px after restart

## How to test

1. Switch to Icons view (Ctrl+2), open the thumbnail-size popover, and drag the slider to a non-default size (e.g. 200px)
2. Quit and relaunch Strata
3. Switch to Icons view again
4. Expected: thumbnails render at the previously chosen size immediately, and the popover slider shows that same value — not reset to 64px
5. With two windows open, dragging the slider in one moves the other's slider and resizes its thumbnails to match

## Related issue

Closes #1113

---

Note on validation: docker/podman aren't set up on this machine, so `./scripts/quality.sh` and the containerized E2E suite weren't run. Validated instead with `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, the extended exhaustive multi-window preferences fixture/coverage tests (`ui::preferences::`, `ui::browser_modes::`, `ui::window::`, `ui::window::composition::`, `ui::settings::tests::preferences`, `ui::browser::tests::preferences`), and manual GUI verification (with an isolated `XDG_CONFIG_HOME`) of the slider surviving a real app restart, shown above.


l0gicgate:
Reviewed the updated diff and pushed `e86eff2b` + `92d69b09`:

- Merged verified upstream `main` at `a75fa8332a82a4fc0d085d8f5f04191c6e2d8227`, retaining both thumbnail-size and date-format defaults in the sole conflict.
- Strengthened existing preference regressions to check the live Icons slider and a saved 203px startup value, including mode changes and multiple browsers.
- Audited every added code comment and removed both redundant doc comments. No UI implementation changes were needed.

No existing PR conversation comments, submitted reviews, or inline threads were present; no unresolved code blockers found.

Validation: rootless Podman `STRATA_CONTAINER_ENGINE=podman ./scripts/quality.sh all` passed formatting, Clippy, and all-target/all-feature Rust tests (2,138 passed, 20 intentionally ignored). `STRATA_CONTAINER_ENGINE=podman ./scripts/e2e.sh` passed all 773 cases. GUI checks used private displays/buses. An initial unrelated package-probe `ETXTBSY` failure passed both its exact reproduction and the subsequent full run. Full local validation was run at the owner's explicit request. No separate manual restart/drag session was performed during this review; the author's existing visual evidence remains unchanged.

Approved the pending fork CI workflow execution and watched all four required checks pass on `92d69b09`. GitHub still requires an approving review from an eligible reviewer; the PR has not been merged. The review commits are unsigned, so retain the repository's signed-commit policy when selecting the eventual merge method (do not bypass it).

PR #1136, opened 2026-09-19
fix(archives): remove empty destination left by failed subfolder extraction

### LINKED ISSUE #1135: Extracting an unsafe archive leaves an empty destination folder behind
## Strata version
v0.19.0 (reproduced on current main, source build)

## Installation method
Source build; not install-method specific (logic is in the archive operation path)

## Environment
Linux, GTK — platform-independent

## Steps to reproduce
1. Create a ZIP containing a member with an unsafe path, e.g. `../outside` (zip slip)
2. Activate the archive in the browser (open/double-click), which extracts to a new subfolder named after the archive
3. The extraction correctly fails: a popup reports the archive is unsafe ("Refusing unsafe archive path")

## Expected behavior
The destination folder created for the extraction is removed when nothing was extracted, matching the cleanup added in #827 for extractor-created destinations.

## Actual behavior
An empty folder named after the archive (e.g. `unsafe/`) is left behind next to the archive.

## Root cause
`extract_entry_to_subfolder` reserves the destination with `fs::create_dir` before dispatching (`src/ui/browser/archive.rs`), so `local_operations::archive::extract` sees `dest_dir.exists()` and clears `created_dest`. The failure/cancel cleanup only runs for destinations the extractor itself created, so the caller-created empty folder survives.

## Notes
- "Extract here" and "Extract to…" are unaffected: their destinations either pre-exist or are created by the extractor itself.
- Fix: let `ExtractRequest` declare the destination was created for the operation so the existing empty-folder cleanup applies.

JoeJoeflyn:
## Description

Opening an archive extracts it into a new subfolder named after the archive. That folder is reserved with `create_dir` before the operation is dispatched, so the extractor's failure cleanup — which only removes destinations it created itself — never applied to it. When extraction then fails (for example an archive with an unsafe `../` member), an empty `<archive-name>/` folder was left behind.

`ExtractRequest` now carries `created_destination`, letting a caller that already reserved the destination mark it operation-created so the existing empty-folder cleanup applies. Only the extract-to-subfolder path sets the flag: "Extract here", "Extract to…", and password retries keep their behavior, a folder that already existed is still never removed, and a folder containing extracted content is kept since the cleanup only removes empty destinations.

## Visual evidence

https://github.com/user-attachments/assets/141f060c-33de-470e-bf4c-ef7d7f49265c

## How to test

1. Create a zip with an unsafe member: `python3 -c "import zipfile; zipfile.ZipFile('/tmp/evil.zip','w').writestr('../evil.txt','x')"`.
2. Open/activate `evil.zip` in the browser — the "unsafe archive" error popup appears.
3. Dismiss the popup and inspect the folder.

**Expected result:** no empty `evil/` folder remains beside the archive. A normal archive still extracts into its subfolder, and an archive that wrote files before failing keeps its folder with the partial content.

## Related issue

Closes #1135


wmfeht:
## Review (`6d731be7ec3161da103c9f404b236b7e58055909`)

**Verdict: `approve-with-comments`** — no blockers. Nits do not need another code round.

Judged this SHA on its own (the earlier pass on `a7cfee82` was also `approve-with-comments`). The second commit only merges `main` and does not change the archive fix. CI on this SHA is green (format/lint/test, Rust shards, E2E, quality, dependency/spelling, metadata, release scripts; `[code]smith` skipped). `mergeable_state` is `behind` because `main` moved after that merge, not because a check failed.

### What holds

#1135: activating an archive reserves `<stem>/` with `create_dir` before dispatch, so the extractor saw `dest_dir.exists()` and skipped empty-folder cleanup. An unsafe `../` member still failed closed and left that empty folder behind.

- `ExtractRequest::created_destination` ORs into the existing `created_dest` check. Extractor-created destinations still clean up via `!dest_dir.exists()`.
- Activation (`ExtractRequested` → `extract_entry_to_subfolder`) is the only `true`, and only after `create_extraction_subfolder` wins an exclusive `create_dir` (collisions, including files and dangling symlinks, step to `stem (n)`). “Extract here”, “Extract to…”, and the password retry pass `false`.
- Failure and cancel cleanup is unchanged: `created_dest`, outcome is not `Completed`, and `read_dir` is empty, then `remove_dir`. A pre-existing folder stays unless the caller set the flag. A folder that already holds extracted members stays.
- Zip slip still refuses the member before `extract_member` (`enclosed_name` → `Refusing unsafe ZIP path`). The new test also asserts `outside` was not written.

### Blockers

None.

### Test quality

`unsafe_extraction_removes_a_caller_created_destination` is the #1135 path: pre-created empty dest, `created_destination: true`, zip member `../outside`, failure message contains `unsafe`, dest gone, no write outside. It runs through `LocalOperationProvider.extract`. It fails if the flag is ignored, because a pre-existing dest with the flag false is kept (`failed_extraction_preserves_a_pre_existing_destination`, and the `unsafe.zip` case in `extraction_failures_stop_progress_and_preserve_error_distinctions`). It sits next to `failed_extraction_removes_a_newly_created_empty_destination` (extractor-created, dest absent at start).

The other test edits only add `created_destination: false` so existing `ExtractRequest` literals compile.

Not excess. Not tautological.

### Nits / suggestions

1. `Browser::extract` takes `created_destination` as a positional `bool`. Four call sites; the subfolder path is the only `true`. The name on `ExtractRequest` is the clear contract. A small enum at the browser boundary (`CallerCreated` / `UseExisting`) would make a missed `true` on activation, or a stray `true` on “Extract here”, obvious at the call.

### Notes

- No UI test that `extract_entry_to_subfolder` passes `true`. The adapter test sets the flag itself. Activation in `events.rs` calls `extract_entry_to_subfolder`, and that function passes `true` after `create_extraction_subfolder` succeeds. A wiring-only test would not add a failure mode.
- Empty pre-existing “Extract to…” folders stay, because that call passes `false` and the directory already exists. The keep-user-content rule is covered with a file inside the folder.
- Password retry always passes `false`. After a failed empty subfolder extract, cleanup has already removed the directory, so the retry’s `!dest_dir.exists()` path recreates it and still removes it if that attempt also leaves it empty. Threading the flag through `pending_extract_retry` is unnecessary for #1135. The reserved folder can disappear before the password dialog opens; submit creates it again.

l0gicgate:
Merged verified upstream `main` at `14034a71596fdcc62f51720b76f9366db35822d8` and reconciled the extraction regression with its new archive-path sanitization. The old `../` fixture now extracts safely, so the regression uses a malformed ZIP to exercise failed extraction and empty-folder cleanup. Added the new `ExtractRequest` field to the new upstream test literal and tightened its contract comment. The positional-boolean suggestion in [the prior review](https://github.com/lgse/strata/pull/1136#issuecomment-5768436668) remains a non-blocking API clarity nit: all four call sites were checked, and only the caller-reserved subfolder path passes `true`. Verified the updated cleanup test, full pinned quality suite (2318 passed, 20 ignored), and canonical E2E (818 passed). The native full Rust run had one unrelated chooser test fail transiently; its isolated rerun passed, as did the complete pinned quality run. No new UI changes were made in this review; the existing PR visual evidence remains applicable. No separate manual GUI run was performed.

wmfeht:
## Review (`81d6f69a2376a54fb0220e215288186a27a660a2`)

**Verdict: `approve-with-comments`** — no blockers. Nits do not need another code round.

Head has not moved. Judged this SHA on its own. The prior pass on `6d731be7` was also `approve-with-comments`. Since then the branch merged `main` at `14034a71` and this commit adapted the cleanup regression. CI on this SHA is green (format/lint/test, Rust shards, E2E, quality, dependency/spelling, metadata, release scripts; `[code]smith` skipped). A Blacksmith comment about `test_keep_both_preserves_archives_and_selects_each_numbered_output` was posted before those E2E jobs finished; the recorded check runs concluded success. `mergeable_state` `blocked` is required reviews, not a failing check.

### What holds

#1135: activating an archive reserves `<stem>/` with `create_dir` before dispatch, so the extractor saw `dest_dir.exists()` and skipped empty-folder cleanup.

- `created_dest = request.created_destination || !dest_dir.exists()`. Extractor-created destinations still clean up when the path is absent at start.
- `ExtractRequested` → `extract_entry_to_subfolder` is the only `true`, and only after `create_extraction_subfolder` wins an exclusive `create_dir` (collisions, including files and dangling symlinks, step to `stem (n)`). “Extract here”, “Extract to…”, and the password retry pass `false`. Those are the only four `Browser::extract` call sites.
- Failure and cancel cleanup is unchanged: not `Completed`, `read_dir` empty, then `remove_dir`. A pre-existing folder stays when the flag is false. A folder that already holds members stays.
- `..` is no longer a failure fixture. `sanitized_archive_path` clamps parent traversal and continues. Absolute paths, drive prefixes, and paths that sanitize to empty still fail closed. `extraction_provider_sanitizes_parent_paths_without_failure` writes `../escaped.txt` inside the destination and does not emit `Failed`.

### Blockers

None.

### Test quality

`failed_extraction_removes_a_caller_created_destination` is the #1135 path under the current failure mode: pre-created empty dest, `created_destination: true`, body `not an archive`, last event `Failed`, dest gone. It runs through `LocalOperationProvider.extract`. It fails if the flag is ignored, because a pre-existing dest with the flag false is kept (`failed_extraction_preserves_a_pre_existing_destination`). It is the other side of `failed_extraction_removes_a_newly_created_empty_destination` (flag false, dest absent at start).

`81d6f69` replaced `unsafe_extraction_removes_a_caller_created_destination`. That fixture used `../outside`, which this head extracts successfully, so asserting `Failed` and “unsafe” would not test cleanup. Dropping the outside-file assertion matches that. Matching any `Failed` is enough; the dest assertion is what catches a skipped cleanup. The other edits only add `created_destination: false` so existing `ExtractRequest` literals compile.

Not excess. Not tautological.

### Nits / suggestions

1. `Browser::extract` still takes `created_destination` as a positional `bool`. Four call sites; the subfolder path is the only `true`. The field name on `ExtractRequest` is the clear contract. A small enum at the browser boundary (`CallerCreated` / `UseExisting`) would make a missed `true` on activation, or a stray `true` on “Extract here”, obvious at the call. All four sites were checked on this head; this stays a clarity suggestion.

### Notes

- Prior GUI QA of zip-slip activation (`../evil.txt`, and `ok.txt` then `../nope.txt`) expected an error and no write outside. This head does not regress the cleanup fix. Those expectations are stale: merged main sanitizes `..` and completes the extract, so the subfolder stays because it holds the sanitized member. A malformed archive is the failure this fix still cleans up.
- No UI test that `extract_entry_to_subfolder` passes `true`. The adapter test sets the flag itself. Activation in `events.rs` calls `extract_entry_to_subfolder`, and that function passes `true` after `create_extraction_subfolder` succeeds.
- Empty pre-existing “Extract to…” folders stay, because that call passes `false`. The keep-user-content rule is covered with a file inside the folder.
- Password retry always passes `false`. After a failed empty subfolder extract, cleanup has already removed the directory, so the retry’s `!dest_dir.exists()` path recreates it and still removes it if that attempt also leaves it empty. The reserved folder can disappear before the password dialog opens; submit creates it again.

PR #1155, opened 2026-09-20
fix(e2e): restore native menus on Arch GTK 4.22

l0gicgate:
## Description

Move the shared quality/E2E environment from Ubuntu GTK 4.14 to a signed, dated Arch snapshot with GTK 4.22.5. Build a matching-source, test-image-only GTK accessibility-coordinate patch and verify its loaded path, checksum, and rendered bounds during image construction.

Repair the GTK 4.22 menu lifecycle, keyboard anchoring, focus restoration, modified-drag behavior, and submenu Left/Right navigation exposed by the upgraded toolkit. Re-enable the five ignored Rust menu tests and all 256 quarantined E2E cases, remove the quarantine machinery, and refresh the reviewed GTK 4.22 visual baselines. Auto worker selection now also respects cgroup task limits so Arch toolkit/media threads cannot exhaust the container.

Keep initial keyboard focus on the first available context-menu action as MIME lookup completes, without letting a stationary pointer take ownership. Replace the overlapping submenu return handlers with native navigation that clears GTK's open-submenu state, so Left followed by Up/Down stays in the parent menu.

Keep inline rename/create text aligned to its presentation: centered in Icons, left-aligned in List and Miller Columns, and mode-aware in filtered results and chooser views. Filtered Properties → Rename now edits the displayed result inline instead of opening a rename dialog. Shared modal viewports constrain oversized dialogs to the available window, with responsive Properties actions at large text scales. Use the canonical Lucide trash icon everywhere, removing the fill-level and open-lid variants and their count-based switching logic.

Replace the separate single-pane filtered-results view with a stable shared collection configurable for row or icon presentation, single or multiple selection, activation, focus, and selection callbacks. Icons now filters into an icon grid, List uses rows, Columns retains its native collection, and file choosers consume the same result behavior. Progressive updates preserve result identity, selection, focus, thumbnails, inline rename, drag/drop, quick preview, and real recursive locations for actions such as Open file location.

Resolve the eight guided-test findings: hide files in folder-only choosers, keep recursive queries/results consistent with file-type filters, retain matching renamed results (including shared searches and undo), remove redundant filename banners and option tooltips, contain keyboard background menus within their window, and scale dialog bezels with the text while filling the heading height. Retry thumbnail admission after newly mapped slots are allocated, so renamed image results render without a manual refresh.

## Visual evidence

[...]

Issue #1154, opened 2026-09-20
fix(menus): repair and re-enable quarantined native-menu regressions


l0gicgate:
### Strata version
PR #1085, v0.19.0, source commit `0c8eeaa312ca67b085494ccb3c370f6afbb7b49c`.

### Installation method
Built from source.

### Linux distribution and desktop environment
Canonical Ubuntu 24.04 container; GTK 4.14.5; Rust 1.98.1; private Xvfb and D-Bus/AT-SPI. No desktop session required. Separate experimental Arch/GTK 4.24.0 / GLib 2.90.0 runs are not the restored CI baseline.

### What happened?
The native GtkPopoverMenu migration in #1085 regressed menu lifecycle, focus and action dispatch. CI run [35483970576](https://github.com/lgse/strata/actions/runs/35483970576) reports three failed Rust cases and 104 failed parameterized E2E cases across the three shards. Symptoms include timeouts after menu actions, incorrect keyboard focus, and actual GTK popup-layout crashes. These are not all accessibility-coordinate errors.

The owner explicitly requested restoring the previous image/CI configuration, temporarily disabling confirmed failing tests, and tracking repairs here so the large feature PR can proceed. Native GTK submenus remain; skips do not establish that the behavior is safe or fixed. Host GTK/GLib upgrades are not reverted. The separate newer-GTK accessibility-coordinate investigation is preserved separately from these baseline failures.

### Steps to reproduce
1. Check out the failing source revision above.
2. Run the canonical pinned container with a private display/bus:
   `STRATA_CONTAINER_ENGINE=podman ./scripts/e2e.sh tests/e2e/scenarios/test_dialogs_and_menus.py -k 'file_properties or enter_submits_compress_and_extract_to_dialogs or enter_submits_the_copy_to_dialog' -n 0`.
3. Compress a file from its context menu, submit with Enter, then open the generated archive's context menu. Also exercise Properties, Copy to, repeated menu replacement/dismissal, and keyboard menu focus.
4. For each listed Rust case, use `./scripts/test-headless.py <fully-qualified-test-name>` on this revision. After quarantine lands, add `--ignored` to exercise an ignored Rust case; E2E will provide `--run-quarantined` for opt-in reproduction.

### Expected behavior
Native menus preserve selection/focus, dispatch exactly once, close cleanly, reopen on other entries, and do not crash or block subsequent dialogs/input. All quarantined regressions are restored to the normal required suites once fixed.

### Screenshots or video
[...]

PR #1157, opened 2026-09-20
fix(browser): retire deleted entries from the Recent view

### LINKED ISSUE #1156: Deleted files stay visible in the Recent view
### Strata version

0.19.0 (built from source, commit 984706cd)

### Installation method

Built from source

Omarchy, Hyprland (Wayland)

### What happened?

After deleting a file, it remains visible in an already-open Recent view until the view is manually reloaded. The file itself is correctly moved to Trash (or permanently deleted); only the Recent listing goes stale.

The Recent view resolves each bookmark to its real target path when enumerating `recent:///`, but the deletion update path only splices removed entries from the column showing the deleted file's parent directory. A `recent:///` column never matches that parent, so it is never told to drop the entry. The same applies to files moved elsewhere and to removals reported by the filesystem monitor.

### Steps to reproduce

1. Open the Recent view.
2. Select a listed file and press Del (or Shift+Del and confirm).
3. Observe that the entry is still shown in Recent even though the file is gone.
4. Navigate away and back to Recent: the entry is finally gone.

### Expected behavior

Deleting (or moving away) a file should remove its entry from an open Recent view immediately, matching the next reload.

### Additional context

JoeJoeflyn:
## Description

Deleting a file left it visible in an already-open Recent view until the view was manually reloaded. Recent entries resolve to their real target paths during enumeration, but deletion updates only spliced removed entries from the column showing the deleted file's parent directory — a `recent:///` column never matched that parent.

- `remove_deleted_locations` now also matches open `recent:///` columns when splicing deletions (covers deleting from the Recent view itself) and includes `recent:///` in the bulk-deletion refresh set.
- `handle_directory_change` fans the removal side of `Remove`/`Move` changes out to other open Recent columns, so monitor-driven removals and moves retire stale targets too. Only the removal side fans out: a moved target's new location is not a Recent member.

## Visual evidence

https://github.com/user-attachments/assets/e115e7ae-7db6-4984-a48a-c6989855f879

## How to test

1. Open the Recent view.
2. Select a listed file and press Del (or Shift+Del and confirm permanent deletion).
3. Observe the list.

**Expected result:** the entry disappears from Recent immediately; previously it stayed visible until navigating away and back. Deleting a file in a normal folder while Recent stays open also drops it from Recent.

## Related issue

Closes #1156


JoeJoeflyn:
The `End-to-end GUI suite` gate failure is the known native-menu quarantine (#1154): `test_context_menu_baseline` is in `tests/e2e/quarantined.json`, and `verify_reports` counts the skip as a non-pass. Unrelated to this diff — all shards passed; gate should go green once #1155 lands and this is rebased/rerun.

wmfeht:
## Review

**Verdict: request-changes**

Head `6e08115a305a731632d5e039a07a14b2f82ea274`. Closes #1156. No `working-docs/1156/` or `working-docs/1157/` on the branch. CI on this SHA is green (format/lint, Rust shards, E2E build and shards, end-to-end GUI suite, quality, dependency/spelling, metadata, release scripts). `[code]smith` skipped. The branch merges cleanly onto current `main` (`c5044ee`); it is behind, and `mergeable_state` is behind rather than conflicting.

### What holds

Deleting a file from Recent, or from another open column, drops that target immediately.

- `remove_deleted_locations` treats every open `is_recent_root()` column as a watched parent and splices `DirectoryChange::Remove` for the target path. Recent rows store that target (`recent_target_location` / `location_for_file`), so the splice matches. This runs after `deletion_operation` is cleared, which is the in-app delete completion path. A Recent-only Del no longer waits for a reload.
- Bulk deletes (`len > 64`) add `recent:///` to the refresh set, so an open Recent column reloads instead of trying to splice every path.
- A `Remove` or `Move` delivered to some other open column fans the old path out to other Recent columns. A file move drops the old Recent row and does not insert the destination. Recent's own monitor still collapses to `Rescan`.
- File deletes do not close later columns: `Path::starts_with` is component-wise, and a file is not an ancestor of an open directory.

### Blockers

1. **A directory Move fanned out to Recent closes later columns instead of following the rename.** `removed_location` turns `DirectoryChange::Move` into `DirectoryChange::Remove`, then `handle_directory_change` runs the full live path on the Recent column before the original column sees the Move. `path_after_external_change` for `Remove` truncates at the first later column whose location `is_within` the old path, and `apply_live_directory_change` `restore_path`s that truncated path. `restore_path` reloads what remains and drops the rest, so `relocate_open_columns` never rebases the open folder.

   Columns layout: Recent, then `~/Documents` (a Recent folder), then `~/Documents/project`. Rename `project` from the Documents column. The Documents monitor emits `Move`. The fan-out `Remove(~/Documents/project)` on Recent sees column 2 inside that path, resets navigation to Recent + Documents, and the project column is gone. Before this branch, the same Move rebased column 2 onto `project2`.

   `moved_targets_retire_the_old_recent_location_without_inserting_the_new_one` moves the file `/fixture/stale` while the only later column is `/fixture`, which is not inside that file, so the collapse stays green.

2. **Renaming or moving a Recent entry while its parent directory is not an open column leaves the row in place.** That is the Recent-only session in #1156, applied to "moving away". `publish_rename` returns immediately when no open column `is_within` the old path. `recent:///` is not inside the target file, so a Recent-only F2 never reaches the new fan-out. Cut/paste completion emits `TransferFinished` and only clears the clipboard; it does not splice Recent. The parent monitor is the only Move/Remove source, and it is not installed when that directory is not a column. Recent's own watch turns every event into `Rescan`, and a target rename does not change the recent store, so no rescan runs either. The old row stays until a manual reload. Delete is covered because `remove_deleted_locations` matches Recent directly. Rename and move are not.

### Test quality

Scanned the added and changed tests. No excess and no tautologies. Not asking for more tests beyond the two failing cases above.

- `deleted_targets_are_spliced_out_of_the_recent_view` fails if Recent is no longer spliced on delete. The name check is the user-visible result; the splice-and-no-reload check pins the incremental path.
- `removals_in_other_columns_retire_matching_recent_targets` fails if fan-out of `Remove` stops.
- `moved_targets_retire_the_old_recent_location_without_inserting_the_new_one` fails if the old file stays or the destination is inserted. It uses a file, so it does not cover blocker 1.
- `large_deletions_reload_an_open_recent_view` fails if bulk delete stops refreshing Recent. The deleted paths are under `/bulk`, so a reload of depth 0 only happens because `recent:///` was added.
- `test_quarantine_exempts_skips_only` fails if a listed skip is rejected or a quarantined teardown failure is accepted. It belongs to the unrelated gate commit (see nits). It is a real gate test, not a restatement of the Recent fix.

### Nits

These do not by themselves require another round.

1. `fix(e2e): tolerate quarantined skips in the coverage gate` is outside #1156. This SHA has no `tests/e2e/quarantined.json`. Current `main` removed that list in #1155. The merged gate only reads the file when it exists, so today's suite still requires every phase to pass. The exemption and its unit test are dormant. Drop that commit when rebasing.
2. Bulk refresh matches the literal URI `recent:///` only. `is_recent_root()` also accepts `recent://`, and location input allows it. A delete of more than 64 paths does not refresh a column opened with that spelling. The per-entry splice path does, because it uses `is_recent_root()`.

### Notes

Fan-out recurses through `handle_directory_change` and only skips the depth it is already on. A second Recent column would cycle. `navigate` replaces the path with one location, and Recent targets are real files, so a second Recent root is not reachable from normal navigation.

External deletes of a Recent target whose parent directory is not open still wait for a Recent `Rescan` or a manual reload. In-app delete does not, because of `remove_deleted_locations`.

l0gicgate:
Review follow-up: merged verified `main` at `dc008ba0`; fixed the two blocking cases in the prior review ([discussion](https://github.com/lgse/strata/pull/1157#issuecomment-5768466711)). Recent now retires targets on rename and completed cut/move even without the parent column open; monitor-driven directory moves retire the old Recent row without collapsing open descendant columns. Added focused regressions for all three paths. Bulk refresh also recognizes `recent://`. Removed the now-obsolete E2E quarantine exception because current main has no quarantine list. Manual Recent inspection used private Xvfb/D-Bus; the capture was discarded because the host Recent backend exposed unrelated personal history. No new visual attachment was uploaded; the original PR visual remains. Local pinned quality suite (format, Clippy, 2,388 Rust tests) and pinned E2E suite (821 tests) passed.

PR #1160, opened 2026-09-20
feat(archive): bundle spilled extractions under the archive stem

### LINKED ISSUE #1159: Extraction should bundle multi-root archives like Finder
### Problem or use case

Extracting an archive that contains several top-level entries ("Extract here", double-click, or "Extract to…") drops every entry loose into the destination folder. A zip of fifty loose photos spills fifty files into the current directory, mixing them with whatever already lives there.

The opposite wart also exists: activating an archive always wraps its contents in a folder named after the archive, so `foo.zip` that already contains a single `foo/` directory extracts to `foo/foo/`.

### Proposed solution

Match Finder / Archive Utility: extract an archive's contents into the destination, and when more than one top-level entry was written, bundle them into a fresh folder named after the archive (`photos.zip` → `photos/`). A single-root archive — one loose file or one top-level directory — extracts as-is, never double-wrapped.

### Alternatives considered

- Keep "Extract here" verbatim and "Extract to…" for wrapped extraction (status quo): both ends are wrong — spill for multi-root archives, double-wrap for tidy single-folder archives.
- Always create the wrapper folder before extraction (previous activation behavior): cannot know whether the wrapper is needed without inspecting the archive, and produces `foo/foo/`.

### Visual examples

N/A

### Strata version

main @ 528b1ab0

### Additional context

Implemented by bundling after decode: the extraction session reports the distinct resolved top-level paths it wrote, and the worker moves them into a reserved `stem`/`stem (n)` folder when there is more than one. Same-directory renames only, no archive pre-scan, cancellation and conflict naming unchanged.

JoeJoeflyn:
## Description

Extracting a multi-root archive now behaves like Finder: contents are written into the destination, then bundled into a fresh `<archive stem>/` folder only when more than one top-level entry landed. Single-root archives — one loose file or one top-level directory — extract as-is.

This replaces the previous split behavior where "Extract here" spilled loose files into the current folder and archive activation always wrapped (producing `foo/foo/` for tidy archives). All three entry points — "Extract here", double-click/Enter, and "Extract to…" — now share the same outcome, and password retry re-runs the same path.

Implementation: `ExtractionSession` reports the distinct resolved top-level paths it wrote; the worker reserves `stem`/`stem (n)` and moves the roots in with same-directory renames after a successful decode. No archive pre-scan, so `.tar.gz` is not decompressed twice, and cancellation/conflict-rename handling is unchanged.

## Visual evidence

https://github.com/user-attachments/assets/de3194de-1474-401c-8b92-83c8929751a6

## How to test

- Right-click → **Extract here** on a zip containing multiple top-level entries → contents land in `<archive name>/`, selected afterward.
- Same on a zip containing a single top-level folder (e.g. `foo.zip` with `foo/` inside) → folder extracts as-is, no `foo/foo/` double-wrap.
- Same on a zip containing one loose file → file lands loose, no wrapper.
- Double-click an archive → same smart behavior as Extract here.
- Extract the same multi-root archive twice → second run produces `name (1)/`, first stays intact.
- `./scripts/test-headless.py adapters::local_operations::archive` — covers zip, tar, tar.gz bundling, verbatim single-root, and name-collision suffixing.

## Related issue

Closes #1159


JoeJoeflyn:
CI note: all E2E shards pass. The `End-to-end GUI suite` aggregation gate fails only because quarantined `test_context_menu_baseline` (issue #1154, native-menu regression on GTK 4.22) counts as a non-pass — repo-wide until #1155 lands; unrelated to this diff.

l0gicgate:
Updated in `ea35df82` and `d51d8f98`.

### Changes
- Merged `main` (`0278a8c2`) and resolved the three conflicts, preserving its archive path sanitization and failed-destination cleanup fixes.
- Removed the obsolete quarantine documentation/gate exceptions and their test now that main has retired that mechanism. No new skips or weakened checks.
- Moved bundling into the confined destination abstraction. The extraction result retains the already-open directory handle through bundling; wrapper opens use `NOFOLLOW`, and descriptor-relative moves use `NOREPLACE`.
- Folder creation/move failures now report failure instead of silently claiming success. Already-extracted data is retained; an incomplete bundle can leave entries split between the destination and wrapper, which the error explains.
- Track actual created names (including duplicate-member conflict renames), with hash-based deduplication rather than repeatedly scanning the root list.
- Removed redundant activation screenshot capture and repeated input-routing code. Retained meaningful filesystem/content assertions, and extended existing coverage for duplicate members, directory replacement, bundle failures, 7z, Extract here, and Extract to. No layout/constant-echo tests added.

### Validation
Full validation was selected because the owner requested it and this merge crosses archive/UI and test-environment changes. Commands ran in the isolated worktree, with a task-owned rootless Podman store/wrapper on PATH; GUI checks used private displays/buses.

- `STRATA_CONTAINER_ENGINE=podman ./scripts/quality.sh test`: **2,392 passed, 20 explicitly ignored, 0 failed**. Existing ignores include manual benchmarks and separately-run GUI/GVfs/D-Bus cases; they are not claimed as passes.
- `STRATA_CONTAINER_ENGINE=podman ./scripts/e2e.sh`: **823 passed, no skips** (128 upstream Python/GI deprecation warnings).
- `STRATA_CONTAINER_ENGINE=podman ./scripts/quality.sh fmt` and `./scripts/quality.sh clippy`: **passed** after resolving formatting and two test-only Clippy diagnostics.
- Final scoped `./scripts/test-headless.py adapters::local_operations::archive`: **168 passed, 1 manual benchmark ignored** after those final changes. The full suites were preserved rather than unnecessarily repeated for formatting/assertion-message changes.
- Targeted archive activation/error E2E run: **21 passed**, including the formerly quarantined password-retry and Extract-to paths.
- `git diff --check`: passed. Independent read-only filesystem-safety review: **no blockers**.

Pinned environment: published base input key `803375aac22e811398d4735497f466405fd4c6e77b0629e5edaff52b8428aa6e`, GTK 4.22.5, Rust 1.98.1. No image rebuilds or visual-baseline updates. Existing PR visual evidence remains attached.

Remaining coverage limits: no dedicated multi-root RAR integration fixture or deterministic concurrent mutation test. Cancellation retains the existing decode-phase reporting; final bundling does not poll cancellation between renames. Fresh CI on the pushed head remains authoritative; this summary reports local results, not an unobserved CI pass.

wmfeht:
## Review

**Verdict: approve-with-comments**

Head: `d51d8f984ba3fbc34cc0a9f08745ba59f027f340`

#1159: multi-root archives should land in a fresh folder named after the archive, and a single top-level file or directory should extract as-is. Extract here, activation, and Extract to share that outcome. The wrapper is chosen after decode, from the distinct resolved top-level names the session actually wrote, so tidy archives do not become `foo/foo/` and `.tar.gz` is not decompressed twice.

No blockers.

What holds at this SHA:

- `ExtractionSession` records the resolved top-level component after a member is written, including conflict renames, and dedups with a set so nested members stay one root. `finish` hands those names to the caller with the already-open destination fd. Cancellation still returns the existing completed / failed / not-attempted report and does not bundle a partial extract.
- `ExtractionDestination::bundle_roots` creates `stem` or `stem (n)` with `mkdirat` on that pinned fd, opens the new directory with `DIRECTORY | NOFOLLOW`, and moves each root with `renameat_with(..., NOREPLACE)`. One root, or an unusable stem (empty, all dots, or containing `/`), stays put and the returned name is that root. `EEXIST` skips files, directories, and dangling symlinks. A mkdir or rename error fails the operation and says where the bytes remain; already-moved entries are left in place.
- `extract` bundles only a completed decode, on the blocking thread, before `Extracted`. `first_name` is then the wrapper, or the single root, and the existing completion path selects it. Activation calls `extract_entry` instead of pre-creating a subfolder. Password retry still targets the same destination the attempt used.
- A stem that collides with an extracted root (`photos.zip` containing `photos/` plus `extra.txt`) becomes `photos (1)/`, so the extracted directory is not merged into itself. An existing `bundle` file and `bundle (1)` symlink are left alone and the wrapper is `bundle (2)`.

CI on this SHA is green (format/lint, both Rust shards, E2E plan and shards, aggregate gates, dependency/spelling, metadata, release scripts). `[code]smith` skipped. `mergeable_state` is `blocked` because a review is required. CI was not re-run for this review.

## Blockers

None.

## Test quality

Scanned added and changed tests. No excess or tautological tests. No additional tests requested.

- `bundling_reports_failures_without_discarding_extracted_files` fails if an overlong wrapper name moves `first.txt`, or if a missing root rolls back the `first.txt` move that already succeeded.
- The bundle added to `pinned_destination_survives_path_replacement` fails if the wrapper is created through the replaced path instead of the pinned directory.
- `spilled_members_bundle_under_the_archive_stem` fails if zip, tar, tar.gz, or 7z leaves `a.txt` or `dir/` loose or reports a name other than `bundle`. One shared assertion over the four decoders, not a format × entry-point matrix.
- `single_root_extraction_lands_verbatim` fails if a one-file zip grows a `note/` folder or a one-directory zip grows `foldered/`.
- `a_bundle_name_colliding_with_an_extracted_root_is_suffixed` and `bundled_extraction_reserves_a_fresh_name_against_existing_entries` fail if the colliding extracted directory, the existing file, or the dangling symlink is overwritten.
- Decoder and session assertions that changed from `Option<String>` to the root list still read file bytes. They fail if a conflict rename or a second top-level member is dropped. The bundle check added to `unsafe_member_paths_are_sanitized_without_stopping_extraction` fails if sanitized names are not the roots that move.
- The activation end-to-end test still fails if a repeated extract overwrites `activation/activated.txt`. Extract here and Extract to are zip-only. The single-root RAR case is a separate test, not another axis on the multi-member loop. Removed UI helper tests belonged to the pre-created subfolder that this PR deletes.

## Nits / suggestions

These do not require another code round.

1. `bundle_roots` keeps its own suffix list (`.tar.gz`, `.tgz`, `.tar`, `.zip`, `.7z`, `.rar`) beside `ArchiveFormat::from_extension`, which `extract` has already used. The two match today, including `.tgz`. A later format or alias added to only one of them would wrap under the full filename, or under the wrong stem, while still decoding. Strip the extension the detected format already implies, and keep `.tgz` as the same alias `from_extension` uses.

2. If `mkdirat` succeeds and the following `openat` fails, the empty wrapper is left behind and the error only says the extracted entries remain in the destination. The rename-failure text already names the wrapper. The open failure can say the same.

## Notes

- A failed or cancelled decode does not bundle, so activation no longer contains a partial extract inside a pre-created folder. Loose members stay in the destination, which is the same spill Extract here already had. Empty caller-created destinations are still removed only when the directory is empty.
- The rename loop does not read the cancel flag. A cancel that arrives after decode returns still emits `Extracted` once the moves finish. The loop is a handful of same-directory renames.
- Multi-root RAR is not in the provider sweep. RAR still goes through `ExtractionSession`, and the single-member fixture covers the verbatim path. That gap is not a request for another test.

PR #1161, opened 2026-09-20
feat(sidebar): reorder special destinations and gate drags behind a hold

### LINKED ISSUE #1158: feat(sidebar): allow rearranging special destinations, including Recent
## Problem or use case

Recent is a useful sidebar destination, but it sits in a fixed position with the other special places (Home, Trash, Network). I want **Recent at the top** of the sidebar so it is the first place I click.

XDG folders (Desktop, Documents, Downloads, Pictures, Videos) and pins can already be drag-reordered. Special destinations cannot, so there is no way to put Recent first without a code change.

## Proposed solution

Let users drag special sidebar destinations into a custom order, using the same interaction as the existing XDG folder reorder, and persist that order across restarts.

At minimum, Recent should be movable so it can sit at the top. Ideally Home, Trash, Network, and Recent all participate in that shared order (or can be mixed with the reorderable XDG folders).

Pins and Devices can stay in their own sections.

## Alternatives considered

- Pinning a folder as a stand-in for Recent does not work: Recent is a virtual list of recently used files, not a folder.
- Hiding other places (#1024) can reduce clutter but does not move Recent to the top.
- A setting that only pins Recent first would cover my case, but full reorder of special destinations is more consistent with the current XDG/pin drag behavior.

## Visual examples

N/A — same drag-and-drop as Documents / Downloads / pins.

## Strata version

0.19.0

## Additional context

Related (not duplicates):

- #1082 added the Recent destination but did not make it reorderable
- #184 / #214 cover pin reorder only
- #1024 covers show/hide of default places

`sidebar_order` today only stores the XDG place ids (`desktop`, `documents`, `downloads`, `pictures`, `videos`). Recent is appended as a fixed `recent:///` row.

spandan11106:
## Description

Home, Trash, Network and Recent join the XDG folders in the saved `sidebar_order`, so any of them can be dragged into a custom position — Recent can now sit at the top. They render as one block (the separator that used to split specials from folders is gone, since the two sets now mix). An order saved before this change lists only the five XDG ids; the missing special ids are slotted back beside their default neighbours instead of being appended, so existing sidebars look unchanged after the upgrade.

Reorder drags are now armed by a press-and-hold (GTK's long-press time, ~500 ms) rather than starting on any movement while the button is down. A plain click navigates as before; holding the row highlights it and switches the cursor to a grab before the drag begins. This applies to pinned rows too, since they share the same drag plumbing.

## Visual evidence

Pending — I have local before/after captures of the sidebar (separator removed, Desktop now in the same block as Home/Trash/Network) at:

- `sidebar-before.png` — main
- `sidebar-after.png` — this branch

`gh` cannot upload attachments, so these need to be dragged into this description through the GitHub editor.

## How to test

1. Start Strata and look at the sidebar: Home, Trash, Network (and Recent, where the backend is available) sit in the same block as Documents/Downloads/etc.
2. Click Recent (or Home) — it navigates, no drag.
3. Press and hold a row for about half a second: it highlights and the cursor becomes a grab. Keep the button down and drag it above Home, then release.
4. Restart Strata.

**Expected result:** the row moves to the dropped position, a plain click never reorders anything, and the new order survives the restart.

Note: the headless E2E harness can start the reorder drag but never delivers the drop to a sidebar row (no existing scenario covers sidebar reordering either), so the drop half of step 3 is manual. Ordering and persistence are covered by the Rust sidebar tests.

## Related issue

Closes #1158


spandan11106:
https://github.com/user-attachments/assets/f6f3ac3a-4b6f-4b97-9053-38cf0c4c0a70

l0gicgate:
Merged latest upstream `main` (through `dc008ba0`) and fixed the hold-to-drag blocker: GTK ends `GestureLongPress` when motion crosses the drag threshold, so disarming on its `end` prevented `DragSource` from preparing a drag. The row now remains armed until button release or drag end; only primary-button holds arm it. Removed redundant comments during the PR comment audit. In an isolated GUI session I captured the Trash row before hold, after hold, and during a started drag; the held outline and dragging state appear as expected. A plain Home click navigated in the targeted E2E check. The sidebar drop/persistence gesture remains manually unverified (order/persistence Rust tests passed). Updated captures are local because `gh` cannot upload attachments; visual evidence upload remains pending via the GitHub editor.

PR #1165, opened 2026-09-20
fix(highlight): fix selection border styling

fadilasif:
## Description

Replace the theme text color border with theme accent border while preserving the existing selection geometry and behavior. The border thickness now matches the existing side-panel selection border.

## Visual evidence

https://github.com/user-attachments/assets/3f9dba82-d884-48c9-9664-4cc365334d54

## How to test

1. Open Strata and navigate to a directory containing multiple files.
2. Use keyboard navigation to move the selection/focus between files.

**Expected result:**

The selection border uses the current theme accent color, matches the side-panel border thickness, and retains the existing geometry and behavior.


l0gicgate:
Updated the Icons focus outline and inset to 1px, matching its selection outline and the List/Columns focus outlines. Tests skipped with explicit owner approval; `git diff --check` passed. Visual verification and updated visual evidence remain pending.

PR #1179, opened 2026-09-21
feat(browser): mirror keyboard selection in columns like Finder

### LINKED ISSUE #1178: Columns mode: mirror keyboard selection like Finder, and unify filter placeholders
## Problem or use case

In Columns mode, arrow-key selection does not update the pane to the right: pressing Up/Down onto a folder leaves the previous folder's column open (stale contents), and landing on a file leaves the folder column open. Finder's column view always mirrors the current selection in the next column — folder selection shows its contents, file selection shows a preview — so keyboard-only browsing requires an extra Right/Enter press per folder and can display misleading contents.

Separately, the per-pane filter placeholder is inconsistent: Columns shows a live count (`Filter N items…`) while List and Icons show a static `Filter items…`.

## Proposed solution

- In Columns mode, debounced keyboard selection mirroring: landing on a folder shows its contents in the next pane without moving focus; landing on a previewable file opens Quick Preview; landing on any other file closes the child pane. Pointer (mouse) selection keeps the configured click behavior. An explicit column close must not be reopened by the mirror, and the mirror must stay quiet for unreadable folders (inline error state, no dialog).
- Share the `Filter N items…` placeholder across all three view modes so List and Icons show the live item count too.

## Alternatives considered

- Right-arrow/Enter to enter folders works today but is slower than Finder's live mirror for keyboard navigation.
- Removing the count from Columns would unify placeholders the other way, but the count is useful when several panes with different counts are visible.

## Additional context

Finder behavior reference: in column view the right-hand column always reflects the current selection (folder → contents, file → preview column). Synchronous per-keystroke loading would be wasteful for rapid arrow presses, so the mirror should be debounced to the same 75 ms cadence already used for preview-follows-focus.

JoeJoeflyn:
## Description

Two Columns-mode gaps versus Finder, plus a placeholder inconsistency:

- **Filter placeholders** (`fix` commit): Columns already rendered `Filter N items…` while List and Icons kept a static `Filter items…`. The count formatter is now shared and refreshed from `show_count`, so all three modes show the live count.
- **Selection mirror** (`feat` commit): Up/Down in Columns now mirrors the selection in the next pane — a folder shows its contents without moving focus, a previewable file opens Quick Preview (gated by `single_click_previews`), and any other file closes the child pane. The mirror is debounced at the same 75 ms as `FOCUS_PREVIEW_DELAY` so rapid arrow presses only load the landing item, and it re-verifies focus at fire time so a stale selection or an explicit close cannot act or reopen. Pointer selection keeps the configured click behavior; inline rename and pending new-entry flows are guarded; unreadable folders surface their inline column error instead of a dialog.

## Visual evidence

https://github.com/user-attachments/assets/f1794838-816e-4efb-abb3-32c510f540c0

## How to test

1. Columns mode: arrow Up/Down onto a folder — the next pane shows its contents while focus stays in the current pane; onto a `.txt`/image/PDF — Quick Preview opens and the child pane closes; onto an unsupported file — the child pane closes.
2. Hold Down through a mixed list — only the landing folder loads (no per-keystroke lag).
3. Right on a folder enters the already-open child pane; Escape clears selection, then closes the pane — and it stays closed.
4. Ctrl+F in Icons, List, and Columns — all show `Filter N items…` with the live count, updating as entries appear.

**Expected result:** the next column always reflects the keyboard selection, like Finder's column view; placeholders are identical across modes.

## Related issue

Closes #1178


JoeJoeflyn:
still working on this

wmfeht:
Request changes on `25df0eb3eda40f6231dc15dd9ee958e1e17b2310`.

**Right into a mirrored column selects model index 0, not the first visible entry.**

`Browser::focus_child` is how Right enters an already-open child pane. A mirrored column is loaded with no selection (`show_child` passes `select_first_on_load: false`). The new branch then both sets `select_first_on_load` and calls `select(depth, 0)`.

`Navigation::select` takes `entries[0]` with no visibility check. The first-visible skip lives only in `apply_batch` / `install_snapshot`. Defaults are `show_hidden: false` and `folders_first: true`, and hidden names sort first, so index 0 is often `.git`, `.config`, or `.gitignore`. The column filter hides that row, but it stays in `column.entries`. After the load has finished there is no later batch to consume `select_first_on_load` (`finish` already cleared it, and this call sets it again). Right therefore focuses a hidden entry. That `FocusChanged` is still keyboard input, so the mirror opens that hidden directory in the next pane, or previews that hidden file, while the visible column shows nothing selected. A later sort snapshot still sees the flag and replaces whatever the user selected with the first visible row.

A fresh descend only sets `select_first_on_load` and lets the batch pick the first visible entry. `select(depth, 0)` is also a no-op while the column is still empty, so the bug shows up once the mirrored pane has finished loading — the Right-into-the-open-pane case.

`column_keyboard_selection_mirrors_folder_and_closes_on_file` never enters the child pane, and its fixture has no hidden names, so this path is unguarded.

l0gicgate:
Merged verified upstream `main` (ae677e2d) and resolved the browser import conflict. Addressed the [Right-into-mirrored-column review concern](https://github.com/lgse/strata/pull/1179#issuecomment-5768379806): entering an already-loaded child selects its first visible entry rather than model index 0; an in-flight load still selects on arrival. Added a regression with a hidden directory before a visible file and removed redundant new comments. Verified the Columns keyboard-preview interaction in isolated Xvfb (capture attached). Pinned Rust quality passed (2374 tests, 20 ignored); canonical E2E confirmation passed (821 tests). The initial E2E run had one intermittent List-navigation failure; its exact test and the confirmation run passed.

Columns keyboard selection showing the CSV preview

